PR #3, opened 2022-01-31
PTS-1 Allow installing Cairo contracts

kasperski95:
### Changes
- configured simple CLI using `argparse`
- added `install` command which accepts URL to a GitHub repository

### How to test it?
`python protostar.py install https://github.com/software-mansion/protostar`


src/commands/install/install.py
```diff
@@ -0,0 +1,48 @@
+from enum import Enum
+from os import path
+from typing import Optional
+from git import InvalidGitRepositoryError, NoSuchPathError, Repo
+from git.objects import Submodule
+from typing_extensions import TypedDict
+from .utils import map_url_to_package_name, verify_repo_url
+
+
+class InstallationErrorEnum(Enum):
```
Arcticae, line 10: In python we would preferably extend exception class and catch it where necessary using this class

src/commands/install/install.py
```diff
@@ -0,0 +1,48 @@
...
+    INCORRECT_URL = "INCORRECT_URL"
+    INVALID_LOCAL_REPOSITORY = "INVALID_LOCAL_REPOSITORY"
+    INVALID_PACKAGE_NAME = "INVALID_PACKAGE_NAME"
+
+
+InstallationResultType = TypedDict(
+    "InstallationResult", {"error": Optional[InstallationErrorEnum]}
```
Arcticae, line 17: I think we should not return `"error": None` since it might implicate some hidden exception being covered up

src/commands/install/install.py
```diff
@@ -0,0 +1,48 @@
...
+)
+
+
+def install(
```
Arcticae, line 21: That command does not work for me when i `rm -rf` ./lib - can we make it work like `node-modules`?

src/commands/install/install.py
```diff
@@ -0,0 +1,48 @@
...
+    url: str,
```
Arcticae, line 22: We probably should allow just repo name as the parameter, and perhaps an optional full URL if we want to support installing from other sources than github

src/commands/install/utils/verify_repo_url.py
```diff
@@ -0,0 +1,5 @@
+import re
+
+
+def verify_repo_url(url: str) -> bool:
+    return bool(re.match(r"^https:\/\/github.com\/[^\/]*\/[^\/]*$", url))
```
Arcticae, line 5: would ssh clones be possible?

src/commands/install/utils/map_url_to_package_name.py
```diff
@@ -0,0 +1,16 @@
+import re
+from typing import Optional
+
+
+def map_url_to_package_name(url: str) -> Optional[str]:
```
Arcticae, line 5: i think we should throw whenever we encounter a mapping error, this will prevent us from having `None`s in the flow later which would be hard to catch

Arcticae (changes requested):
Nice work! Some adjustments needed but really clean overall. Can you also set up `poe` task for tests and coverage? (you can copy it from starknet.py)

src/commands/install/install.py
kasperski95: The developer doesn't know what errors can be thrown by a function when that approach is used. Therefore, in Elixir and in expo projects (TypeScript, TBV) errors that don't require including stack trace are usually returned rather than thrown.

I'll change it to use exception classes because to keep things consistent. Is there a naming convention for marking functions as throwable @Arcticae?

src/commands/install/install.py
kasperski95: I'll try to figure something out when implementing the `uninstall` command. Removing lib directory is not enough because submodule information is added in `.git` directory.

src/commands/install/install.py
kasperski95: I'll add that option in the future. The repo name is not enough, it must be account-name/repo-name. Pasting a URL is easier than manually extracting package name that's why I focus on this approach.

src/commands/install/utils/verify_repo_url.py
kasperski95: I am not sure what you mean by `ssh clones`. Can you provide an example?

src/commands/install/install.py
kasperski95: https://github.com/software-mansion/protostar/issues/6

src/commands/install/install.py
kasperski95: https://github.com/software-mansion/protostar/issues/7

Arcticae (approved):
LGTM

PR #8, opened 2022-02-02
Allow removing dependencies

kasperski95:
### Changes
- added `remove` command


kasperski95:
> Seems like the changes which were already merged are shown in the diff, can we fix that?

It should be fixed now.

Arcticae (approved):
Looks good overall! Can we make the imports absolute in this one as well before merging?

PR #11, opened 2022-02-04
Support installing by account  repo  tag

kasperski95:
Added a support for installing packages by the following convention `ACCOUNT_NAME/REPO_NAME@TAG`


src/commands/install/utils/extract_info_from_repo_id.py
```diff
@@ -0,0 +1,90 @@
+import re
+from typing import NamedTuple, Optional
+
+from .. import installation_exceptions
```
Arcticae, line 4: Relative imports are inadvisable in python, for `__init__`s they are fine but i'd rather have an explicit import in the implementation, when repo happens to change its structure somehow

src/commands/install/install.py
```diff
@@ -5,14 +5,21 @@
 from git.repo import Repo
 
 from . import installation_exceptions
-from .utils import map_url_or_ssh_to_package_name
+from .utils import extract_info_from_repo_id
```
Arcticae, line 8: Can we have the import explicit here as well?

src/commands/install/utils/extract_info_from_repo_id.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class ExtractInfoFromRepoIdResult(NamedTuple):
```
Arcticae, line 7: I'd use dataclass here, and name is longish, maybe `RepoInfo` would suffice

src/commands/install/utils/extract_info_from_repo_id.py
```diff
@@ -0,0 +1,90 @@
...
+                url=_map_name_to_url(
+                    splitted_account_repo_name[0] + "/" + splitted_repo_name_tag[0]
+                ),
+            )
+
+    if result is None:
+        raise installation_exceptions.InvalidPackageName()
+
+    (package_name, tag, url) = result
+    return ExtractInfoFromRepoIdResult(
```
Arcticae, line 54: with dataclass, you can use `dataclasses.replace(...)`  here for simpler prop substitutions

Arcticae (changes requested):

src/commands/install/utils/extract_info_from_repo_id.py
kasperski95: By `explicit` do you mean `absolute` imports? 
Why are relative imports inadvisable in Python? Usually, I choose relative or absolute imports based on whichever gives fewer jumps between directories.

Arcticae: Yep that's what i meant

src/__init__.py
```diff
@@ -1,3 +1,3 @@
-from src.commands.install import installation_exceptions
-from .commands import install
+from src.commands import *
```
Arcticae, line 1: i dislike the asterisk as well, since it may pollute the namespace when code is changed inside the module with it

Arcticae (changes requested):

Arcticae:
You can merge after fixing the one issue, seems fine to me :)

kasperski95:
> You can merge after fixing the one issue, seems fine to me :)

(You can mark this PR as approved and enable resolving PR threads before merge.)

PR #19, opened 2022-02-07
`init` command - create project template

MaksymilianDemitraszek:
- [x] Add contracts field to the config
- [x] Remove dependencies fields
```
poetry run python protostar.py init
```


protostar.py
```diff
@@ -11,4 +12,7 @@
 cmd_remove_parser = root_subparsers.add_parser("remove")
 cmd_remove_parser.add_argument("package", type=str)
 
-cli(root_parser.parse_args())
+cmd_remove_parser = root_subparsers.add_parser("new")
+cmd_remove_parser.add_argument("project_name", type=str)
```
kasperski95, line 16: - `cmd_remove_parser` is meant to be used by remove command (`protostar remove <package_name>`). Each command should have its own subparser.
- I think `init` command would be a better name than `new` because it is a verb and is used more commonly ([foundry uses init](https://onbjerg.github.io/foundry-book/forge/creating-a-new-project.html)).

kasperski95, line 16: - `cmd_remove_parser` is meant to be used by remove command (`protostar remove <package_name>`). Each command should have its own subparser.
- I think `init` command would be a better name than `new` because it is a verb and is used more commonly ([foundry uses init](https://onbjerg.github.io/foundry-book/forge/creating-a-new-project.html)).

kasperski95 (changes requested):
Requesting changes due to:
- `new` command using wrong parser
- command for initializing a new project should be named `init`

src/commands/new/new.py
```diff
@@ -0,0 +1,58 @@
...
+    package.name = project_name
+    package.write()
+
+
+def copy_template(script_root, template_name: str, project_path: Path):
+    template_path = f"{script_root}/templates/{template_name}"
+    shutil.copytree(template_path, project_path)
+
+
+class PackageConfig:
```
Arcticae, line 25: I'd use dataclass here, with default values

templates/default/package.toml
```diff
@@ -0,0 +1,10 @@
+["protostar.general"]
+name = "template_package"
+description = "example description"
```
Arcticae, line 3: i'd leave it empty as a default

Arcticae (changes requested):

MaksymilianDemitraszek:
@kasperski95 
> I think init command would be a better name than new because it is a verb and is used more commonly ([foundry uses init](https://onbjerg.github.io/foundry-book/forge/creating-a-new-project.html)).

`init` will be different command used to initialize the `package.toml` in the existing project. This is more idiomatic in my opinion (example `git init`)

kasperski95 (approved):
LGTM. (I'd change the description field to be empty by default.)

protostar.py
```diff
@@ -3,11 +3,11 @@
 from argparse import ArgumentParser
 from pathlib import Path
 from re import Pattern
-
 import argparse
-
 from src import cli
 
+SCRIPT_ROOT = Path(__file__).parent
```
Arcticae, line 9: can't we just import the script root later on, instead of passing it as a parameter with the calls?

Arcticae (approved):
If the import can be done by importing - please try doing it that way, otherwise it looks good to me :)

PR #22, opened 2022-02-08
Project compilation

Arcticae:


src/commands/compile/test_compile.py
```diff
@@ -0,0 +1,54 @@
+import json
```
kasperski95, line 1: We should stick to one naming convention for test files.
`compile_test.py`

src/commands/compile/test_compile.py
```diff
@@ -0,0 +1,54 @@
...
+
+current_directory = Path(__file__).parent
+mock_outputs = Path(current_directory, "mock_output")
+
+
+@pytest.fixture(name="clear_mock_outputs")
+def clear_mock_outputs():
+    if mock_outputs.is_dir():
+        shutil.rmtree(str(mock_outputs))
+    mock_outputs.mkdir()
```
kasperski95, line 18: Output files can be stored in `tmpdir`, which is automatically cleared after each test.

kasperski95 (approved):
LGTM

src/commands/test/utils.py
```diff
@@ -9,6 +9,16 @@ def collect_subdirectories(root_dir: Path) -> List[str]:
     return [x[0] for x in os.walk(root_dir)]
 
 
+def collect_immediate_subdirectories(root_dir: Path) -> List[str]:
+    assert root_dir.is_dir(), f"{root_dir} is supposed to be a directory!"
+    result = []
+    for (root, dirs, _) in os.walk(str(root_dir.resolve())):
+        for directory in dirs:
+            result.append(str(Path(root, directory).resolve()))
+        break
```
MaksymilianDemitraszek, line 18: Maybe simpler but I'm not sure
```suggestion
    (root, dirs, _) = next(os.walk(str(root_dir.resolve())))
    return [str(Path(root, directory).resolve()) for directory in dirs]
```

protostar.py
```diff
@@ -53,12 +63,47 @@ def directory(directory_path: str) -> Path:
...
 )
 
+
+cmd_compile_parser = root_subparsers.add_parser("compile")
+cmd_compile_parser.add_argument(
+    "input-files", type=cairo_file, nargs="+", help="Input files"
+)
+cmd_compile_parser.add_argument(
+    "--output",
+    type=argparse.FileType("w", encoding="utf-8"),
```
MaksymilianDemitraszek, line 80: This is nice, but I am little bit worried that you have to remember to close it everywhere, do you have any idea how to prevent forgetting it?

src/commands/test/utils.py
```diff
@@ -9,6 +9,16 @@ def collect_subdirectories(root_dir: Path) -> List[str]:
     return [x[0] for x in os.walk(root_dir)]
 
 
+def collect_immediate_subdirectories(root_dir: Path) -> List[str]:
```
MaksymilianDemitraszek, line 12: Being honest i am not sure what this function does, could you rethink the naming?

src/utils/starknet_compilation.py
```diff
@@ -41,8 +46,17 @@ def preprocess_contract(self, cairo_file_path: Path) -> StarknetPreprocessedProg
         assert isinstance(context.preprocessed_program, StarknetPreprocessedProgram)
         return context.preprocessed_program
 
-    def compile_contract(self, cairo_file_path: Path) -> ContractDefinition:
-        preprocessed = self.preprocess_contract(cairo_file_path)
+    def compile_contract(
+        self, source: Optional[Path] = None, sources: Optional[List[Path]] = None
```
MaksymilianDemitraszek, line 50: What do you think about removing source and leaving just sources here?

MaksymilianDemitraszek: You could simply pass single source as `[source]`

MaksymilianDemitraszek (changes requested):
Generally very nice, some comments/ideas below

protostar.py
Arcticae: That's not opening the file handle itself, it should be done by `argparse` module itself automatically

src/commands/compile/compile.py
```diff
@@ -0,0 +1,33 @@
...
+
+from src.commands.test.utils import collect_immediate_subdirectories
+from src.utils.starknet_compilation import StarknetCompiler
+
+
+def compile_contract(
+    input_files: List[Path],
+    libraries_root: Path,
+    cairo_path: List[Path],
+    output_file: IO[str] = sys.stdout,
```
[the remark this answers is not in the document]
Arcticae: The output itself should be provided either by `argparse`, and will be closed automatically, and the `sys.stdout` doesn't have to be closed explicitly

protostar.py
```diff
@@ -13,7 +15,16 @@ def regexp(input_string: str) -> Pattern:
 
 def directory(directory_path: str) -> Path:
     pth = Path(directory_path)
-    assert pth.is_dir(), "Given argument is not a valid directory path"
+    assert pth.is_dir(), f'"{str(pth)}" is not a valid directory path'
+    return pth
+
+
+def cairo_file(file_path: str) -> Path:
```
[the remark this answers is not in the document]
Arcticae: That's just the type we provide for `argparse`, and thus it's named `cairo_file` since we provide it as a `type` parameter. I think if we have another case of validation of cairo files extension, we can consider extracting it somewhere else

src/cli.py
```diff
@@ -31,3 +32,11 @@ async def cli(args):
             cairo_paths=args.cairo_path,
             cairo_paths_recursive=args.cairo_path_recursive,
         )
+    elif args.command == "compile":
```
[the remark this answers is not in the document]
Arcticae: can you create a task for that one? we should at least consider it in the future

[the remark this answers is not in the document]
MaksymilianDemitraszek: done

PR #32, opened 2022-02-09
Support installing packages under custom name as a way of avoiding conflicts

kasperski95:
`python protostar.py install ... --name foo`


src/commands/install/install_package_from_repo.py
```diff
@@ -0,0 +1,49 @@
...
+    destination: str,
+):
+    logger = getLogger()
+
+    try:
+        repo = Repo(repo_root_dir)
+    except InvalidGitRepositoryError as _err:
+        raise installation_exceptions.InvalidLocalRepository()
+    except NoSuchPathError as _err:
+        raise installation_exceptions.InvalidLocalRepository()
```
MaksymilianDemitraszek, line 27:
```suggestion
```
Why do we catch this exceptions?

src/commands/install/install_package_from_repo_test.py
```diff
@@ -0,0 +1,27 @@
...
+import pytest
+from git.repo import Repo
+
+from src.commands.install import installation_exceptions
+from src.commands.install.install_package_from_repo import install_package_from_repo
+
+
+@pytest.fixture(name="repo_url")
+def fixture_repo_url():
+    return "https://github.com/starkware-libs/cairo-lang"
```
MaksymilianDemitraszek, line 12: I think we should't call external source from our tests, it is more problematic than it brings value.

src/commands/install/extract_info_from_repo_id.py
```diff
@@ -55,16 +62,16 @@ def extract_info_from_repo_id(repo_id: str) -> PackageInfo:
...
-def _map_ssh_to_url(ssh: str) -> Optional[str]:
+def _map_ssh_to_url(ssh: str) -> str:
     slug = _extract_slug_from_ssh(ssh)
     domain_match = re.search(r"(?<=git@).*(?=:)", ssh)
 
     if domain_match is None:
-        return None
+        raise installation_exceptions.InvalidPackageName("Couldn't map SSH to URL.")
 
     return f"https://{domain_match.group()}/{slug}"
```
MaksymilianDemitraszek, line 76: What do you think about writing tests for those functions? It can be error prone

MaksymilianDemitraszek (commented):
Generally very good job, some ideas/questions in the comments

src/commands/install/handle_install_command.py
```diff
@@ -0,0 +1,20 @@
+from os import getcwd
+from typing import Any
+
+from src.commands.install.extract_info_from_repo_id import extract_info_from_repo_id
+from src.commands.install.install_package_from_repo import install_package_from_repo
+
+
+def handle_install_command(args: Any) -> None:
+    if args.command != "install":
```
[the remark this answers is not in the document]
kasperski95: Nope. The function skips installation process if a command is different from `install`. (There's `return` below that if.)

src/commands/install/handle_install_command.py
```diff
@@ -0,0 +1,20 @@
+from os import getcwd
+from typing import Any
+
+from src.commands.install.extract_info_from_repo_id import extract_info_from_repo_id
+from src.commands.install.install_package_from_repo import install_package_from_repo
+
+
+def handle_install_command(args: Any) -> None:
```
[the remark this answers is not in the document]
kasperski95: - `install` seems to be too generic name in hindsight. Personally, I prioritize readability over consistency.
- `Namespace` type doesn't provide correct typing. `Any` type informs developer to not rely on typing.

src/commands/install/install_package_from_repo.py
```diff
@@ -0,0 +1,49 @@
...
+from git.objects import Submodule
+from git.repo import Repo
+
+from src.commands.install import installation_exceptions
+
+
+def install_package_from_repo(
+    name: str,
+    url: str,
+    tag: Optional[str],
```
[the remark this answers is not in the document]
kasperski95: Ok

src/commands/install/install_package_from_repo.py
kasperski95: It is a foundation for making errors “user-friendly”.  I am planning to create an abstract `ProtostarException` that all custom exceptions in the app should extend. In this way, we can later determine whether to display a nice error message, e.g. “Git repository is not initialized. Please run `git init`.” or  yeet stack trace.

src/commands/install/install_package_from_repo_test.py
kasperski95: Agree. I want lazy here.

src/commands/install/extract_info_from_repo_id.py
kasperski95: Ok

src/commands/install/install_package_from_repo.py
kasperski95: I'll add handling validation error in this task: https://github.com/software-mansion/protostar/issues/33.

protostar.py
```diff
@@ -33,6 +32,9 @@ def cairo_file(file_path: str) -> Path:
 
 cmd_install_parser = root_subparsers.add_parser("install")
 cmd_install_parser.add_argument("package", type=str)
+cmd_install_parser.add_argument(
+    "--name", type=str, help="Custom package's directory name"
```
Arcticae, line 36: is that argument required?

src/commands/install/handle_install_command.py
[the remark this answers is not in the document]
Arcticae: Actually we use this function like this:
```
   if args.command == "install":
        handle_install_command(args)
```
So maybe assertion would be more appropriate?

Arcticae (changes requested):

protostar.py
kasperski95: It's optional. All named arguments are optional by default.

src/commands/install/handle_install_command.py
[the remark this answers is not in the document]
kasperski95: Added assertion.

Arcticae (approved):

PR #41, opened 2022-02-14
Consistent package referencing across commands

kasperski95:
This PR adds support for the following forms of package referencing in `update` and `remove` commands:
- `sofware-mansion/protostar`
- `https://github.com/software-mansion/protostar`
- `protostar` (normalized and not normalized)
- the name provided to the `--name` flag during installation


src/commands/update/handle_update_command.py
```diff
@@ -1,13 +1,19 @@
 from os import getcwd, listdir, path
 
 from src.commands.update.update_package import update_package
+from src.utils import retrieve_real_package_name
 
 
 def handle_update_command(args) -> None:
     assert args.command == "update"
 
+    root_repo_dir = getcwd()
```
Arcticae, line 10: I will have a solution for getting those (pkg root, lib root) in a separate PR - can you make me a task and leave a todo here to integrate it?

Arcticae (approved):
LGTM

src/commands/update/handle_update_command.py
kasperski95: https://github.com/software-mansion/protostar/issues/55

PR #52, opened 2022-02-16
Build project from configuration

Arcticae:
- Hook up to the .toml config file 
- Create utilities to retrieve current package's config
- Rename `compile` to `build` 
- Fix an issue where test sources would not compile properly
- A few other fixes 

TODO:
- [ ] Write unit tests to cover diff


src/cli.py
```diff
@@ -18,6 +19,7 @@ async def cli(args, script_root):
     handler = StreamHandler()
     handler.setFormatter(StandardLogFormatter())
     logger.addHandler(handler)
+    current_package = Package.current()
```
kasperski95, line 27: Nits:
- `Package` in this project may refer to a dependency or to the project itself. Maybe `Package` should be renamed to `Project` because protostar project may not always be intended to be used as a package.
- current method name is a bit confusing to me. I would name it `get_instance()` or `create()`. Usually, methods and functions should start with a verb.

src/commands/init/init.py
```diff
@@ -1,73 +1,46 @@
...
-    copy_template(script_root, "default", project_path)
-
-    package = PackageConfig(project_path=project_path)
+    project_root = Path() / project_name
+    copy_template(script_root, "default", project_root)
+    package = Package(project_root=project_root)
 
-    package.name = project_name
+    def colored_input(message: str):
+        return input(f"{Fore.CYAN}{message}:{Fore.RESET} ")
```
kasperski95, line 19: Please use `src.utils.log_color_provider`. Sometimes a user may want to disable colors, for example when running a script on the CI.

kasperski95 (approved):
LGTM

PR #53, opened 2022-02-16
`upgrade` command

MaksymilianDemitraszek:
Maybe in the future: https://github.com/software-mansion/protostar/issues/54


src/commands/upgrade/upgrade.py
```diff
@@ -0,0 +1,116 @@
...
+        self.current_version = self.get_current_version()
+        self.latest_version_tag = self.get_latest_release()["tag_name"]
+        self.latest_version = version.parse(self.latest_version_tag)
+
+    def is_current_latest(self):
+        return self.latest_version == self.current_version
+
+    def upgrade(self):
+        logger.info("Looking for a new version ...")
+        if self.latest_version <= self.current_version:
```
kasperski95, line 47: I am not sure whether this condition will always work as expected e.g. "0.0.10" < "0.0.9"

src/commands/upgrade/upgrade.py
```diff
@@ -0,0 +1,116 @@
...
+        raise UpgradeManagerException(f"{platform} is not supported")
+
+    @classmethod
+    def get_latest_release(cls):
+        headers = {"Accept": "application/json"}
+        response = requests.get(f"{PROTOSTAR_REPO}/releases/latest", headers=headers)
+        return response.json()
+
+    def get_current_version(self):
+        path = self.protostar_dir / "dist" / "protostar" / "info" / "pyproject.toml"
```
kasperski95, line 113: Nice. I always used `os.path.join`.

kasperski95 (changes requested):
Requesting changes due to:
- comparing tags as strings

src/commands/upgrade/upgrade.py
MaksymilianDemitraszek: We do not compare them as strings but as parsed version objects

kasperski95: Yep, my bad. I checked the type of `latest_version_tag` rather than `latest_version`.

kasperski95 (approved):

PR #56, opened 2022-02-17
Add integration/e2e tests

kasperski95:
### Notes
- tests don't generate code coverage, therefore codecov reports might be misleading
- these tests use `python ./protostar.py` rather than binary itself
  - stuck at enabling python environment created by poetry while building an image
    - `RUN /bin/bash -c "sh $(poetry env info -p)/bin/activate && pyinstaller protostar.spec"`
    - `RUN poetry shell`

  - using pyinstaller directly on M1 is also problematic

### Changes
- added setup and basic e2e tests
- added test_installation github action


.github/workflows/test_installation.yml
```diff
@@ -0,0 +1,26 @@
+name: Test installation process
+
+on:
+  release:
+    types: [created]
```
Arcticae, line 5: We should probably trigger that on pull requests

kasperski95: What's the point of running this job on pull requests if the latest release doesn't change?

Arcticae: If it's run on a release it will fail silently unless you explicitly check it, and even if you check it, the release is already in progress

kasperski95: It's true, but running it on pull requests doesn't fix this issue because `install` script fetches the latest release anyway. Do you have an idea how to solve this problem?

Also, if the script fails, GitHub should send a notification.

Arcticae: We could modify the installation script to accept the gh revision, and run it on pull requests with that parameter

kasperski95: In order to make this work, we would have to build, upload and publish protostar. Then trigger this job once the package is ready to download, and modify half of the `install.sh` to support a custom url in order to download that specific package.

Adding such a test doesn't add value because the most fragile part of the script would have to be replaced.

PR #66, opened 2022-03-03
Interface importing

Arcticae:
A simple example - you can run tests by running pytest with:
`--target src/commands/test/runner_test.py::test_run_valid_tests`
Basically what this does is include `examples` path (with basic.cairo) within `cairo_path`, and runs the test function


MaksymilianDemitraszek:
Not relevant anymore

PR #67, opened 2022-03-07
Add `exceptRevert` cheatcode

kasperski95:
Based on: https://github.com/software-mansion/protostar/pull/65


kasperski95:
https://github.com/software-mansion/protostar/pull/68

PR #68, opened 2022-03-09
Add `exceptRevert` cheatcode (alternative way)

kasperski95:


src/commands/test/cheatcodes/syscall_handler.py
```diff
@@ -53,6 +56,44 @@ def _get_caller_address(
...
         return super()._get_caller_address(segments, syscall_ptr)
 
+    # mock_call
+    mocked_calls: Dict[AddressType, Dict[SelectorType, List[int]]] = {}
+
+    def register_mock_call(
+        self, contract_address: AddressType, selector: int, ret_data: List[int]
+    ):
+        if contract_address not in self.mocked_calls:
+            self.mocked_calls[contract_address] = {}
```
MaksymilianDemitraszek, line 66: you can probably use 
```
collections.defaultdict(dict)
```
here

src/commands/test/cheatcodes/curry_run_from_entrypoint.py
```diff
@@ -0,0 +1,13 @@
+def curry_run_from_entrypoint(fn, test_runner):
```
MaksymilianDemitraszek, line 1: I think name is a bit misleading because it doesn't only curry a function but edits input data either.

src/commands/test/cheatcodes/syscall_handler.py
```diff
@@ -53,6 +56,44 @@ def _get_caller_address(
...
+        self.mocked_calls[contract_address][selector] = ret_data
+
+    def unregister_mock_call(self, contract_address: AddressType, selector: int):
+        if contract_address not in self.mocked_calls:
+            raise Exception(
+                f"Contract {contract_address} doesn't have mocked selectors."
+            )
+        if selector not in self.mocked_calls[contract_address]:
+            raise Exception(
```
MaksymilianDemitraszek, line 75: What do you think about making this exception custom

src/commands/test/cheatcodes/curry_run_from_entrypoint.py
```diff
@@ -0,0 +1,13 @@
...
+    def curried_run_from_entrypoint(
+        *args,
+        **kwargs,
+    ):
+        if "hint_locals" in kwargs and kwargs["hint_locals"] is not None:
+            kwargs["hint_locals"]["__test_runner"] = test_runner
```
MaksymilianDemitraszek, line 7: Ok, I get now what you meant during the call, it's smart. I like it

src/commands/test/cheatcodes/curry_run_from_entrypoint.py
kasperski95: I'll change it to `inject_protostar_hint_locals`.

src/commands/test/cheatcodes/syscall_handler.py
kasperski95: I can create `CheatcodeException`  to avoid throwing generic exceptions.

kasperski95:
Marked this PR as draft again because `exceptRevert` doesn't work on the CI.

src/commands/test/runner.py
```diff
@@ -47,6 +52,11 @@
         match_pattern: Optional[Pattern] = None,
         omit_pattern: Optional[Pattern] = None,
     ):
+        original_run_from_entrypoint = CairoFunctionRunner.run_from_entrypoint
```
Arcticae, line 55: can we make it more generic, i.e. as an instance's context manager?

src/commands/test/runner.py
```diff
@@ -95,15 +106,37 @@
...
-                self.reporter.report(
-                    subject=test_subject, case_result=PassedCase(tx_info=call_result)
-                )
+                if self._is_test_error_expected:
+                    self.reporter.report(
+                        subject=test_subject,
+                        case_result=FailedCase(
+                            file_path=test_subject.test_path,
+                            function_name=function["name"],
+                            exception=BaseException("Excpected Revert."),
```
Arcticae, line 115: Excpected -> Expected (dot is redundant as well i think)

src/commands/test/cheatcodes/test_cheats.cairo
```diff
@@ -97,18 +97,18 @@ func test_mock_call_returning_struct{syscall_ptr : felt*, range_check_ptr}(contr
...
+    %{
+        mocked_fn_name="get_felt"
+        mocked_ret_data = [42]
+    %}
+    mock_call(EXTERNAL_CONTRACT_ADDRESS)
+    let (res) = ITestContract.get_felt(EXTERNAL_CONTRACT_ADDRESS)
+    assert res = 42
+
+    clear_mock_call(EXTERNAL_CONTRACT_ADDRESS)
+    except_revert()
```
Arcticae, line 111: this should probably be named `expect_revert` - right?

src/utils/modules.py
```diff
@@ -23,3 +23,9 @@
         return with_replaced_class
 
     return replace_class_instance
+
+
+def patch_class(full_path: str, new_module):
```
Arcticae, line 28: This seems to be unused

Arcticae (changes requested):
Looks good! Few nitpicks, functionality-wise it seems perfectly fine

src/commands/test/runner.py
kasperski95: Initially, I tried to use subclass approach but `@replace_class` decorator didn't work, and I have to somehow pass an instance of `TestRunner` to `CairoFunctionRunner`.

MaksymilianDemitraszek: Yess It will be a bit tricky, but we won't get away with this solution long term. It won't work for isolated states and won't parallelize. As I said imo we can merge it for now, I will let you know if I manage to figure out something better ;)

kasperski95: Ok, I am leaving this code for the time being.

Arcticae (approved):

PR #76, opened 2022-03-16
Feature/integration mvp

MaksymilianDemitraszek:


src/commands/test/examples/cheats/test_cheats.cairo
```diff
@@ -102,3 +102,38 @@ func test_clearing_mocks{syscall_ptr : felt*, range_check_ptr}(contract_address
...
+    func get_balance() -> (res : felt):
+    end
+end
+
+@external
+func test_deploy_contract{syscall_ptr : felt*, range_check_ptr}():
+    alloc_locals
+
+    local contract_a_address : felt
+    %{ ids.contract_a_address = deploy_contract("./src/commands/test/examples/basic.cairo") %}
```
kasperski95, line 121: I think `deploy_contract` should return dataclass (contract wrapper) with an address property. It would allow us to expand the interface without introducing breaking changes.

src/commands/test/custom_function_runner.py
```diff
@@ -0,0 +1,10 @@
+from starkware.cairo.common.cairo_function_runner import CairoFunctionRunner
+
+
+class CustomCairoFunctionRunner(CairoFunctionRunner):
```
kasperski95, line 4: Is this class used?

kasperski95 (changes requested):
Requesting changes due to:
- `deploy_contract` returning an address

src/commands/test/custom_function_runner.py
MaksymilianDemitraszek: Nah, pushed by mistake

src/commands/test/examples/cheats/test_cheats.cairo
MaksymilianDemitraszek: That's interesting idea, what could such interface include? Because for now wrapper would be only overhead for a user. We can always introduce a new interface by introducing a new function, so im not sure if simplicity is worth the sacrifice at this point.

kasperski95: We could allow user calling methods directly from hints, for example.

`Truffle` uses this approach, if I am not mistaken.

kasperski95: Adding new function increases the learning curve.

kasperski95 (approved):
Looks good, but I'd change `deploy_contract` to return contract object rather than contract address.

src/commands/test/examples/broken/test_basic_broken.cairo
```diff
@@ -7,6 +7,6 @@ func test_broken_contract{syscall_ptr : felt*, range_check_ptr}():
     alloc_locals
 
     local contract_address
-    %{ id.contract_address = deploy_contract("src/commands/test/examples/broken/basic_broken.cairo") %}
+    %{ id.contract_address = deploy_contract("src/commands/test/examples/broken/basic_broken.cairo").contract_address %}
```
kasperski95, line 10: I'd change `contract_address` to `address`. It's hard to confuse it with a different type of `address`, it's easier to type, and this property will often be used.

MaksymilianDemitraszek: It is cairo internal object, I don't like the idea of mapping it because of slightly simpler name

PR #77, opened 2022-03-16
Fix `syscall_ptr` error thrown in `mock_call`

kasperski95:


kasperski95:
@MaksymilianDemitraszek You can merge it to your PR. It should fix the problem.

PR #84, opened 2022-03-18
Add asserts

kasperski95:
- added asserts:
  - assert_eq
  - assert_not_eq
  - assert_{unsigned, signed}_{lt, le, gt, ge}

Asserts are written in hints and don't expect any implicit argument.


src/commands/test/asserts/asserts_test.py
```diff
@@ -0,0 +1,18 @@
...
+
+from src.commands.test import TestRunnerWithCheatcodes
+
+CURRENT_DIR = Path(__file__).parent
+
+
+@pytest.mark.asyncio
+async def test_asserts():
+    runner = TestRunnerWithCheatcodes(
+        include_paths=[path.join(CURRENT_DIR, "..", "..", "..", "..", "cairo")]
```
Arcticae, line 14: can we use pathlib for paths manipulation everywhere?

Arcticae (approved):
good job!

src/commands/test/asserts/asserts_test.py
kasperski95: We can, but maybe not in this PR. I'd wait for https://github.com/software-mansion/protostar/pull/76 to be merged.

kasperski95:
@Arcticae 
Moved asserts from cairo/asserts.cairo to cairo/protostar/asserts.cairo.
I think the PROTOSTAR_DIR shouldn't be included in CAIRO_PATH because imports should be prefixed with `protostar`. However protostar directory cannot be created because it conflicts with protostar.py. I think `PROTOSTAR_DIR/cairo` should be attached to CAIRO_PATH during installation.

PR #87, opened 2022-03-21
Added `--existing` flag, adopting protostar to existing project

MaksymilianDemitraszek:


protostar.py
```diff
@@ -79,6 +79,12 @@ def cairo_file(file_path: str) -> Path:
 
 
 cmd_init_parser = root_subparsers.add_parser("init")
+cmd_init_parser.add_argument(
+    "--existing",
+    default=False,
+    help="Adapts protostar to existing project",
+    action="store_true",
+)
```
kasperski95, line 87: I think we should avoid adding things that a user must memorize. Protostar could detect that directory is not empty when `protostar init` is run and ask user "Directory is not empty. Do you want to adapt current directory to protostar project?"

src/commands/init/init.py
```diff
@@ -1,53 +1,84 @@
...
+        self.interactive_input()
+        self.project_creation()
 
-    if lib_dir and not lib_pth.is_dir():
-        lib_pth.mkdir(parents=True)
-
-    project.write_config(
-        ProjectConfig(
+    def interactive_input(self):
+        project_name = ProjectCreator.request_input("Project name: ")
```
kasperski95, line 39: Currently, Protostar crashes when a user provides empty `project_name`. Can you fix it?

kasperski95 (changes requested):
Requesting changes due to:
- adding unnecessary argument to the CLI

src/commands/init/init.py
MaksymilianDemitraszek: Good catch

protostar.py
Arcticae: I'd go with the solution proposed by @kasperski95, or have 2 separate commands: `new` and `init`

kasperski95: >  if you create a project 99% of a time directory in which you run the script won't be empty
1) In the future, that might not be the case. This command will be rarely run, therefore it's quite likely a user will forget about this flag.

2) When a user runs `protostar init` in not empty directory, protostar should suggest `--existing` flag then.

@Arcticae @Solpatium Thoughts?

MaksymilianDemitraszek: > When a user runs protostar init in not empty directory, protostar should suggest --existing flag then.
We could print such suggestion, I like this soultion

MaksymilianDemitraszek: Two separate commands would be an overkill in my opinion, both would do basically the same

MaksymilianDemitraszek: We could specifically detect `.cairo` files (for example in 2/3 levels of depth?) to prevent some amout of false positives. We should still leave the flag though, somebody may find it usefull.

kasperski95 (changes requested):
Requesting changes due to:
- `protostar init` doesn't adapt a project in which a user initialized empty repo before calling init
- `protostar init --existing` doesn't create repository

tests/e2e/test_misc_commands.py
```diff
@@ -21,3 +22,55 @@ def test_init(project_name: str):
...
+    open(Path() / "example.cairo", "a").close()
+
+    child = pexpect.spawn(f"python {path.join(ACTUAL_CWD, 'protostar.py')} init")
+    child.expect("There are cairo.*", timeout=10)
+    child.sendline("y")
+    child.expect("Project name:", timeout=5)
+    child.sendline("")
+    child.expect("Please provide a non-empty project name:", timeout=5)
+    child.sendline(project_name)
+    child.expect("Project description:", timeout=1)
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: Can you make a task for that? I think this PR will become too big.

MaksymilianDemitraszek:
W also have to create new repo if it not exists

src/commands/init/init.py
```diff
@@ -1,53 +1,122 @@
...
+    files_depth_3 = glob.glob("*/*/*")
+    is_any_cairo_file = any(map(lambda f: f.endswith(".cairo"), files_depth_3))
+
+    out = False
+    if is_any_cairo_file:
+        out = input_yes_no(
+            "There are cairo files in your working directory.\n"
+            "Do you want to adapt current working directory "
+            "as a project instead of creating a new project?."
+        )
```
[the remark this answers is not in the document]
kasperski95: You're right, there's no point for checking any file. However, we must check if `.git` exists because package management commands won't work otherwise.

src/commands/init/init.py
```diff
@@ -1,53 +1,128 @@
...
+    can_be_a_project = any(
+        map(lambda f: f.endswith(".cairo") or f == ".git", files_depth_3)
+    )
+
+    out = False
+    if can_be_a_project:
+        out = input_yes_no(
+            "Your current directory may be a cairo project.\n"
+            "Do you want to adapt current working directory "
+            "as a project instead of creating a new project?."
```
kasperski95, line 125:
```suggestion
            "Do you want to adapt current working directory "
            "to a Protostar project instead of creating a new project?."
```

kasperski95 (approved):
Looks good, however I noticed few non-critical problems which can be solved in different PR:
- `protostar.toml` is overwritten when Protostar adapts current directory to a Protostar project
- protostar should ask to adapt a current dir if `.git` is present and `protostar.toml` is not

src/commands/init/init.py
kasperski95: (Maybe `convert` is a better word. `Adapt as` sounds wrong to me, but I suck at English.)

kasperski95:
I've just noticed that Protostar throws an error when a user cancels project configuration phase (cmd + c).

src/commands/init/init.py
MaksymilianDemitraszek: I think adapt is better, we adjust. not transform

PR #94, opened 2022-03-23
Include libs roots in test runner, and fix default libs directory on init

Arcticae:


src/utils/config/project.py
```diff
@@ -40,15 +42,24 @@ def config_path(self) -> Path:
...
         result["protostar.contracts"] = self.config.contracts
         return result
 
+    def get_include_paths(self) -> List[str]:
+        self._ensure_config_loaded()
+        libs_path = Path(self.project_root, self.config.libs_path)
+        return [
+            str(self.project_root),
+            str(libs_path),
+            *collect_immediate_subdirectories(libs_path),
```
kasperski95, line 64: Why is `collect_immediate_subdirectories` needed?

src/utils/config/project.py
```diff
@@ -40,15 +42,24 @@ def config_path(self) -> Path:
 
     @property
     def ordered_dict(self):
+        self._ensure_config_loaded()
```
kasperski95: I'd rename it to `_load_config_if_necessary`.

src/commands/test/run_test_runner.py
```diff
@@ -3,12 +3,11 @@
 from typing import List, Optional
 
 from src.commands.test.runner import TestRunner
-from src.utils.config.project import Project
 
 
 async def run_test_runner(
     tests_root: Path,
-    project: Optional[Project] = None,
+    project: Optional["Project"] = None,
```
kasperski95: Why did you replace import with string literal?

kasperski95 (approved):
LGTM

src/utils/config/project.py
Arcticae: Because when we're loading sources from libraries, the libraries refer to the source files from the root of the library. See openzeppelin for example - they import straight from openzeppelin which is not in our `cairo_path` if we don't include `cairo_contracts` in `cairo_path`

Arcticae:
@kasperski95 it already does - see this error:
```
starkware.cairo.lang.compiler.import_loader.ImportLoaderError: /Users/timelock/swm/starkware/access-protocol-starknet/lib/cairo_contracts/openzeppelin/token/erc20/library.cairo:13:6: Could not find module 'openzeppelin.utils.constants'. Searched in the following paths:
/Users/timelock/swm/starkware/access-protocol-starknet/openzeppelin/utils/constants.cairo
/Users/timelock/swm/starkware/access-protocol-starknet/lib/openzeppelin/utils/constants.cairo
/Users/timelock/swm/starkware/access-protocol-starknet/openzeppelin/utils/constants.cairo
/Users/timelock/.protostar/dist/protostar/openzeppelin/utils/constants.cairo
from openzeppelin.utils.constants import TRUE, FALSE, UINT8_MAX
     ^**************************^
```

This -> `/Users/timelock/.protostar/dist/protostar/openzeppelin/utils/constants.cairo`

PR #100, opened 2022-03-24
Pretty print Cairo compilation errors

kasperski95:


src/commands/build/build_project.py
```diff
@@ -12,18 +13,21 @@ def build_project(
...
+            contract = StarknetCompiler(
+                include_paths=project_paths,
+                disable_hint_validation=disable_hint_validation,
+            ).compile_contract(
+                *[Path(component) for component in contract_components],
+            )
+        except Exception as err:
+            raise CairoCompilationException(
+                f"Protostar couldn't compile '{contract_name}' contract\n{str(err)}"
+            ) from err
```
MaksymilianDemitraszek, line 30: I think catching `Exception` is a bit too general, what do you think about catching `StarkException` only? We would have clear distinction between cairo code error and protostar code error.

Another idea is to pretty-print it only, instead raising python error. Imo would be nicer and cleaner from user perspective. 

What do you think?

MaksymilianDemitraszek (changes requested):
Good job! Some minor questions

We probably would benefit from having integration test/s for building. Could you add it here/create a task for that?

src/commands/build/build_project.py
kasperski95: > I think catching Exception is a bit too general, what do you think about catching StarkException only? We would have clear distinction between cairo code error and protostar code error.

I used `Exception` because `HintException` was thrown in the use case described by Jakub. I'll try using `StarkException`, Cairo is probably rethrowing`HintException` as `StarkException` error.

> Another idea is to pretty-print it only, instead raising python error. Imo would be nicer and cleaner from user perspective.

`CairoCompilationException` extends `ProtostarException` and all `ProtostarException`s are pretty printed, therefore it wouldn't change anything from user perspective.

kasperski95: https://github.com/software-mansion/protostar/issues/105

kasperski95: I can't use `StarkException`. If I used it, Protostar doesn't catch hint exception.

MaksymilianDemitraszek: Why can't you use both `StarkException` and `HintException`?

kasperski95: 1) There might another unexpected exception I didn't think of.
2) For all exceptions, I'd use the same code anyway.

What's the benefit of catching exceptions one by one in this case?

MaksymilianDemitraszek: > What's the benefit of catching exceptions one by one in this case?

Because if we have a bug in our python code, now or in the future, it will catch it as an cairo exception. It is not super critical here as we print it anyway, but would give a user feedback if his code is broken or it's just we messed up. I believe that, long term, handling such stuff explicitly leads to a better and more predictable software, even if it requires extra work and doesn't seem necessary now. You could use `VmException` or from cairo lib to make it more general. 

We probably will have to build up such list of exceptions, because we need to use them for tests either (We have to add `VmException` there as we catch `StarkException` only). Leaving the final decision to you. 

> For all exceptions, I'd use the same code anyway.
FYI, you can handle multiple error types using the same code.
```
except (StarkException, VmException) as err:
```

MaksymilianDemitraszek (approved):

PR #104, opened 2022-03-24
Add protostar's cairo dir to caito_path

kasperski95:
- [x] bundle cairo dir

Protostar searches for `protostar` and uses its location to deduce the location of the Protostar's Cairo directory.

This PR is hard to test since the code depends on `protostar` binary being added to `PATH`. I am going to merge it despite failing test-coverage checks.


src/cli.py
```diff
@@ -41,18 +45,24 @@ async def cli(args, script_root):
...
                 getattr(args, "tests-root"),
                 project=current_project,
                 omit=args.omit,
                 match=args.match,
-                cairo_paths=args.cairo_path,
+                cairo_paths=cairo_paths,
             )
         elif args.command == "build":
+            cairo_paths: List[Path] = args.cairo_path
+            if protostar_dir:
```
MaksymilianDemitraszek, line 60: Will be any problem in case of duplicate? I guess not, just want to make sure.

protostar.py
```diff
@@ -1,14 +1,23 @@
...
 from pathlib import Path
 from re import Pattern
+from typing import Optional
 
 from src import cli
 
 SCRIPT_ROOT = Path(__file__).parent
 
+protostar_dir: Optional[Path] = None
+protostar_path = shutil.which("protostar")
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: We probably should unify the way we handle the installation dir, as for upgrade we use fixed path.

MaksymilianDemitraszek (changes requested):
Good job, minor questions

src/cli.py
kasperski95: I don't know tbh. I changed the code to skip append if path is already included.

MaksymilianDemitraszek:
https://github.com/software-mansion/protostar/issues/109

MaksymilianDemitraszek (approved):

PR #108, opened 2022-03-28
Improved default template

MaksymilianDemitraszek:


templates/default/src/main.cairo
```diff
@@ -0,0 +1,5 @@
+func main(a : felt, b : felt) -> (res : felt):
+    let x = 'Hello World'
+    return (a+b)
+end
```
kasperski95, line 4: I'd use Balance example from cairo docs.
https://www.cairo-lang.org/docs/hello_starknet/intro.html#your-first-contract

Balance contract seems to be the "Hello World" for blockchain technologies.

kasperski95 (approved):
LGTM

PR #129, opened 2022-04-01
Added adapter object

MaksymilianDemitraszek:


src/commands/test/runner.py
```diff
@@ -230,3 +235,12 @@ def deploy_contract(contract_path: str):
...
     def expect_revert(self):
         self._is_test_error_expected = True
+
+
+class DeployedContact:
+    def __init__(self, starknet_contract: StarknetContract):
+        self._starknet_contract = starknet_contract
+
+    @property
+    def contract_address(self):
```
MaksymilianDemitraszek, line 245: @kasperski95 I know you opted for naming this `address` before but I think we should keep the interface similar to the one used by starknet tesing framework

kasperski95 (approved):
LGTM

PR #132, opened 2022-04-01
Improve `expect_revert`

kasperski95:
- updated `expect_revert` to accept regex 
- updated test runner to expect test to fail if its starts from `test_fail_` 

```python
        def expect_revert(
            error_type: str = ".*", error_message: str = ".*"
        ) -> Callable[[], None]: ...
```

### Usage
```cairo
@external
func test_fail_error_was_not_raised_before_stopping_expect_revert{
        syscall_ptr : felt*, range_check_ptr}():
    alloc_locals

    %{ stop_expecting_revert = expect_revert("UNINITIALIZED_CONTRACT") %}
    local contract_a_address = 42
    %{ stop_expecting_revert() %}

    return ()
end
```

- [x] except revert should take regex as r"..." or treat current strings as strings


src/commands/test/runner.py
```diff
@@ -150,22 +166,45 @@ async def invoke_test_function(self, function_name: str):
         )
 
         func = getattr(self.test_contract, function_name)
+        is_failure_expected = (
+            function_name.startswith("test_fail_")
+            if self._is_test_fail_enabled
+            else False
+        )
```
MaksymilianDemitraszek, line 173:
```suggestion
        is_failure_expected = function_name.startswith("test_fail_") and self._is_test_fail_enabled
```

src/commands/test/runner.py
```diff
@@ -150,22 +166,45 @@ async def invoke_test_function(self, function_name: str):
...
+                return call_result
+
+            except StarkException as ex:
+                is_ex_unexpected = (
+                    self._expected_error is None
+                    or re.compile(self._expected_error.name).search(ex.code.name)
+                    is None
+                    or re.compile(self._expected_error.message).search(ex.message or "")
+                    is None
                 )
```
MaksymilianDemitraszek, line 193: I would implement a `ExpectedError.match(ex)` or something similar

src/commands/test/runner.py
```diff
@@ -150,22 +166,45 @@ async def invoke_test_function(self, function_name: str):
...
-        except StarkException as ex:
-            if not self._is_test_error_expected:
-                raise StarkExceptionReportedException(ex) from ex
+                if is_ex_unexpected:
+                    raise StarkExceptionReportedException(ex) from ex
 
+                if is_failure_expected:
+                    raise TestNotFailedReportedException() from ex
+        except TestNotFailedReportedException as ex:
+            raise ex
```
MaksymilianDemitraszek, line 201: Why do we catch it?

src/commands/test/runner.py
```diff
@@ -150,22 +166,45 @@ async def invoke_test_function(self, function_name: str):
...
+            if self._is_test_fail_enabled
+            else False
+        )
         # TODO: Improve stacktrace
         try:
-            call_result = await func().invoke()
-            if self._is_test_error_expected:
-                raise MissingExceptReportedException(
-                    "Expected a transaction to be reverted"
+            try:
```
MaksymilianDemitraszek, line 176: Please move this try into separate method to decrease cognitive complexity

src/commands/test/runner.py
```diff
@@ -150,22 +166,45 @@ async def invoke_test_function(self, function_name: str):
...
+                call_result = await func().invoke()
+                if self._expected_error is not None:
+                    raise MissingExceptReportedException(
+                        "Expected a transaction to be reverted"
+                    )
```
MaksymilianDemitraszek, line 181: I think we should provide information here about expected error and provide proper `__str__` for the class as it's being used by the reporter to print data.

MaksymilianDemitraszek (changes requested):
LGTM, In the future we have to get rid of `test_failed` patter though, but it's ok for now

src/commands/test/runner.py
kasperski95: If I move it to a separate method, the code will be still hard to understand. I think even more, because
```python
except TestNotFailedReportedException as ex:
            raise ex
```
is justified by the logic of the code inside the inner `try`.

The code is hard to understand because it supports unintuitive use-cases e.g.:
if a test reverts, an exception was expected, and the test was expected to fail, the test will fail.

src/commands/test/runner.py
kasperski95: To avoid `TestNotFailedReportedException` being caught by the code below. Otherwise, `TestNotFailedReportedException` would never be thrown from that function.

src/commands/test/runner.py
kasperski95: I think this is an anti-pattern because the code `ex.code.name` is specific to `StarkException` and `match` should accept `BaseException`.

src/commands/test/runner.py
kasperski95: I added info about the expected error there.

This exception is used in various places. The message changes, thereby the way how an expected error should be interpolated into the error message.

src/commands/test/test_environment_exceptions.py
```diff
@@ -10,6 +10,11 @@ class MissingExceptReportedException(ReportedException):
     pass
 
 
+class TestNotFailedReportedException(ReportedException):
```
Arcticae, line 13: This naming is not really intuitive, can we skip `ReportedException` in the names?

src/commands/test/runner.py
```diff
@@ -150,22 +166,44 @@ async def invoke_test_function(self, function_name: str):
...
         # TODO: Improve stacktrace
         try:
-            call_result = await func().invoke()
-            if self._is_test_error_expected:
-                raise MissingExceptReportedException(
-                    "Expected a transaction to be reverted"
+            try:
+                call_result = await func().invoke()
+                if self._expected_error is not None:
+                    raise MissingExceptReportedException(
```
Arcticae, line 178: I would advise strongly to thoroughly test this function if we're not going to be simplifying the logic, so when other people try to tweak the logic we'll have a clear idea what the cases are here.

Arcticae (changes requested):
Other than adding some test cases to cover all branches i think it's perfectly fine 👌

src/commands/test/test_environment_exceptions.py
kasperski95: Removed `Reported` part, but left `Exception` part.

Arcticae (approved):

PR #146, opened 2022-04-06
Add paths to the deployment of contracts in test instance

Arcticae:


src/commands/test/runner.py
```diff
@@ -135,15 +135,21 @@ class ExpectedError:
 
 
 class TestExecutionEnvironment:
-    def __init__(self, is_test_fail_enabled: bool):
+    def __init__(self, is_test_fail_enabled: bool, include_paths):
```
kasperski95, line 138: It would be nice to have a type here. Sometimes paths are `str`ings, sometimes `Path`s.

kasperski95 (approved):
LGTM

src/commands/test/runner.py
Arcticae: Ah sorry i missed it

PR #168, opened 2022-04-11
error mismatch

MaksymilianDemitraszek:


src/commands/test/runner.py
```diff
@@ -189,17 +195,15 @@ async def invoke_test_function(self, function_name: str):
...
+                assert self._expected_error is not None
+                if not self._expected_error.match(ex):
+                    raise ExceptMissmatchException(
+                        expected_name=self._expected_error.name,
+                        expected_message=self._expected_error.message,
+                        received=ex,
+                    ) from ex
 
             if not is_ex_expected:
                 raise StarkReportedException(ex) from ex
```
kasperski95, line 209:
```suggestion
            if self._expected_error:
                if not self._expected_error.match(ex):
                    raise ExceptMissmatchException(
                        expected_name=self._expected_error.name,
                        expected_message=self._expected_error.message,
                        received=ex,
                    ) from ex
            else:
                raise StarkReportedException(ex) from ex
```

src/commands/test/test_environment_exceptions.py
```diff
@@ -10,6 +11,28 @@ class MissingExceptException(ReportedException):
     pass
 
 
+class ExceptMissmatchException(ReportedException):
+    def __init__(
+        self,
+        expected_name: Union[str, None],
+        expected_message: Union[str, None],
```
kasperski95, line 18: > You can use the [Optional](https://docs.python.org/3/library/typing.html#typing.Optional) type modifier to define a type variant that allows None, such as Optional[int] (Optional[X] is the preferred shorthand for Union[X, None])

```suggestion
        expected_name: Optional[str],
        expected_message: Optional[str],
```

src/commands/test/test_environment_exceptions.py
```diff
@@ -10,6 +11,28 @@ class MissingExceptException(ReportedException):
     pass
 
 
+class ExceptMissmatchException(ReportedException):
```
kasperski95, line 14:
```suggestion
class ExceptMismatchException(ReportedException):
```

src/commands/test/test_environment_exceptions.py
```diff
@@ -10,6 +11,28 @@ class MissingExceptException(ReportedException):
...
+        self.expected_message = expected_message
+        self.received = received
+        super().__init__()
+
+    def __str__(self) -> str:
+        message = [
+            "Expected:",
+            f"name: {self.expected_name}, message: {self.expected_message}",
+            "Instead got:",
+            f"name: {self.received.code.name}, message: {self.received.message}",
```
kasperski95, line 31: We should consider printing the message in a new line. Usually, the message is very long.

kasperski95 (changes requested):
Nits

src/commands/test/test_environment_exceptions.py
MaksymilianDemitraszek: I thought optional can only be a keyword arg for some reason

PR #199, opened 2022-04-13
Add `expect_events` cheatcode

kasperski95:
- added `expect_events` cheatcode
- updated `expect_revert` logic, to allow expecting errors raised by other cheatcodes
- updated docs


src/commands/test/utils.py
```diff
@@ -25,6 +29,82 @@ def extract_core_info_from_stark_ex_message(msg: Optional[str]) -> Optional[str]
...
+            + '"'
+        )
+        if self.data:
+            result += ', "data": "' + str(self.data) + '"'
+        if self.from_address:
+            result += ', "from_address": "' + str(self.from_address) + '"'
+        result += "}"
+        return result
+
+    @classmethod
```
MaksymilianDemitraszek, line 71:
```suggestion
    @staticmethod
```

src/commands/test/utils.py
```diff
@@ -25,6 +29,82 @@ def extract_core_info_from_stark_ex_message(msg: Optional[str]) -> Optional[str]
...
+    def find_first_expected_event_not_included_in_state_events(
+        cls, expected_events: List["ExpectedEvent"], state_events: List[Event]
```
MaksymilianDemitraszek, line 73:
```suggestion
        expected_events: List["ExpectedEvent"], state_events: List[Event]
```

src/commands/test/utils.py
```diff
@@ -25,6 +29,82 @@ def extract_core_info_from_stark_ex_message(msg: Optional[str]) -> Optional[str]
...
+                    recent_match_state_event_index = se_index
+                    is_match = True
+                    continue
+
+            if matches_count == len(expected_events):
+                return None
+
+            if not is_match:
+                return expected_event
+        assert False, "Unreachable code"
```
MaksymilianDemitraszek, line 95: We could do something like this to make it more readable (untested)
```suggestion
        assert len(expected_events) > 0
        for expected_event in expected_events:
            try:
                state_events  = self.get_next_match(expected_event, state_events)
            expect NotFoundError:
                return expected_event
     return None
            
def get_next_match(expected, state):
    for (se_index, state_event) in enumerate(state):
                if expected_event.match(state_event):
                    return state[se_index+1:]
    raise NotFoundError
            
```

src/commands/test/runner.py
```diff
@@ -280,9 +305,40 @@ def clear_mock_call(contract_address: int, fn_name: str):
...
+                RevertableException(error_type=error_type, error_message=error_message)
+            )
+
+        @register_cheatcode
+        def expect_events(
+            *raw_expected_events: ExpectedEvent.CheatcodeInputType,
+        ) -> None:
+            assert self.starknet is not None
+
+            def stop_expecting_emit():
```
MaksymilianDemitraszek, line 319: It's probably unnecessary

src/commands/test/runner.py
```diff
@@ -196,29 +184,66 @@ async def invoke_test_function(self, function_name: str):
...
+            raise ex
         finally:
             CairoFunctionRunner.run_from_entrypoint = original_run_from_entrypoint
             self._expected_error = None
 
+    def subscribe_to_test_finish(
+        self, listener: Callable[[], None]
+    ) -> Callable[[], None]:
+        listener_id = uuid()
+        self._test_finished_listener_map[listener_id] = listener
```
MaksymilianDemitraszek, line 240: If we remove ` stop_expecting_emit` we can use a `set()` here

MaksymilianDemitraszek: I would also name it `add_test_finish_hook` subscribe suggests some signaler

src/commands/test/runner.py
```diff
@@ -196,29 +184,66 @@ async def invoke_test_function(self, function_name: str):
...
+            except RevertableException as ex:
+                if self._expected_error:
+                    if not self._expected_error.match(ex):
+                        raise ExpectedRevertMismatchException(
+                            expected=self._expected_error,
+                            received=ex,
+                        ) from ex
+                else:
+                    if ex.original_exception:
+                        raise ex.original_exception
```
MaksymilianDemitraszek, line 220: Do we need this extra `try-except` layer? Can't we just raise `StandardReportedException` here? Would be much much easier to comprehend.

MaksymilianDemitraszek (changes requested):
Looks good, some changes requested

src/commands/test/runner.py
MaksymilianDemitraszek: Aside, I really like flexibility of hook-based approach, good job

src/commands/test/runner.py
```diff
@@ -196,29 +184,66 @@ async def invoke_test_function(self, function_name: str):
 
         func = getattr(self.test_contract, function_name)
 
-        # TODO: Improve stacktrace
+        # pylint: disable=too-many-nested-blocks
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: If the change i suggested below can be done, we might not need this second layer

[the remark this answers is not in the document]
kasperski95: (Removing the `BaseException` doesn't add reduce layer count.)

I agree that nesting `try` is a bad practice. However, splitting logic of this function will be artificial, and it makes the code harder to understand due to boilerplate.

...

I've just noticed the suggestion at the very bottom, but I came to the same conclusion.

src/commands/test/runner.py
kasperski95: `stop_expecting_emit` isn't returned by the cheatcode. It is only called by the "test finished hook", so I have to preserve this function (under new name).

src/commands/test/test_environment_exceptions.py
```diff
@@ -1,61 +1,118 @@
...
+    The exception used for catching unexpected errors thrown from test cases and as a base class.
+    """
+
     def __str__(self) -> str:
         return str(super().__repr__())
 
 
-class MissingExceptException(ReportedException):
-    pass
+class RevertableException(BaseException):
```
MaksymilianDemitraszek, line 13: Wouldn't be better to inherit here from ` ReportedException` instead?

src/commands/test/test_environment_exceptions.py
```diff
@@ -1,61 +1,118 @@
...
+        return "\n".join(result)
+
+    def match(self, other: "RevertableException") -> bool:
+        return (self.error_type is None or self.error_type == other.error_type) and (
+            self.error_message is None
+            or self.error_message in (other.error_message or "")
+        )
+
+
+class StandardRevertableException(RevertableException):
```
MaksymilianDemitraszek, line 43: Wouldn't `StarknetRevertableException` or something similar be a better name?

src/commands/test/test_environment_exceptions.py
```diff
@@ -1,61 +1,118 @@
...
+    """
+    The exception commonly used to display errors encountered during test execution.
```
MaksymilianDemitraszek, line 45: We should mention here it's an abstraction over all errors raised in cairo runtime

src/commands/test/test_environment_exceptions.py
MaksymilianDemitraszek: Maybe `StandardCairoRuntimeRevertableException`? It gets too long...

tests/e2e/test_testing.py
```diff
@@ -42,5 +42,5 @@ def test_expect_revert(protostar, copy_fixture):
         "test_expect_revert.cairo::test_error_was_not_raised_before_stopping_expect_revert_fail_expected"
         in result
     )
-    assert "name: RANDOM_ERROR_NAME, message:" in result
+    assert "[error_type] RANDOM_ERROR_NAME" in result
```
MaksymilianDemitraszek, line 45: I think we should have integration tests to check the expected revert and expected events error messages specifically. Logic there is still complicated. You don't have to do this in this PR but please create a task for it. Would be nice to have cases checking if proper information is provided for:
- error mismatch
- error missing
- events mismatch

MaksymilianDemitraszek (changes requested):
It's great, much cleaner than before. Amazing job!!! Some questions/nitpick in the comments.

src/commands/test/test_environment_exceptions.py
MaksymilianDemitraszek: Being honest I have no idea

kasperski95: This exception is also raised from cheatcodes. That's why I avoided using `Starknet` in the name.

src/commands/test/runner.py
```diff
@@ -180,30 +167,45 @@ async def invoke_test_function(self, function_name: str):
...
+                details=ex.message,
+            ) from ex
+
+    def add_test_finish_hook(self, listener: Callable[[], None]) -> Callable[[], None]:
+        self._test_finish_hooks.add(listener)
+
+        def remove_hook():
+            self._test_finish_hooks.remove(listener)
+
+        return remove_hook
```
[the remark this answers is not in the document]
kasperski95: I am going to use it in `expectCall` and `expectEmit` where scope is important.

src/commands/test/test_environment_exceptions.py
kasperski95: After using RevertableException in cheatcodes changing the name to Starknet... makes sense.

tests/e2e/test_testing.py
kasperski95: This cheatcode can't throw a mismatch error, but I can make it print all events from Starknet State instead.

kasperski95: However, it's pointless because it's not possible to convert a selector to name.

kasperski95: https://github.com/software-mansion/protostar/issues/227

src/commands/test/runner.py
```diff
@@ -185,14 +187,12 @@ async def _call_test_function(self, function_name: str):
         try:
             func = getattr(self.test_contract, function_name)
             call_result = await func().invoke()
-            for hook in self._test_finish_hooks:
-                hook()
 
             if self._expected_error is not None:
                 raise ExpectedRevertException(self._expected_error)
```
MaksymilianDemitraszek, line 192: I would move also this

MaksymilianDemitraszek (approved):

PR #216, opened 2022-04-15
Improved tests performance

MaksymilianDemitraszek:


src/commands/test/contract_fork.py
```diff
@@ -0,0 +1,28 @@
...
+class ForkableStarknet(Starknet):
+    @classmethod
+    async def empty(
+        cls, general_config: Optional[StarknetGeneralConfig] = None
+    ) -> "Starknet":
+        return ForkableStarknet(
+            state=await StarknetState.empty(general_config=general_config)
+        )
+
+    def plug_from_different_state(self, deployed_contact: StarknetContract):
```
kasperski95, line 19:
```suggestion
    def copy_and_adapt_contract(self, starknet_contract: StarknetContract):
```
- [ ] fix typo: `contact` -> `contract`
- Does the contract have to be deployed?
- The name of this function is a bit confusing. I'd change it to `copy_and_adapt_contract`.

src/commands/test/contract_fork.py
```diff
@@ -0,0 +1,28 @@
+import copy
```
kasperski95, line 1: Consider changing the name of this file to `forkable_starknet.py`.

src/commands/test/runner.py
```diff
@@ -159,10 +158,24 @@ async def empty(
...
-        env.starknet = await Starknet.empty()
+        env.starknet = await ForkableStarknet.empty()
         env.test_contract = await env.starknet.deploy(contract_def=test_contract)
         return env
 
+    def fork(self):
+        assert self.starknet
+        assert self.test_contract
+
+        n_env = TestExecutionEnvironment(
```
kasperski95, line 169: What does the `n` mean in this case? `new`?

src/commands/test/runner.py
```diff
@@ -159,10 +158,24 @@ async def empty(
...
+            is_test_fail_enabled=self._is_test_fail_enabled,
```
kasperski95, line 170: I think the `is_test_fail_enabled` is no longer used.

kasperski95 (approved):
LGTM

src/commands/test/contract_fork.py
MaksymilianDemitraszek: Yea, name is terrible, your is much better, thanks. Contact doesn't have to be deployed, we copy state

src/commands/test/runner.py
MaksymilianDemitraszek: we can rename it to new

src/commands/test/runner.py
MaksymilianDemitraszek: True, merge conflicts

PR #217, opened 2022-04-15
Added test_performance util script

MaksymilianDemitraszek:
Not much to review here, right now it's just a util script for internal use, I will improve it in the future and maybe even run it on ci, but its not important for now.


kasperski95 (approved):
LGTM

PR #220, opened 2022-04-19
Add git version check

kasperski95:
- added git version check
- updated docs
- made versions result optional

### DIFF
--- a/src/utils/protostar_directory.py
+++ b/src/utils/protostar_directory.py
@@ -34,7 +34,7 @@ class VersionManager:
         self._protostar_directory = protostar_directory
 
     @property
-    def protostar_version(self) -> VersionType:
+    def protostar_version(self) -> Optional[VersionType]:
         path = (
             self._protostar_directory.directory_root_path
             / "dist"
@@ -47,11 +47,11 @@ class VersionManager:
                 version_s = tomli.loads(file.read())["tool"]["poetry"]["version"]
                 return VersionManager.parse(version_s)
         except FileNotFoundError:
-            getLogger().warning("Couldn't read Protostar version.")
-            return VersionManager.parse("0.0.0")
+            getLogger().warning("Couldn't read Protostar version")
+            return None
 
     @property
-    def cairo_version(self) -> VersionType:
+    def cairo_version(self) -> Optional[VersionType]:
         path = (
             self._protostar_directory.directory_root_path
             / "dist"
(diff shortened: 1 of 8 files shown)


src/utils/protostar_directory.py
```diff
@@ -46,24 +47,37 @@ def protostar_version(self) -> VersionType:
...
+        if type(output) is str:
+            return version.parse(output.replace("git version ", ""))
+        return None
 
     def print_current_version(self) -> None:
-        print(f"Protostar version: {self.protostar_version}")
-        print(f"Cairo-lang version: {self.cairo_version}")
+        print(f"Protostar version: {self.protostar_version or 'unknown'}")
+        print(f"Cairo-lang version: {self.cairo_version or 'unknown'}")
+        print(f"Git version: {self.git_version or 'unknown'}")
```
MaksymilianDemitraszek, line 83: I think we should not include git here, git has it's own command to check version, it's not a responsibility of protostar.

src/utils/protostar_directory.py
```diff
@@ -46,24 +47,37 @@ def protostar_version(self) -> VersionType:
...
+                return VersionManager.parse(version_s)
+        except FileNotFoundError:
+            getLogger().warning("Couldn't read cairo-lang version")
+            return None
+
+    @property
+    def git_version(self) -> Optional[VersionType]:
+        output = Git().execute(["git", "--version"])
+        # pylint: disable=unidiomatic-typecheck
+        if type(output) is str:
```
MaksymilianDemitraszek, line 76: why do not use isinstance here?

src/utils/protostar_directory.py
```diff
@@ -46,24 +47,37 @@ def protostar_version(self) -> VersionType:
                 version_s = tomli.loads(file.read())["tool"]["poetry"]["version"]
                 return VersionManager.parse(version_s)
         except FileNotFoundError:
-            getLogger().warning("Couldn't read Protostar version.")
-            return VersionManager.parse("0.0.0")
+            getLogger().warning("Couldn't read Protostar version")
+            return None
```
MaksymilianDemitraszek, line 51: Maybe we should re raise error here? Output could be useful for user to debug it in some cases WDYT?

MaksymilianDemitraszek (changes requested):
Looks good, but we should not include git version into protostar version check, it would be unidiomatic, also some minor nitpicks

src/utils/protostar_directory.py
kasperski95: There was a problem with type narrowing. I couldn't use `replace` command on the output in the line below.

src/utils/protostar_directory.py
kasperski95: This warning should never be visible to the user. It's more a change for us to be able to run `python ./protostar.py`.

src/utils/protostar_directory.py
kasperski95: I made this change to print everything what is required by the updated issue template. I'll modify the issue template instead.

PR #222, opened 2022-04-19
Deprecate stop expecting revert

kasperski95:


src/commands/test/runner.py
```diff
@@ -292,6 +294,7 @@ def expect_revert(self, expected_error: ExpectedError) -> Callable[[], None]:
 
         def stop_expecting_revert():
             if self._expected_error is not None:
+                logger.warning("The callback returned by the `expect_revert` is deprecated.")
                 raise MissingExceptException(
```
MaksymilianDemitraszek, line 298:
```suggestion
        def stop_expecting_revert():
            logger.warning("The callback returned by the `expect_revert` is deprecated, and will be removed in the future versions.")
            if self._expected_error is not None:
                raise MissingExceptException(
```

MaksymilianDemitraszek: We should always log it

MaksymilianDemitraszek (approved):

MaksymilianDemitraszek:
Should this point to master or development?

PR #224, opened 2022-04-20
Commands refactor + CLI docs generator + Configuring from `protostar.toml`

kasperski95:
- refactored the code between entry and commands and added unit tests
- added an option to configure commands from the configuration file
- added a possibility to generate reference docs


src/cli/cli_app.py
```diff
@@ -0,0 +1,29 @@
...
+        for command in self.commands:
+            self._command_mapping[command.name] = command
+
+    async def run(self, args: Any) -> bool:
+        if not args.command:
+            return False
+
+        command = self._command_mapping[args.command]
+        await command.run(args)
+        return True
```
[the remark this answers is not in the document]
kasperski95: I planned to do some action when no command was executed. Currently, if you type `protostar` nothing is returned.

src/commands/build/build_project.py
```diff
@@ -9,10 +9,56 @@
...
+            Command.Argument(
+                name="output",
+                description="An output directory that will be used to put the compiled contracts in.",
+                type="path",
+                default="build",
+            ),
+        ]
+
+    async def run(self, args):
+        build_project(
```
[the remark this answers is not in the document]
kasperski95: Maybe in one of the future PRs. I'd have to update code and tests for every command. This PR is already too big.

src/utils/protostar_directory.py
```diff
@@ -22,6 +22,13 @@ def protostar_binary_dir_path(self) -> Optional[Path]:
     def directory_root_path(self) -> Path:
         return self._protostar_binary_dir_path / ".." / ".."
 
+    def inject_protostar_cairo_dir(self, cairo_paths: List[Path]) -> List[Path]:
```
Arcticae, line 25: this method is a bit weird. it pushes into the list and returns it as well. maybe a list spread would be more appropriate? and then we could name it `add_protostar_cairo_dir` instead of `inject`, the naming is a bit confusing for me as well

src/protostar_cli.py
```diff
@@ -0,0 +1,100 @@
...
+        version_manager: VersionManager,
+    ) -> None:
+        self.project = project
+        super().__init__(
+            commands=[
+                InitCommand(script_root, version_manager),
+                BuildCommand(project),
+                InstallCommand(project),
+                RemoveCommand(project),
+                UpdateCommand(project),
```
Arcticae, line 49: We considered registration using a decorator, is that still on the table or is that impossible?

src/protostar_cli.py
```diff
@@ -0,0 +1,100 @@
...
+    ProtostarDirectory,
+    StandardLogFormatter,
+    VersionManager,
+    log_color_provider,
+)
+
+
+class ProtostarCLI(CLIApp):
+    @classmethod
+    def create(cls, script_root: Path):
```
Arcticae, line 29: please place it after \_\_init\_\_

src/main.py
```diff
@@ -0,0 +1,36 @@
+# pylint: disable=import-outside-toplevel
+import asyncio
+import sys
+from pathlib import Path
+
+
+def main(script_root: Path):
+    try:
+        from src.cli import ArgumentDefaultValueFromConfigProvider, ArgumentParserFacade
```
Arcticae, line 9: is top level import causing a import loop?

src/cli/argument_parser_facade.py
```diff
@@ -0,0 +1,129 @@
...
+    def _add_root_argument(self, argument: Command.Argument) -> "ArgumentParserFacade":
+        assert (
+            argument.is_positional is False
+        ), f"A root argument ({argument.name}) cannot be positional"
+
+        self._update_arg_default_value_if_necessary(None, argument)
+        ArgumentParserFacade._add_argument(self.argument_parser, argument)
+        return self
+
+    def _update_arg_default_value_if_necessary(
```
Arcticae, line 63:
```suggestion
    def _update_arg_default_value(
```

Arcticae: that method name seems confusing to me, namely i don't really know when it's necessary

src/cli/argument_parser_facade.py
```diff
@@ -0,0 +1,129 @@
...
+            ArgumentParserFacade._add_argument(command_parser, arg)
+
+        return self
+
+    def _add_root_argument(self, argument: Command.Argument) -> "ArgumentParserFacade":
+        assert (
+            argument.is_positional is False
+        ), f"A root argument ({argument.name}) cannot be positional"
+
+        self._update_arg_default_value_if_necessary(None, argument)
```
Arcticae, line 59:
```suggestion
        self._update_arg_default_value_if_necessary(argument=argument)
```

Arcticae: Or swap the places of args and use positional - up to you

src/cli/argument_parser_facade.py
```diff
@@ -0,0 +1,129 @@
...
+                *names,
+                help=argument.description,
+                action="store_true",
+                default=argument.default,
+            )
+            return argument_parser
+
+        arg_type = str
+
+        if argument.type == "directory":
```
Arcticae, line 100: can we make it as a defaultdict mapping?

src/cli/argument_parser_facade.py
```diff
@@ -0,0 +1,129 @@
...
+
+        names = [name]
+        if short_name:
+            names.append(short_name)
+
+        if argument.type == "bool":
+            assert (
+                argument.is_positional is False
+            ), "A boolean can't be positional argument"
+            assert argument.is_array is False, "Array of booleans is not allowed"
```
Arcticae, line 89: why is that?

Arcticae: not that i'm going to do anything like this, but wondering why - maybe a comment would be nice here

src/cli/cli_app.py
```diff
@@ -0,0 +1,29 @@
+from typing import Any, List, Mapping, Optional
+
+from src.cli.command import Command
+
+
+class CLIApp:
```
Arcticae, line 6: I'm wondering why there are 2 classes instead of just one - ProtostarCLI. I wouldn't bother with dividing those if they aren't re-used anywhere

src/cli/reference_docs_generator_test.py
```diff
@@ -0,0 +1,141 @@
...
+
+def test_generating_markdown_for_command_arguments(foo_command: FooCommand):
+    docs_generator = ReferenceDocsGenerator(
+        CLIApp(commands=[foo_command], root_args=[])
+    )
+
+    result = docs_generator.generate_cli_reference_markdown()
+    splitted_result = result.split("\n")
+
+    assert f"### `{foo_command.name}`" in splitted_result
```
Arcticae, line 36: I wouldn't check the .md formatting, we might want to change it up in the future and testing for content only seems more intuitive and less error-prone

src/commands/build/build_project.py
```diff
@@ -9,10 +9,56 @@
...
+        return "Compile contracts."
+
+    @property
+    def example(self) -> Optional[str]:
+        return "$ protostar build"
+
+    @property
+    def arguments(self) -> List[Command.Argument]:
+        return [
+            Command.Argument(
```
Arcticae, line 37: Can we do typed dicts instead of `Command.Argument`, and wrap it underneath?

src/commands/init/init.py
```diff
@@ -1,15 +1,49 @@
...
 from src.utils.config.project import Project, ProjectConfig
 
 
+class InitCommand(Command):
+    def __init__(self, script_root: Path, version_manager: VersionManager) -> None:
+        super().__init__()
+        self._script_root = script_root
+        self._version_manager = version_manager
+
+    @property
```
Arcticae, line 20: Those could be defined as properties on the class, right? We don't really do any dynamic calculation in command definition

src/commands/init/init.py
```diff
@@ -1,15 +1,49 @@
...
+                description="Adapt current directory to a Protostar project.",
+                type="bool",
+            )
+        ]
+
+    async def run(self, args):
+        get_creator(args)(self._script_root, self._version_manager).run()
+
+
+# TODO: remove me
```
Arcticae, line 46: Is there a task associated with that?

src/commands/init/init.py
Arcticae: Or perhaps a part of those properties, definition would be more terse as a result

src/commands/install/install_command.py
```diff
@@ -0,0 +1,81 @@
...
+                description="A custom package name. Use it to resolve name conflicts.",
+                type="str",
+            ),
+        ]
+
+    async def run(self, args):
+        handle_install_command(args, self._project)
+
+
+def handle_install_command(args: Any, project: Project) -> None:
```
Arcticae, line 56: i don't see a reason why it'd be defined in a separate method

Arcticae (changes requested):

src/utils/protostar_directory.py
kasperski95: `inject` is not the best choice. I'll change it to `add_protostar_cairo_dir_if_necessary`.

src/protostar_cli.py
kasperski95: In this case decorators are not the best choice because we need to pass dependencies (Project, VersionManager etc.).

src/cli/argument_parser_facade.py
kasperski95: I'll change it to be more explicit. The default values are loaded from the config file.

src/cli/argument_parser_facade.py
kasperski95: I think `argparse` complained about that, or I added this restriction to avoid creating such CLI. I'll improve assert descriptions.

src/cli/cli_app.py
kasperski95: It's easier to test `ReferenceDocsGenerator` and `ArgumentParserFacade`.

src/commands/build/build_project.py
kasperski95: I'd rather avoid `TypedDict` because they make refactoring (renaming specifically) harder. Why do you want to use `TypedDict`?

src/commands/init/init.py
kasperski95: I couldn't find a way to create `@abstractproperty` that is not a method. 
The alternative approach is to pass all information through `super().__init__(...)`.

src/commands/install/install_command.py
kasperski95: Avoiding changing more code in this PR. I'd have to update the code and tests for every function otherwise.

src/commands/init/init.py
MaksymilianDemitraszek: I actually like this approach, maybe is a bit `Java`, but quite readable

src/cli/cli_app.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: I don't see why it's a reason to return `True` here

src/main.py
MaksymilianDemitraszek: Why it does?

src/commands/install/install_command.py
MaksymilianDemitraszek: Please create a task for inlining this methods then

scripts/generate_reference_docs.py
```diff
@@ -0,0 +1,12 @@
+from pathlib import Path
```
MaksymilianDemitraszek, line 1: Please move this reference scripts into separate directory, they should not be a part of the main codebase

src/cli/argument_parser_facade.py
```diff
@@ -0,0 +1,129 @@
...
+from abc import ABC, abstractmethod
+from argparse import ArgumentParser
+from pathlib import Path
+from typing import Any, Optional, Sequence
+
+from src.cli.cli_app import CLIApp
+from src.cli.command import Command
+
+
+class ArgumentDefaultValueProvider(ABC):
```
MaksymilianDemitraszek, line 11: Why do we need this base class? It seems only one class implements it

kasperski95: I used Delegate Pattern here. I wanted to avoid making `ArgumentParserFacade` dependent on the `Project` class.

MaksymilianDemitraszek: I understand why we have `ArgumentDefaultValueFromConfigProvider`, we may need some mapping between project and facade, which is fine. I don't understand the purpose of an abstract base class, imo it's unnecessary abstraction for now, we can isolate it later when we will have more ValueProviders

src/main.py
kasperski95: Idk. GitPython throws an error during importing. I've just noticed this error isn't caught here, so I'll move try/except to the `protostar.py`.

However, I'll leave importing `ArgumentParserFacade` here because import patching in `main_test.py` stops working when top-level import is used and I didn't figure it out why it behaves like that.

src/cli/argument_parser_facade.py
kasperski95: This abstract class was meant to be an interface. This class is implemented in the test file (`argument_parser_facade_test`).

src/commands/install/install_command.py
kasperski95: https://github.com/software-mansion/protostar/issues/241

src/commands/init/init.py
kasperski95: https://github.com/software-mansion/protostar/issues/241

src/cli/cli_app.py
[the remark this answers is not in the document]
kasperski95: It passed previous check(s) and a command was found and executed that's why this function returns `True` here.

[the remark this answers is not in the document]
MaksymilianDemitraszek: We should use exceptions instead of error codes

src/protostar_cli.py
MaksymilianDemitraszek: I dont understand why it makes using decorators a problem, you can register classes instead of objects

src/cli/cli_app.py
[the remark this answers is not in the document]
kasperski95: I hesitated using exception because `protostar --version` is a valid usage.
I'll use an exception to make the code easier to read.

src/main.py
MaksymilianDemitraszek: Isn't the same issue with name shadowing we had once already? I think we have to figure it out instead of making dirty workaround. If you want my help with it let me know

src/protostar_cli.py
MaksymilianDemitraszek: You can do it in different PR though

src/main.py
kasperski95: Yep, I need help with it. It looks like a different issue. I spent already too much time on it and ran out of ideas. I think we should try to fix it in a different PR.

src/cli/argument_parser_facade.py
MaksymilianDemitraszek: Python is duck typed, interfaces exist only for documentation reasons. Having an interface makes sense when you have multiple implementations of it. Here we introduce confusing and premature abstraction which makes code harder to comprehend. We have nothing to generalize. Please remove the interface and use class directly. For test you can use mock.

src/protostar_cli.py
kasperski95: How am I supposed to provide dependencies to the registered classes without using some DI framework?

src/main.py
MaksymilianDemitraszek: Ok, please finish the requested changes and I will take a look on it

src/cli/argument_parser_facade.py
kasperski95: The point of this class is not an abstraction nor generalization, but making the code loosely coupled. I wanted to make the `cli` package completely independent of the `project`.

Fine, I'll remove it.

src/cli/argument_parser_facade.py
MaksymilianDemitraszek: Is this resolved?

src/commands/build/build_project.py
```diff
@@ -9,10 +9,56 @@
...
+                name="cairo-path",  # TODO: use plural name
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: Is this resolved?

src/utils/config/project_test.py
```diff
@@ -64,10 +82,7 @@ def test_handling_not_supported_version(version_manager: VersionManager):
...
 
 
-def test_no_project_found(version_manager: VersionManager):
-    proj = Project(
-        version_manager,
-        project_root=Path(current_directory, "examples"),
-    )
+def test_no_project_found(version_manager: VersionManager, tmpdir):
+    proj = Project(version_manager, Path(tmpdir))
     with pytest.raises(NoProtostarProjectFoundError):
```
MaksymilianDemitraszek, line 87: We probably should be consistent with naming, `Error`/`Exception`, not a matter of this PR though.

src/commands/remove/remove_command.py
```diff
@@ -0,0 +1,67 @@
...
+                name="package",
+                description=INTERNAL_DEPENDENCY_REFERENCE_DESCRIPTION,
+                type="str",
+                is_required=True,
+                is_positional=True,
+            ),
+        ]
+
+    async def run(self, args):
+        handle_remove_command(args, self._project)
```
MaksymilianDemitraszek, line 48: Do we have task for inlining this methods?

src/commands/remove/remove_package.py
```diff
@@ -6,6 +6,7 @@
 from src.commands.remove import removal_exceptions
 
 
+# TODO: https://github.com/software-mansion/protostar/issues/241
```
MaksymilianDemitraszek, line 9: We do we need this comment? We captured issue already

MaksymilianDemitraszek: I bet we will close the issue and it will remain :p

.pylintrc
```diff
@@ -319,7 +319,7 @@ ignore-imports=no
 ignore-signatures=no
 
 # Minimum lines number of a similarity.
-min-similarity-lines=4
+min-similarity-lines=20
```
MaksymilianDemitraszek, line 322: Why it has been triggered?

src/__init__.py
```diff
@@ -1,4 +1,11 @@
-from src.cli import cli
-from src.commands import handle_install_command, removal_exceptions, remove, upgrade
-from src.protostar_exception import ProtostarException
-from src.utils.config import Project
+# Python complains about importing `Project`` if this line is removed
```
MaksymilianDemitraszek, line 1: What line this comment describes?

src/cli/argument_default_value_from_config_provider.py
```diff
@@ -0,0 +1,21 @@
...
+from src.utils import Project
+
+
+class ArgumentDefaultValueFromConfigProvider:
+    def __init__(self, project: Project) -> None:
+        self._project = project
+        super().__init__()
+
+    def get_default_value(
+        self, command: Optional[Command], argument: Command.Argument
```
MaksymilianDemitraszek, line 13: I think it should be 
```suggestion
        self, command: Command, argument: Command.Argument
```
We should never request data from shared config it should be only a fallback. It will make interface more defensive

MaksymilianDemitraszek: Also we should have separate section for root parameters. Keeping them inside shared config is misleading

src/cli/command.py
```diff
@@ -0,0 +1,58 @@
...
+
+from attr import dataclass
+from typing_extensions import Literal
+
+InputAllowedType = Literal["str", "directory", "path", "bool", "regexp"]
+
+
+class Command(ABC):
+    @dataclass
+    class Argument:
```
MaksymilianDemitraszek, line 14: Please make those two frozen

src/cli/argument_parser_facade.py
MaksymilianDemitraszek: I would name it `update_from_config` default is misleading

src/cli/argument_default_value_from_config_provider.py
```diff
@@ -0,0 +1,21 @@
+from typing import Any, Optional
+
+from src.cli.command import Command
+from src.utils import Project
+
+
+class ArgumentDefaultValueFromConfigProvider:
```
MaksymilianDemitraszek, line 7: We we use `Default` in the name here? I think it's a bit misleading in connection with python defaults.
`ArgumentValueFromConfigProvider` would be better

src/cli/argument_default_value_from_config_provider.py
MaksymilianDemitraszek: We can make two methods
```
def get_command_config_value( self, command: Command, argument: Command.Argument)
def get_root_config_value( self, argument: Command.Argument)
```
Interface would be simpler and more defensive

src/cli/argument_parser_facade.py
```diff
@@ -0,0 +1,126 @@
...
+                command, argument
+            )
+            if new_default is not None:
+                argument.default = new_default
+        return argument
+
+    @staticmethod
+    def _add_argument(
+        argument_parser: ArgumentParser, argument: Command.Argument
+    ) -> ArgumentParser:
```
MaksymilianDemitraszek, line 74: Don't return if it mutates, it's misleading and considered bad practice

MaksymilianDemitraszek (changes requested):
Good job, close to being merge, few more comments

src/protostar_cli.py
MaksymilianDemitraszek: Python is really dynamic so it won't be hard, not a concern of this PR though. Can you make task for this?

kasperski95: It's overengineering at this point.
https://github.com/software-mansion/protostar/issues/245

tests/e2e/test_building.py
```diff
@@ -4,23 +4,9 @@
 import pytest
```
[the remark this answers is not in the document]
kasperski95: It's already there — `test_cairo_path_loaded_from_command_config_section_in_config_file`

src/commands/remove/remove_command.py
kasperski95: The task is created. The TODO pointing to that task is in the `handle_remove_command.py`

src/protostar_cli.py
MaksymilianDemitraszek: Agreed

src/commands/remove/remove_package.py
kasperski95: This command should be in the `remove_command.py` file above `handle_remove_command`. I'll move it there.

src/__init__.py
kasperski95: The line below. I change `line` to `import`.

.pylintrc
kasperski95: By explicit command imports. This rule gives too many false flags, and it can't be ignored from the file level.

src/__init__.py
MaksymilianDemitraszek: "if import below is removed"

tests/e2e/test_building.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: I mean `testing` not `building`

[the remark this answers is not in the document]
kasperski95: We can, there are more important things to do.

src/cli/argument_default_value_from_config_provider.py
kasperski95: Is it misleading? Root args (--no-color) are shared across commands. The only exception is the `--version`, which behaves like command.

> We should never request data from shared config it should be only a fallback. It will make interface more defensive
I don't get it. It's already used as a fallback.

tests/e2e/test_building.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: Nope, testing is much more important than new features

src/cli/argument_default_value_from_config_provider.py
MaksymilianDemitraszek: Giving it afterthought leave it as it is

tests/e2e/test_building.py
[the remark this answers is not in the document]
kasperski95: There are unit tests for loading configs from the config file, and there's e2e test that tests loading stuff from the config file. Adding another e2e test that checks loading from parameters from the config file doesn't add much value.

src/cli/argument_default_value_from_config_provider.py
kasperski95: The config file changes the default value of argument. However, this class shouldn't care how it's used, therefore I'll rename it.

tests/e2e/test_building.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: But the code injecting cairo_path is different for tests, having it for building doesn't cover it. 

Unit tests are not enough, we need to test stuff e2e. I had at least few situations when unit tests were passing and e2e tests catched it (the error we had had with all tests passing is an example).

Please add one or two tests, it's not much code and is extremely useful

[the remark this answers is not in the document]
kasperski95: Fine

src/cli/command.py
kasperski95: I've made the `Argument` class Frozen. I'll leave `Command` class as it is because it doesn't have attributes.

tests/e2e/test_testing.py
```diff
@@ -44,3 +45,24 @@ def test_expect_revert(protostar, copy_fixture):
     )
     assert "[error_type] RANDOM_ERROR_NAME" in result
     assert "Unknown location" not in result
+
+
+def test_loading_cairo_path_from_config_file(protostar, my_private_libs_setup):
+    (my_private_libs_dir,) = my_private_libs_setup
+
+    # TODO: Protostar shouldn't crash when imports are incorrect
+    # https://github.com/software-mansion/protostar/issues/230
```
MaksymilianDemitraszek, line 54: Please remove this TODO, we such keep such stuff in the issue tracker. You can add permalink there

MaksymilianDemitraszek (approved):
LGTM, i will take a look on the import issue in a different PR

src/commands/build/build_project.py
```diff
@@ -1,18 +1,65 @@
...
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: The name is derived from the unix PATH not path, which can contain multiple paths, you provide arguments to add it to cairo-path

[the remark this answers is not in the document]
kasperski95: The name comes from this doc:
https://www.cairo-lang.org/docs/how_cairo_works/imports.html?highlight=cairo_path

Our CLI expects paths to be separated by space, whereas cairo-compile's `cairo_path` expects a single string.

PR #231, opened 2022-04-22
Fixed not failing tests

MaksymilianDemitraszek:


src/commands/test/run_test_runner.py
```diff
@@ -21,7 +21,7 @@ async def run_test_runner(
     cairo_path = cairo_paths or []
     include_paths = [str(pth) for pth in cairo_path]
 
-    reporter = TestReporter(tests_root)
+    reporter = reporter if reporter else TestReporter(tests_root)
```
kasperski95, line 24: Consider updating the reporter type to: `Optional[TestReporter]`

kasperski95 (approved):
LGTM

PR #232, opened 2022-04-22
Tests concurrency

MaksymilianDemitraszek:
- [x] Fix deadlock problems
- [x] Add timeouts

### DIFF
--- a/src/commands/test/reporter.py
+++ b/src/commands/test/reporter.py
@@ -1,12 +1,24 @@
-from functools import reduce
+from collections import defaultdict
 from pathlib import Path
+from tkinter import W
 from typing import Dict, List, Optional, Union
+from tqdm import tqdm as bar
 
 from src.commands.test.cases import BrokenTest, FailedCase, PassedCase
 from src.commands.test.utils import TestSubject
 
 CaseResult = Union[PassedCase, FailedCase, BrokenTest]
 
+class Reporter:
+
+    def __init__(self, queue):
+        self.reports_queue = queue
+
+    def report(self, subject: TestSubject, case_result: CaseResult):
+        self.reports_queue.put((subject, case_result))
+    
+    def file_entry(self, file_name: str):
+        pass
 
 class TestReporter:
     _collected_count: Optional[int]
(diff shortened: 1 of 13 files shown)


src/commands/test/test_environment_exceptions.py
```diff
@@ -73,6 +76,14 @@ def __str__(self) -> str:
...
         return "\n".join(result)
 
+    def __reduce__(self):
+        return type(self), (
+            self.error_message,
+            self.error_type,
+            self.code,
+            self.details,
+        )
+
```
kasperski95, line 86: Why do we need this?

kasperski95 (approved):
LGTM. Can you increase test coverage (`codecov/project — 75.26% (-9.37%) `)?

src/commands/test/reporter.py
```diff
@@ -1,69 +1,112 @@
...
-    broken_tests: List[BrokenTest]
-    failed_cases: List[FailedCase]
-    passed_cases: List[PassedCase]
-    failed_tests_by_subject: Dict[Path, List[FailedCase]]
-    collected_subjects: List[TestSubject]
-    tests_root: Path
+class ResultReport(Enum):
+    PASSED_CASE = 1
+    FAILED_CASE = 2
+    BROKEN_CASE = 3
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: It is shared between two classes

src/commands/test/reporter.py
```diff
@@ -1,69 +1,112 @@
...
+
+    @property
+    def failed_count(self):
+        return len(self.failed_cases)
+
+
+class ReportCollector:
+    _collected_count: Optional[int]
+    collected_subjects: List[TestSubject]
+    tests_root: Path
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: Leftover

src/commands/test/test_environment_exceptions.py
MaksymilianDemitraszek: for [pickling](https://docs.python.org/3/library/pickle.html)

kasperski95: Isn't pickling supported by default?

kasperski95: > In most cases, no additional code is needed to make instances picklable. By default, pickle will retrieve the class and the attributes of an instance via introspection.

https://docs.python.org/3/library/pickle.html#pickling-class-instances

MaksymilianDemitraszek: There were some issues because with how we inherit from error class. We would have to make arguments optional. I somewhat prefer this explicit aproach. We can revisit it later, but seems not very painful/hacky for now.

PR #239, opened 2022-04-28
Refine testing output

kasperski95:
- pretty print test results
- test results are logged as soon as they are ready
- handled test interrupting without yeeting a stack trace
- test summary prints info about test suits (files) and test cases
- test summary includes skipped tests (in case a user interrupts tests)

### All cases pass
https://user-images.githubusercontent.com/9629577/165948545-9d16821a-8fa4-4f23-9752-627f0e4df343.mov

### Some cases fail
https://user-images.githubusercontent.com/9629577/165948585-c69dfcad-757d-457c-80bb-43de8e72de36.mov


src/commands/test/run_test_runner.py
```diff
@@ -37,23 +40,34 @@ async def run_test_runner(
...
+                pool.join()
+                return []
+
+
+def worker_init():
+    def signal_handler(_signal_num, _frame):
+        return None
 
-        return results
+    signal.signal(signal.SIGINT, signal_handler)
```
MaksymilianDemitraszek, line 70: ctrl+c does not have to guarantee gracefull handling. In my opion we should leave it as it was, this solution introduces unnecessary layer of complexity.

MaksymilianDemitraszek (changes requested):
Very nice changes some comments below

src/commands/test/run_test_runner.py
kasperski95: Ctrl + c will often be used to stop tests. We must handle it somehow. Currently, Protostar generate a lot of noise when tests are interrupted.

src/commands/test/run_test_runner.py
```diff
@@ -37,23 +40,34 @@ async def run_test_runner(
...
             for subject in test_subjects
         ]
 
-        with Pool(multiprocessing.cpu_count()) as pool:
+        with Pool(multiprocessing.cpu_count(), worker_init) as pool:
             result = pool.starmap_async(run_worker, setups)
-            reporter.live_reporting()
-            results = result.get()
-            reporter.report_summary(results)
+            try:
```
MaksymilianDemitraszek, line 57: We probably can make something like this:
```
        try:
            with Pool(multiprocessing.cpu_count(), worker_init) as pool:
                result = pool.starmap_async(run_worker, setups)
                reporter_coordinator.live_reporting()
                return result.get()
        except KeyboardInterrupt:
                return []


def worker_init():
    signal.signal(signal.SIGINT, signal.SIG_IGN)
```

MaksymilianDemitraszek (approved):

PR #243, opened 2022-05-04
Check for being the entry point when executing the CLI

### LINKED ISSUE #242: Do not run the CLI if main module is imported
Some projects might want to use Protostar python modules. Calling `import protostar` currently executes the CLI command since there is no check for being the entry point.

Bernardstanislas:
Closes #242

### DIFF
--- a/protostar.py
+++ b/protostar.py
@@ -166,13 +166,14 @@ try:
         required=False,
     )
 
-    try:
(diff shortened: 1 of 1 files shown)


kasperski95:
https://github.com/software-mansion/protostar/pull/224#discussion_r859895754

PR #248, opened 2022-05-05
Autodetect `.git` directory

kasperski95:


src/utils/config/project.py
```diff
@@ -42,6 +42,20 @@ def __init__(
         self._version_manager = version_manager
         self._config_dict: Optional[Dict[str, Any]] = None
 
+    @property
+    def repo_path(self) -> Optional[Path]:
```
MaksymilianDemitraszek, line 46: It seems, Repo supports `search_parent_directories` flag, we don't need our own implementation

MaksymilianDemitraszek (changes requested):
Works but we shouldn't implement it by ourselves

MaksymilianDemitraszek (approved):

PR #249, opened 2022-05-06
Simplify test file generated by `protostar init`

kasperski95:
- used unit test workflow in the generated test file

### DIFF
--- a/templates/default/tests/test_main.cairo
+++ b/templates/default/tests/test_main.cairo
@@ -15,13 +15,13 @@ func test_increasing_balance{syscall_ptr : felt*, range_check_ptr, pedersen_ptr 
 end
 
 @view
-func test_revert_when_negative_value_is_provided{
+func test_failing_when_increasing_balance_with_negative_value{
     syscall_ptr : felt*, range_check_ptr, pedersen_ptr : HashBuiltin*
 }():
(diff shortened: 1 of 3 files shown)


templates/default/tests/test_main.cairo
```diff
@@ -1,25 +1,28 @@
...
+    let (result_after) = balance.read()
+    assert result_after = 42
+    return ()
 end
 
-@external
-func test_example{syscall_ptr : felt*, range_check_ptr}():
-    alloc_locals
+@view
+func test_failing_when_increasing_balance_with_negative_value{
```
MaksymilianDemitraszek, line 18:
```suggestion
func test_revert_when_increasing_balance_with_negative_value{
```
failing may suggest test suppose to fail

MaksymilianDemitraszek (approved):

templates/default/tests/test_main.cairo
kasperski95: `test_revert` may suggest a test suppose to revert

I think I'll use `testCannot` (recommended by Foundry).

> A good practice is to use something like `testCannot` in combination with the [expectRevert](https://book.getfoundry.sh/cheatcodes/expect-revert.html) cheatcode

PR #250, opened 2022-05-06
Refactor test command's code

kasperski95:
- refactored the testing code to follow SOLID principles
- normalized naming
- moved large classes to separate files
- moved cheatcodes tests to `tests/integration`


src/commands/test/test_cases.py
```diff
@@ -45,7 +45,7 @@ def __str__(self) -> str:
 
 
 @dataclass(frozen=True)
-class BrokenTest(CaseResult):
+class BrokenTestFile(TestCaseResult):
```
MaksymilianDemitraszek, line 48: much better name

MaksymilianDemitraszek: maybe `Suite`? I find it kinda problematic we use both names, we should settle for one of them

src/commands/test/test_collector.py
```diff
@@ -1,93 +1,128 @@
 import os
```
MaksymilianDemitraszek, line 1: I think collector refactor and related tests could definitely be a separate PR

src/commands/test/test_subject_queue.py
```diff
@@ -0,0 +1,32 @@
...
+    test_functions: List[StarknetCompiler.AbiElement]
+
+
+class TestSubjectQueue:
+    def __init__(
+        self, shared_queue: "queue.Queue[Tuple[TestSubject, TestCaseResult]]"
+    ) -> None:
+        self._shared_queue = shared_queue
+
+    def dequeue(self) -> Tuple[TestSubject, TestCaseResult]:
```
MaksymilianDemitraszek, line 28: put, get are idiomatically used in Python queues

src/commands/test/test_scheduler.py
```diff
@@ -0,0 +1,51 @@
...
+
+            try:
+                with multiprocessing.Pool(
+                    multiprocessing.cpu_count(),
+                    lambda: signal.signal(
+                        signal.SIGINT, signal.SIG_IGN
+                    ),  # prevents showing a stacktrace on cmd/ctrl + c
+                ) as pool:
+                    pool.starmap_async(self._worker, setups)
+                    self._live_logger.log(testing_queue, test_collector_result)
```
MaksymilianDemitraszek, line 49: There is a problem here, what if process raises an unexpected error? We need to get result here, even when unused

src/commands/test/test_subject_queue.py
```diff
@@ -0,0 +1,32 @@
...
+from src.utils.starknet_compilation import StarknetCompiler
+
+if TYPE_CHECKING:
+    import queue
+
+
+@dataclass
+class TestSubject:
+    """
+    A dataclass consisting of identification of a single test bundle, and target functions
```
MaksymilianDemitraszek, line 15: bundle is a third name we use, please commit to one

src/utils/starknet_compilation.py
```diff
@@ -72,3 +73,18 @@ def compile_contract(
         )
         assert isinstance(assembled, ContractDefinition)
         return assembled
+
+    class AbiElement(TypedDict):
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: Would prefer just using cairo internal object (dict?)

src/commands/test/test_environment.py
```diff
@@ -9,102 +8,24 @@
...
-                    case_result=FailedCase(
-                        file_path=test_subject.test_path,
-                        function_name=function["name"],
-                        exception=err,
-                    ),
-                )
-
-
-class TestExecutionEnvironment:
+class TestEnvironment:
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: Only environment can be misleading, does it has something to do with evironmental variables?

src/commands/test/asserts/asserts_test.py
```diff
@@ -1,21 +1,11 @@
-from pathlib import Path
+# @pytest.mark.asyncio
```
MaksymilianDemitraszek, line 1: Please make this work in ths PR

src/commands/test/test_collector.py
MaksymilianDemitraszek: Maybe if it's not a problem you can split it into two, would make it much easier to review...

src/commands/test/test_scheduler.py
MaksymilianDemitraszek: We probably have to make timout it live smaller, but im not sure how small. Not a matter of this pr though

src/utils/starknet_compilation.py
```diff
@@ -72,3 +73,18 @@ def compile_contract(
...
+        name: str
+        type: Literal["function"]
+        inputs: List[Any]
+        outputs: List[Any]
+        stateMutability: NotRequired[str]
+
+    def get_functions(self, cairo_file_path: Path, prefix: str) -> List[AbiElement]:
+        preprocessed = self.preprocess_contract(cairo_file_path)
```
MaksymilianDemitraszek, line 85: It would better fit inside TestCase class

MaksymilianDemitraszek: Probably we want to move compilation to collector in the future (to prevent changing cairo code when tests are running), so keep that in mind, this desing will make this change a bit hard.

src/commands/test/test_command_test.py
```diff
@@ -0,0 +1,59 @@
...
+        ]
+    )
+
+
+def test_test_command_name(mocker: MockerFixture):
+    test_command = TestCommand(mocker.MagicMock(), mocker.MagicMock())
+
+    assert test_command.name == "test"
+    assert len(test_command.description) > 3
+    assert test_command.example
```
MaksymilianDemitraszek, line 50: This test is useless, we are just checking hardcoded data, please remove it introduces unnecessary burden.

src/commands/test/test_command_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_arguments_match_corresponding_dataclass(mocker: MockerFixture):
```
MaksymilianDemitraszek, line 53: This test is useless, we are just checking hardcoded data, please remove it introduces unnecessary burden.

src/commands/test/test_command_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.mark.asyncio
+async def test_test_command_runs_scheduler_properly(mocker: MockerFixture):
+    args = TestCommand.Args(
+        target=Path("foo"),
+        match=re.compile("foo"),
+        omit=re.compile("bar"),
+        cairo_path=[Path() / "baz"],
+    )
+    TestCollectorMock = mocker.patch(
```
MaksymilianDemitraszek, line 22: Why do we need to mock here 🤔

MaksymilianDemitraszek: This way we don't even check if correct arguments are passed, much much better would be to let scheduler run some test

src/commands/test/test_collector_test.py
```diff
@@ -0,0 +1,185 @@
...
+    tmp_bar_path.mkdir(exist_ok=True, parents=True)
+    (tmp_bar_path / "test_bar.cairo").touch()
+    tmp_foo_path = project_root / "foo"
+    tmp_foo_path.mkdir(exist_ok=True, parents=True)
+    (tmp_foo_path / "test_foo.cairo").touch()
+
+
+@pytest.fixture(name="starknet_compiler")
+def starknet_compiler_fixture(mocker: MockerFixture):
+    starknet_compiler_mock = mocker.MagicMock()
```
MaksymilianDemitraszek, line 43: This is a good use of mock

src/commands/test/test_collector_test.py
```diff
@@ -0,0 +1,185 @@
...
+            - test_bar.cairo
+        - foo
+            - test_foo.cairo
+    """
+    tmp_bar_path = project_root / "bar"
+    tmp_bar_path.mkdir(exist_ok=True, parents=True)
+    (tmp_bar_path / "test_bar.cairo").touch()
+    tmp_foo_path = project_root / "foo"
+    tmp_foo_path.mkdir(exist_ok=True, parents=True)
+    (tmp_foo_path / "test_foo.cairo").touch()
```
MaksymilianDemitraszek, line 38: We could name it `bar_test` for better coverge

src/commands/test/test_collector_test.py
```diff
@@ -0,0 +1,185 @@
+import re
```
MaksymilianDemitraszek, line 1: Test cases in this module are very good 🚀

src/commands/test/test_collector_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+def test_logging_no_cases_found(mocker: MockerFixture):
+    logger_mock = mocker.MagicMock()
+
+    TestCollector.Result(
+        test_subjects=[],
+        test_cases_count=0,
+    ).log(logger_mock)
+
+    cast(MagicMock, logger_mock.warn).assert_called_once_with("No cases found")
```
MaksymilianDemitraszek, line 185: test_logging_preprocessing_error is probably missing

MaksymilianDemitraszek (changes requested):
Mostly good changes, some comments for now, but still all of the tests have to be finished
Can we also fix all suits typos here?

src/commands/test/test_collector_test.py
MaksymilianDemitraszek: I'm not sure if collector handles this, but just wanted to mention I don't know what will happen here

src/commands/test/test_runner.py
```diff
@@ -0,0 +1,113 @@
...
+        include_paths: Optional[List[str]] = None,
+    ):
+        self.queue = queue
+        self.include_paths = []
+
+        if include_paths:
+            self.include_paths.extend(include_paths)
+
+    @dataclass
+    class WorkerArgs:
```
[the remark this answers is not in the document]
kasperski95: I can bring it back, but I rather have type-safe and easy to maintain rather than pytonic code.

src/commands/test/test_environment.py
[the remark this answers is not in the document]
kasperski95: I changed it to match the file name. I'll change the file name instead.

src/commands/test/test_runner.py
```diff
@@ -0,0 +1,113 @@
...
+
+from src.commands.test.starkware_patch import CheatableSysCallHandler
+from src.commands.test.test_cases import BrokenTestFile, FailedTestCase, PassedTestCase
+from src.commands.test.test_environment_exceptions import ReportedException
+from src.commands.test.test_environment import TestEnvironment
+from src.commands.test.test_subject_queue import TestSubject, TestSubjectQueue
+from src.utils.modules import replace_class
+from src.utils.starknet_compilation import StarknetCompiler
+
+current_directory = Path(__file__).parent
```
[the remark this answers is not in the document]
kasperski95: Not anymore.

src/commands/test/test_runner.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: For this one acctualy im okay with this class, because it's fir Scheduler and i didn't know at the time I wrote this comment. You can change `starmap` to `map` though.

src/commands/test/test_command_test.py
kasperski95: I don't want to mix unit and integration tests if it's possible. I'll use `assert_called_once_with` to check the args.
Running actual tests will make tests slower, and it won't check if all arguments are passed properly.

src/commands/test/test_collector_test.py
kasperski95: `test_breakage_upon_broken_test_file` checks if `TestCollectingException` is raised. Logging this error is not a responsibility of this module.

src/commands/test/test_command_test.py
MaksymilianDemitraszek: I thought we are testing TestScheduler here, my mistake 😅

src/commands/test/test_collector.py
kasperski95: I'll try. I should've used bottom-up approach to refactoring.

src/commands/test/test_command.py
```diff
@@ -66,13 +72,32 @@ def arguments(self) -> List[Command.Argument]:
...
-        await run_test_runner(
-            args.target,
-            project=self._project,
-            omit=args.omit,
-            match=args.match,
-            cairo_paths=self._protostar_directory.add_protostar_cairo_dir(
-                args.cairo_path
-            ),
+    @dataclass
+    class Args:
```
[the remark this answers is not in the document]
kasperski95: I created this dataclass from the following reasons:
- creating args in `test_command_test.py`
- type-safety
- IDE autocompletion

src/commands/test/test_command_test.py
kasperski95: Yep, I've added it to increase coverage.

src/commands/test/test_command_test.py
kasperski95: This test is not useless. This test checks whether TestCommand.Args and CLI arguments are in sync.

src/commands/test/test_scheduler.py
kasperski95: If the process throws unexpected error, Protostar should crash. It should be a rare case, therefore we shouldn't care about logging results.

src/utils/starknet_compilation.py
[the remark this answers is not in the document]
kasperski95: > it's very cairo implementation dependent,
Yet, we are messing with its internals in various places. I created this type to avoid debugging this object every time we do something with it.

I'll make this type more obscure by leaving fields that are actually used and add proper comment to this type.

src/utils/starknet_compilation.py
kasperski95: I think you meant `TestSubject`. It's actually a better class for this method.

`Probably we want to move compilation to collector in the future (to prevent changing cairo code when tests are running), so keep that in mind, this desing will make this change a bit hard.`

Collector shouldn't work directly on `StarknetPreprocessedProgram`. Compiling at the very beginning will increase initial delay, and we're already dealing with this problem. I am not sure whether other test runners allow modifying test files during running tests. Maybe we should simply copy files when the test command starts.

src/commands/test/test_subject_queue.py
kasperski95: IMO `get` is not obvious whether it modifies the queue or not. I'll rename it to match Python though.

src/utils/starknet_compilation.py
kasperski95: I tried to move `get_functions` to `TestSubject` and I changed my mind. `get_functions` should remain in the `StarknetCompiler` to avoid exposing `StarknetPreprocessedProgram` to other classes. 

`StarknetCompiler` is a facade for `StarknetPreprocessedProgram`. It's useful to use facades, especially if the external library often changes.  We may create a new class —  `StarknetPreprocessedProgramFacade` in the future, if necessary.

src/commands/test/test_scheduler.py
```diff
@@ -0,0 +1,49 @@
...
+
+            try:
+                with multiprocessing.Pool(
+                    multiprocessing.cpu_count(),
+                    lambda: signal.signal(
+                        signal.SIGINT, signal.SIG_IGN
+                    ),  # prevents showing a stacktrace on cmd/ctrl + c
+                ) as pool:
+                    pool.map_async(self._worker, setups)
+                    self._live_logger.log(test_results_queue, test_collector_result)
```
MaksymilianDemitraszek, line 47: > If the process throws an unexpected error, Protostar should crash. It should be a rare case, therefore we shouldn't care about logging results.

You probably misunderstood me here, if you won't provide get protostar **won't** crash in case of unexpected error. 

Please add it.

tests/e2e/test_testing.py
```diff
@@ -17,36 +20,44 @@ def test_complex(protostar, copy_fixture):
...
+        "./tests",
+    )
 
-    result = protostar(["--no-color", "test", "tests/test_expect_revert.cairo"])
+    result = protostar(["--no-color", "test", "tests/expect_revert_test.cairo"])
 
-    assert "[FAIL] tests/test_expect_revert.cairo test_fail_error_message" in result
+    assert "[FAIL] tests/expect_revert_test.cairo test_fail_error_message" in result
 
     assert (
```
MaksymilianDemitraszek, line 43: can we add `[PASS]` asserts here?

MaksymilianDemitraszek (changes requested):
LGTM, close to being merge

src/commands/test/cheatcodes/expect_revert_cheatcode.py
```diff
@@ -0,0 +1,36 @@
...
+from typing import Callable, Optional
+
+from src.commands.test.cheatcodes._cheatcode import Cheatcode
+from src.commands.test.test_environment_exceptions import RevertableException
+
+
+class RevertableTestingExecutionEnvironment(ABC):
+    @abstractmethod
+    def expect_revert(self, expected_error: RevertableException) -> Callable[[], None]:
+        ...
```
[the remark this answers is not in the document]
kasperski95: That's a cyclic dependency. In one of the previous PRs you said that cyclic dependency is a sign of problems with the architecture.

src/commands/test/examples/basic/test_basic.cairo
```diff
@@ -12,27 +10,23 @@ namespace BasicContract:
...
-func test_call_not_deployed{syscall_ptr : felt*, range_check_ptr}():
+func test_cannot_call_methods_of_not_deployed_contract{syscall_ptr : felt*, range_check_ptr}():
     alloc_locals
     %{ expect_revert() %}
 
     local contract_a_address : felt
-    %{ 
-        ids.contract_a_address = 34134124
-    %}
+    %{ ids.contract_a_address = 34134124 %}
```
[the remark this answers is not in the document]
kasperski95: There's an expect revert below alloc_locals. I'll move it lower.

src/commands/test/test_scheduler.py
kasperski95: I'll leave this `get` just to avoid changing the code.

kasperski95:
@MaksymilianDemitraszek I couldn't answer below your comment for some reason.
`TestCommand::Args`:

> IDE - we shouldn't write code for specific IDE, everyone can use different one, or not use one at all. + it's easier to see suggested args for function.
This is not a code for a specific IDE.
 
> type-safety - i don't understand what you mean by that, you can provide types for function arguments. creating args in test_command_test.py
It's true but it makes sure args in tests and in the code are the same.
 
> creating args in test_command_test.py - you can use list if you mean fixtures
> function(*[1,2,3])
> It's inconsistent with other places, and as far as I'm ok with worker args, it's just weird to build something like this for normal function. It's artificial abstraction making code harder to read.

I don't know how to apply your example in this case. I created this class mainly because I didn't know how to pass create args in tests without defining a class. It looks like I have to use `setattr`.

MaksymilianDemitraszek (approved):

PR #252, opened 2022-05-09
Fix runtime imports

MaksymilianDemitraszek:


src/start.py
```diff
@@ -1,16 +1,16 @@
-# pylint: disable=import-outside-toplevel
+# Python complains about importing `Project`` if the import below is removed
```
kasperski95, line 1: Is this line necessary?

src/commands/build/build_project.py
```diff
@@ -83,7 +83,7 @@ def build_project(
                     + '\nDid you forget to update protostar.toml::["protostar.contracts"]?'
                 )
             ) from err
-        except (StarkException, VmException, PreprocessorError) as err:
+        except (StarkException, VmException, PreprocessorError) as err: #type: ignore
```
kasperski95, line 86: Is type ignoring necessary?

kasperski95 (approved):
LGTM

src/start.py
MaksymilianDemitraszek: Nah, i copied it by mistake

src/commands/build/build_project.py
MaksymilianDemitraszek: I don't know why type checker raises error here, it seem correct, do you have any idea?
```
error: Except clause is unreachable because exception is already handled (reportGeneralTypeIssues)
```

kasperski95: It's a bug in `pyright`. It's fixed in newer versions. Run `poetry install`.

PR #258, opened 2022-05-11
Refactor test collector

kasperski95:
This is a sub-PR of https://github.com/software-mansion/protostar/pull/250.


src/commands/test/utils.py
```diff
@@ -86,7 +86,7 @@ def match(self, state_event: Event) -> bool:
 @dataclass
 class TestSubject:
     """
-    A dataclass consisting of identification of a single test bundle, and target functions
+    Deprecated. Use `TestSuite`.
```
MaksymilianDemitraszek, line 89: Why do you deprecate that? The case is we use this one in runner, yup?🤔 
You can make type alias instead `TestSubject = TestSuite` then

src/commands/test/test_collector.py
```diff
@@ -1,93 +1,131 @@
...
+            return self._starknet_compiler.get_functions(file_path, prefix="test_")
         except PreprocessorError as p_err:
             print(p_err)
-            raise CollectionError("Failed to collect test cases") from p_err
-        return [
-            fn
-            for fn in preprocessed.abi
-            if fn["type"] == "function" and fn["name"].startswith("test_")
-        ]
+            raise TestCollectingException("Failed to collect test cases") from p_err
```
MaksymilianDemitraszek, line 131: I am little bit worried about this error, it probably should be handled by TestLiveReporter, and we should run other test suites, maybe not in this PR. Can you make task for this?

src/commands/test/test_collector.py
```diff
@@ -1,93 +1,131 @@
...
+        self, file_path: Path
+    ) -> List[StarknetCompiler.AbiElement]:
         try:
-            preprocessed = StarknetCompiler(
-                include_paths=self.include_paths or [],
-                disable_hint_validation=True,
-            ).preprocess_contract(file_path)
+            return self._starknet_compiler.get_functions(file_path, prefix="test_")
         except PreprocessorError as p_err:
             print(p_err)
```
MaksymilianDemitraszek, line 130: Why do we print it 🤔 here, we probably shouldn't as we raise it again

src/utils/starknet_compilation.py
```diff
@@ -72,3 +73,17 @@ def compile_contract(
         )
         assert isinstance(assembled, ContractDefinition)
         return assembled
+
+    class AbiElement(TypedDict):
+        """NOTE: This type doesn't represent a dictionary faithfully. The dictionary can have more attributes."""
+
+        name: str
+        type: Literal["function"]
```
MaksymilianDemitraszek, line 81: Why do we need this class?, can't we just use string and make aliast `FunctionName = str`?

MaksymilianDemitraszek: type is always function, so we basically introducing misleading abstraction which does kinda nothing

MaksymilianDemitraszek (changes requested):
LGTM some minor issues, please adress them before merge

src/commands/test/utils.py
kasperski95: I tried initially, but it `TestSuite::test_functions` has different type than `TestSubject::test_functions`.

src/commands/test/test_collector.py
kasperski95: `PreprocessorError` contains the line of code that caused this error. This information is hard to extract.

src/utils/starknet_compilation.py
kasperski95: I'll change the type's type to string then. The goal of using Literal was expanding this type when necessary.

`StarknetCompiler::get_functions` uses "type" attribute of this dictionary. It looks like only function name is used in the remaining part of the code.

I'll bring back the dict type in this PR. In the other PR I'll change `test_functions` to `test_functino_names`.

src/commands/test/test_collector.py
kasperski95: https://github.com/software-mansion/protostar/issues/259

src/commands/test/test_suite.py
```diff
@@ -2,12 +2,8 @@
...
-from src.utils.starknet_compilation import StarknetCompiler
-
 
 @dataclass
 class TestSuite:
-    TestFunction = StarknetCompiler.AbiElement
-
     test_path: Path
-    test_functions: List["TestSuite.TestFunction"]
+    test_functions: List[dict]
```
MaksymilianDemitraszek, line 9: I think string would be enough, I don't think we are going to expand this soon

MaksymilianDemitraszek (approved):

src/commands/test/test_suite.py
kasperski95: I'll do it in a different PR.

PR #267, opened 2022-05-13
Feature - add StartPrank(address, address)

MaksymilianDemitraszek:
- [x] Docs


kasperski95 (commented):
The logic seems good. I'll approve when docs are ready.

src/commands/test/cheatable_syscall_handler.py
```diff
@@ -43,25 +43,37 @@ def _get_block_timestamp(self):
             else super()._get_block_timestamp()
         )
 
-    # prank
-    custom_caller_address = None
+    def set_caller_address(self, addr: int, target: Optional[int] = None):
+        target = target if target else self.contract_address
+        if target in self.state.pranked_contracts:
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: `pranked_contracts_map` maybe?

src/commands/test/examples/cheats/test_prank.cairo
```diff
@@ -0,0 +1,78 @@
+%lang starknet
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: > What if somebody provided a wrong address?
Nothing, we can't check that (maybe we can but may be overcomplicated with a low value for the end user)

I will write test cases for both

src/commands/test/examples/cheats/test_prank.cairo
```diff
@@ -73,6 +73,39 @@ func test_syscall_counter_correct{syscall_ptr : felt*, range_check_ptr}():
     %}
     let (caller_addr) = get_caller_address()
     assert caller_addr = 345
+    # We check if syscall counter has been correctly incremented
     let (bn) = get_block_number()
     return ()
```
kasperski95, line 78: There's no assertion for that. Does this function throw an error if syscall_counter wasn't correctly incremented?

website/docs/tutorials/guides/testing.md
```diff
@@ -443,18 +443,68 @@ Deploys a contract given a path relative to a Protostar project root. The sectio
...
 ```python
-def start_prank(caller_address: int) -> None: ...
+def start_prank(caller_address: int, target_contract_addr: Optional[int] = None) -> Callable: ...
 ```
 
-Changes caller address until [`stop_prank`](#stop_prank) cheatcode is used.
+Changes caller address until the returned callable is called. If `target_contract_addr` specified, prank affects only the contract with the specified address. Otherwise prank affects the current contract.
 
-### `stop_prank`
+#### With target
```
kasperski95, line 451: `Target` is a bit too generic name.

website/docs/tutorials/guides/testing.md
```diff
@@ -443,18 +443,68 @@ Deploys a contract given a path relative to a Protostar project root. The sectio
 ### `start_prank`
 
 ```python
-def start_prank(caller_address: int) -> None: ...
+def start_prank(caller_address: int, target_contract_addr: Optional[int] = None) -> Callable: ...
 ```
 
-Changes caller address until [`stop_prank`](#stop_prank) cheatcode is used.
+Changes caller address until the returned callable is called. If `target_contract_addr` specified, prank affects only the contract with the specified address. Otherwise prank affects the current contract.
```
kasperski95, line 449: | ... Otherwise, `start_prank`

(I cannot add a suggestion for some reason.)

website/docs/tutorials/guides/testing.md
```diff
@@ -443,18 +443,68 @@ Deploys a contract given a path relative to a Protostar project root. The sectio
...
+end
 ```
+```cairo title="Remote assert passes"
+@contract_interface
+namespace Pranked:
+    func assert_pranked() -> ():
+    end
+end
+
+@external
```
kasperski95, line 473: Maybe we should stick to `@view` everywhere.

website/docs/tutorials/guides/testing.md
```diff
@@ -443,18 +443,68 @@ Deploys a contract given a path relative to a Protostar project root. The sectio
...
-Resets caller address. Always used with [`start_prank`](#start_prank).
+#### Locally
+```cairo title="Local assert passes"
+
+@external
+func test_remote_prank{syscall_ptr : felt*, range_check_ptr}():
+    alloc_locals
+    local contract_address : felt
+    %{ 
+        ids.contract_address = deploy_contract("./src/commands/test/examples/cheats/pranked.cairo").contract_address
```
kasperski95, line 496: Wrong snippet. Use `test_local_prank`.

website/docs/tutorials/guides/testing.md
```diff
@@ -443,18 +443,68 @@ Deploys a contract given a path relative to a Protostar project root. The sectio
...
+    # Does not raise error
+    Pranked.assert_pranked(contract_address=contract_address)
+    %{ stop_prank_callable() %}
+    return ()
+end
+``` 
 
-Resets caller address. Always used with [`start_prank`](#start_prank).
+#### Locally
+```cairo title="Local assert passes"
```
kasperski95, line 489: Local assert should be the first example. Unit tests are more common than integration tests.

kasperski95 (changes requested):
Requesting changes due to a wrong example used in docs for a local usage of `start_prank`.

website/docs/tutorials/guides/testing.md
kasperski95: Maybe in `In integration tests`.

website/docs/tutorials/guides/testing.md
MaksymilianDemitraszek: I think we probably should stick to `@external `, cairo docs:

> The @view decorator is identical to the @external decorator. The only difference is that the method is annotated as a method that only queries the state rather than modifying it. Note that in the current version this is not enforced by the compiler.

src/commands/test/examples/cheats/test_prank.cairo
MaksymilianDemitraszek: Yes, if you call a syscall and the previous syscall won't increment the syscall counter, it will throw an error. We had this bug before.

MaksymilianDemitraszek: Added more comment about it

website/docs/tutorials/guides/testing.md
MaksymilianDemitraszek: We probably should not deploy in this test case at call...

website/docs/tutorials/guides/testing.md
kasperski95: https://github.com/software-mansion/protostar/issues/272

kasperski95 (approved):

website/docs/tutorials/guides/testing.md
MaksymilianDemitraszek: Nothing, it's just docs

PR #274, opened 2022-05-16
Fix exit codes bug

MaksymilianDemitraszek:


src/commands/test/test_command.py
```diff
@@ -74,12 +74,14 @@ def arguments(self) -> List[Command.Argument]:
...
 
     async def run(self, args) -> TestingSummary:
-        return await self.test(
+        summary = await self.test(
             target=args.target,
             match=args.match,
             omit=args.omit,
             cairo_path=args.cairo_path,
         )
+        summary.assert_all_passed()
```
MaksymilianDemitraszek, line 83: I was wondering what is the best place to run such assert. I concluded here because it's part of the command logic not the testing logic

kasperski95:
@MaksymilianDemitraszek 
Can you merge development to this branch? Currently, it's hard to review this PR.

PR #285, opened 2022-05-17
[Deploy command — 03] Integrate deploying logic with deploy command

kasperski95:


protostar/commands/deploy/deploy_command.py
```diff
@@ -40,10 +48,47 @@ def arguments(self) -> List[Command.Argument]:
...
+    async def deploy(
+        self,
+        contract_name: str,
+        network: str,
+        inputs: Optional[List[str]] = None,
+        token: Optional[str] = None,
+        salt: Optional[str] = None,
+    ):
+        with open(
+            self._build_output_path / f"{contract_name}.json",
```
MaksymilianDemitraszek, line 82: This is an open question but should user manually prebuild a contract before deployment?

protostar/commands/deploy/network_config.py
```diff
@@ -0,0 +1,38 @@
...
+from protostar.utils.config.project import Project
+
+
+class InvalidNetworkConfigException(ProtostarException):
+    pass
+
+
+class NetworkConfig:
+    @classmethod
+    def from_config_file(cls, network_name: str, project: Project) -> "NetworkConfig":
```
MaksymilianDemitraszek, line 13: I don't like we access config in different ways in each module. Can we use `ArgumentValueFromConfigProvider` to access config data?

protostar/commands/deploy/network_config.py
```diff
@@ -0,0 +1,38 @@
...
+        gateway_url = project.load_argument(
+            f"protostar.network.{network_name}", "gateway_url"
+        )
+        if gateway_url:
+            return cls(gateway_url=gateway_url)
+
+        starkware_network_name = project.load_argument(
```
MaksymilianDemitraszek, line 20: We probably need flags for all of the config values either

MaksymilianDemitraszek (changes requested):
I think we should develop a consistent way to access config file to avoid chaos

protostar/commands/deploy/deploy_command.py
kasperski95: Yes. I added suggestion to the user to run `build` command in case such file doesn't exist. Custom commands feature would allow the user to combine `build` and `deploy` into one command.

protostar/commands/deploy/network_config.py
kasperski95: No. Network configuration isn't part of the CLI interface. Also, there might be multiple network configurations.

Maybe I should move network configuration and make it part of deploy command interface.

```toml
[protostar.deploy]
# ...

[protostar.deploy.testnet]
# ...
```
`protostar deploy --profile testnet`

protostar/commands/deploy/deploy_command.py
kasperski95: The suggestion was added in this PR — https://github.com/software-mansion/protostar/pull/288.

protostar/commands/deploy/network_config.py
MaksymilianDemitraszek: Yup i think it should be part of deplot command interface, it would be consistent with other commands

MaksymilianDemitraszek: If current interface to access the config is not flexible enough, we need to change it. I don't want a different interface for each command, because it's slightly different. If you need my help with that let me know.

kasperski95: Added configuration profiles in the separate PR: https://github.com/software-mansion/protostar/pull/291. (Foundry also supports configuration profiles, but in slightly different way).

protostar/commands/deploy/deploy_command.py
```diff
@@ -1,9 +1,30 @@
...
+from protostar.protostar_exception import ProtostarException
+from protostar.utils.config.project import Project
 
 
 class DeployCommand(Command):
+    def __init__(self, project: Project, build_output_path: Path) -> None:
+        self._project = project
+        self._build_output_path = build_output_path
+
+        self._gateway_url_arg = Command.Argument(
```
MaksymilianDemitraszek, line 16: We could make those class properties, as they are static

protostar/commands/deploy/deploy_command.py
```diff
@@ -31,19 +52,78 @@ def arguments(self) -> List[Command.Argument]:
...
+
+            await deploy(
+                gateway_url=network_config.gateway_url,
+                compiled_contract_file=compiled_contract_file,
+                constructor_args=inputs,
+                salt=salt,
+                token=token,
+            )
+
+    def _get_network_config(
```
MaksymilianDemitraszek, line 112: build seems like better prefix imo

protostar/commands/deploy/deploy_command.py
```diff
@@ -31,19 +52,78 @@ def arguments(self) -> List[Command.Argument]:
...
+        network: Optional[str] = None,
+    ) -> NetworkConfig:
+        network_config: Optional[NetworkConfig] = None
+
+        if network:
+            network_config = NetworkConfig.from_starknet_network_name(network)
+        if gateway_url:
+            network_config = NetworkConfig(gateway_url=gateway_url)
+
+        if network_config is None:
```
MaksymilianDemitraszek, line 124: Would be cleaner to check it at the beginning of the function

MaksymilianDemitraszek (approved):

protostar/commands/deploy/deploy_command.py
kasperski95: I think you confused `network` with `network_config`.

MaksymilianDemitraszek: true

PR #287, opened 2022-05-17
[Deploy command — 04] Make deploy command public

kasperski95:


protostar/commands/deploy/deploy_command.py
```diff
@@ -31,23 +34,30 @@ def name(self) -> str:
...
                 type="str",
+                is_required=True,
+                is_positional=True,
             ),
             Command.Argument(
                 name="inputs",
-                description="The inputs to the constructor",
+                short_name="i",
+                description="The inputs to the constructor.",
                 type="str",
```
MaksymilianDemitraszek, line 61: shouldn't this one be a list?

MaksymilianDemitraszek: We need an e2e test for this one

protostar/commands/shared_args.py
```diff
@@ -0,0 +1,9 @@
+from protostar.cli.command import Command
+
+output_shared_argument = Command.Argument(
```
MaksymilianDemitraszek, line 3: I think it's unnecessary, it's good kind of redundancy

protostar/utils/config/project_test.py
```diff
@@ -55,6 +55,14 @@ def test_loading_argument(version_manager: VersionManager):
     assert proj.load_argument("build", "disable-hint-validation") is True
 
 
+def test_loading_nested_argument(version_manager: VersionManager):
+    proj = Project(
+        version_manager,
+        project_root=Path(current_directory, "examples", "nested"),
+    )
+    assert proj.load_argument("deploy.testnet", "network") == "foo"
```
MaksymilianDemitraszek, line 63: Can we test with missing arg?

MaksymilianDemitraszek:
We need e2e tests here either

protostar/commands/shared_args.py
kasperski95: It's a bad redundancy because it makes sure these arguments can be configured in shared_command_configs.

kasperski95: At the other hand, `output` can be misleading for deploy command.

protostar/commands/build/build_project.py
```diff
@@ -45,12 +46,7 @@ def arguments(self) -> List[Command.Argument]:
                 description="Disable validation of hints when building the contracts.",
                 type="bool",
             ),
-            Command.Argument(
-                name="output",
-                description="An output directory that will be used to put the compiled contracts in.",
-                type="path",
-                default="build",
-            ),
+            output_shared_argument,
```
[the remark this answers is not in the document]
kasperski95: I'll change it because deploy command might produce an output in the future.

protostar/commands/shared_args.py
MaksymilianDemitraszek: So we should do it for all of the redundant arguments I guess? can you make task for that?

MaksymilianDemitraszek:
Also we probably need a tutorial for deploying

protostar/commands/shared_args.py
kasperski95: https://github.com/software-mansion/protostar/issues/294

protostar/commands/deploy/deploy_command.py
kasperski95: It is. `is_array=True`.

kasperski95: Added integration test instead.

kasperski95:
> We need e2e tests here either

I'l add it in the next PR because it modifies command output.




> Also we probably need a tutorial for deploying

This is not the scope of this PR.

protostar/commands/deploy/deploy_command.py
```diff
@@ -61,6 +60,12 @@ def arguments(self) -> List[Command.Argument]:
...
                 is_array=True,
             ),
+            Command.Argument(
+                name="build-output",
```
MaksymilianDemitraszek, line 64: It probably should be source-dir, I first thought we are building contract inside this command and was weirded out

MaksymilianDemitraszek (approved):

protostar/commands/deploy/deploy_command.py
kasperski95: Definity it shouldn't be named source-dir. It suggest pointing to src directory.

kasperski95: Maybe the user should simply pass a path(s) to the compiled contract.

MaksymilianDemitraszek: True, maybe `compiled-contract-dir`?

protostar/commands/deploy/deploy_command.py
kasperski95: E2E is pain in the ass when used with deploying. The integration test should be enough.

MaksymilianDemitraszek: Why it's problematic

kasperski95: I had to run devnet before running tests.

Now I'm confused, because the devnet fixture should work. It spawns a new process. I'll leave it to the end. The integration test should be good enough for now because it covers the fragile part, that is mapping args to the deploy method.

MaksymilianDemitraszek: It's weird. If you can't manage to make it work let me know, I will take a look on that. We need an e2e test

kasperski95: I added e2e test in "Deploy command — 05" PR but it deploys to testnet instead. Deploying to devnet works, if devnet is started from the CLI level. It looks like the protostar subprocess doesn't see the devnet subprocess.

IMO, we can have one e2e test that deploys to testnet.

PR #288, opened 2022-05-17
[Deploy command — 05] Pretty print deploy feedback

kasperski95:
- [x] show link to block explorer if deployed to testnet
- [x] show available networks in `--network` description


protostar/commands/deploy/deploy_command.py
```diff
@@ -19,13 +19,24 @@ class CompilationOutputNotFoundException(ProtostarException):
...
 
     network_arg = Command.Argument(
         name="network",
-        description="The name of the StarkNet network.",
+        short_name="n",
+        description=(
+            "\n".join(
+                [
+                    "The name of the StarkNet network.",
+                    "It is required unless `--gateway_url` is provided.",
```
MaksymilianDemitraszek, line 33: It's a mistake, we probably should have`gateway-url` argument. It was probably reason of your problems with deploying on devnet :p

tests/e2e/test_deploying.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+
+@pytest.mark.usefixtures("init")
+def test_deploying_basic_example(protostar):
+    protostar(["build"])
+
+    result = protostar(["deploy", "./build/main.json", "--network", "alpha-goerli"])
```
MaksymilianDemitraszek, line 8: Can we test in on a contract with constructor arguments? We sould also deploy it on devnet. Also please provide a test case for configuration: contract with constructor arguments using config file

MaksymilianDemitraszek (changes requested):
LGTM, but we need some improvements on testing here

MaksymilianDemitraszek (approved):

PR #300, opened 2022-05-24
Bring back updated `start_prank` docs

kasperski95:


website/docs/tutorials/guides/testing.md
```diff
@@ -440,27 +440,12 @@ Deploys a contract given a path relative to a Protostar project root. The sectio
...
-Resets caller address. Always used with [`start_prank`](#start_prank).
 
-<!-- PROTOSTAR 0.2.1 -->
-<!-- 
 ```python
 def start_prank(caller_address: int, target_contract_address: Optional[int] = None) -> Callable: ...
 ```
 
-Changes caller address until the returned callable is called. If `target_contract_address` specified, `start_prank` affects only the contract with the specified address. Otherwise `start_prank` affects the current contract.
+Changes the caller address until the returned callable is called. If `target_contract_address` is specified, `start_prank` affects only the contract with the specified address. Otherwise, `start_prank` affects only the current contract.
```
MaksymilianDemitraszek, line 448:
```suggestion
Changes a caller address returned by `get_caller_address()` syscall until the returned callable is called. If `target_contract_address` is specified, `start_prank` affects only the contract with the specified address. Otherwise, `start_prank` affects only the current contract.
```

MaksymilianDemitraszek: maybe we should be a bit more specific here, because we only override method output

MaksymilianDemitraszek (approved):

PR #307, opened 2022-05-25
[`setup_state`::01]  Detect and try executing hook

kasperski95:
- added skipping integration test for the feature
- made Protostar raise `NotImplementedError` if `setup_state` is detected
  - discovered bug https://github.com/software-mansion/protostar/issues/308

### DIFF
--- a/protostar/commands/test/test_collector.py
+++ b/protostar/commands/test/test_collector.py
@@ -110,10 +110,12 @@ class TestCollector:
                 for test_case_name in test_case_names
                 if test_case_name == target_test_case_name
             ]
+
         return TestSuite(
             test_path=file_path,
             test_case_names=test_case_names,
             preprocessed_contract=preprocessed,
+            setup_state_fn_name=self._find_setup_state_hook_name(preprocessed),
         )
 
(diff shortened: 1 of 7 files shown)


protostar/commands/test/test_collector.py
```diff
@@ -131,7 +133,17 @@ def _get_test_suite_paths(self, target: Path) -> Generator[Path, None, None]:
     def _collect_test_case_names(
         self, preprocessed: StarknetPreprocessedProgram
     ) -> List[str]:
-        return self._starknet_compiler.get_function_names(preprocessed, prefix="test_")
+        return self._starknet_compiler.get_function_names(
+            preprocessed, predicate=lambda fn_name: fn_name.startswith("test_")
```
MaksymilianDemitraszek, line 137: maybe filter would be better name than predicate?

MaksymilianDemitraszek (approved):

protostar/commands/test/test_collector.py
kasperski95: `filter` collides with the global `filter` function

PR #316, opened 2022-05-27
[setup_state::03] Docs

kasperski95:
- [x] update to match 02

### DIFF
--- a/website/docs/tutorials/guides/testing.md
+++ b/website/docs/tutorials/guides/testing.md
@@ -190,6 +190,7 @@ You can find all [assert signatures here](https://github.com/software-mansion/pr
 
 ## `setup_state`
 Often while writing tests you have some setup work that needs to happen before tests run. The hook `setup_tmp_setup` can simplify your test functions and will speed up your tests. Use `tmp_state` to pass the information between `setup_state` and test functions as demonstrated on the example below:
+
 ```cairo
 %lang starknet
 
@@ -209,6 +210,10 @@ func test_something():
     return ()
 end
 ```
+:::note
+Protostar executes `setup_state` only once per test file, and creates a copy of StarkNet state and `tmp_state` object for each test case.
+:::
+
 
 :::warning
 Protostar copies `tmp_state` dictionary defined in `setup_state` to a test function shallowly.
--- a/website/docs/tutorials/guides/testing.md
+++ b/website/docs/tutorials/guides/testing.md
@@ -188,6 +188,32 @@ If your IDE supports Cairo and doesn't know how to import `protostar`, add the f
 
 You can find all [assert signatures here](https://github.com/software-mansion/protostar/blob/master/cairo/protostar/asserts.cairo).
 
+## `setup_state`
+Often while writing tests you have some setup work that needs to happen before tests run. The hook `setup_tmp_setup` can simplify your test functions and will speed up your tests. Use `tmp_state` to pass the information between `setup_state` and test functions as demonstrated on the example below:
+```cairo
+%lang starknet
+
+@view
+func setup_state():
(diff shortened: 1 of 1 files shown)


website/docs/tutorials/guides/testing.md
```diff
@@ -188,6 +188,39 @@ If your IDE supports Cairo and doesn't know how to import `protostar`, add the f
...
 
+## `setup_state`
+Often while writing tests you have some setup work that needs to happen before tests run. The hook `setup_tmp_setup` can simplify your test functions and will speed up your tests. Use `tmp_state` to pass the information between `setup_state` and test functions as demonstrated on the example below:
+
+```cairo
+%lang starknet
+
+@view
+func setup_state():
+    %{ tmp_state.contract_a_address = deploy_contract("./tests/integration/testing_hooks/basic_contract.cairo").contract_address %}
```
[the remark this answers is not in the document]
kasperski95: `context`

website/docs/tutorials/guides/testing.md
```diff
@@ -188,6 +188,39 @@ If your IDE supports Cairo and doesn't know how to import `protostar`, add the f
 
 You can find all [assert signatures here](https://github.com/software-mansion/protostar/blob/master/cairo/protostar/asserts.cairo).
 
+## `__setup__`
+Often while writing tests you have some setup work that needs to happen before tests run. The hook `__setup__` can simplify and speed up your tests. Use `context` hint local to pass data from `__setup__` to test functions as demonstrated on the example below:
```
MaksymilianDemitraszek, line 192:
```suggestion
Often while writing tests you have some setup work that needs to happen before tests run. The hook `__setup__` can simplify and speed up your tests. Use `context` variable to pass data from `__setup__` to test functions as demonstrated on the example below:
```

website/docs/tutorials/guides/testing.md
```diff
@@ -438,7 +471,9 @@ class DeployedContract:
 ```
 Deploys a contract given a path relative to a Protostar project root. The section [Deploying contracts from tests](#deploying-contracts-from-tests) demonstrates a usage of this cheatcode.
 
-
+:::warning
+Deploying contract is a slow operation. If it's possible try using this cheatcode in the [`__setup__` hook](#__setup__).
```
MaksymilianDemitraszek, line 475:
```suggestion
Deploying a contract is a slow operation. If it's possible try using this cheatcode in the [`__setup__` hook](#__setup__).
```

website/docs/tutorials/guides/testing.md
```diff
@@ -188,6 +188,39 @@ If your IDE supports Cairo and doesn't know how to import `protostar`, add the f
...
+
+    return ()
+end
+```
+:::warning
+Hint local `context` can only store strings, integers and booleans.
+:::
+
+:::info
+Protostar executes `__setup__` only once per a [test suite](https://en.wikipedia.org/wiki/Test_suite). Then, for each test case Protostar copies the StarkNet state and `context` object.
```
MaksymilianDemitraszek, line 218: I think we don't need this one. We don't explain harder words ex. `StarkNet state`, and it's not our job. 
```suggestion
Protostar executes `__setup__` only once per a test suite. Then, for each test case Protostar copies the StarkNet state and `context` object.
```

MaksymilianDemitraszek (approved):

website/docs/tutorials/guides/testing.md
kasperski95: We don't require technical background from the target audience, and we use `test suite` term in the CLI output, so it's worth to at least add a link. It doesn't change anything for the guys familiar with this term.
`pros` > `cons`

PR #320, opened 2022-05-30
Display emitted events when cheatcode failed

MaksymilianDemitraszek:


MaksymilianDemitraszek:
Follow up: https://github.com/software-mansion/protostar/issues/321

protostar/commands/test/expected_event.py
```diff
@@ -40,38 +42,45 @@ def __init__(
     def __str__(self) -> str:
         result: List[str] = []
 
-        result.append(f'"name": "{self.name}"')
+        result.append(f'"name": ({self.name})"{get_selector_from_name(self.name)}"')
```
kasperski95, line 45: The string representation of `ExpectedEvent` should match it's interface. The user may try to use a "selector" instead of a "name" in the `expect_events` cheatcode to fix the error. I'd create a new virtual field — "selector" just to not confuse the user.

kasperski95 (changes requested):
Requesting changes due to:
- selectors displayed in the name field

kasperski95 (approved):

PR #322, opened 2022-05-30
[globs::01] Add `collect_from_globs` method to `TestCollector`

kasperski95:
Extended `TestCollector` to search and ignore, `test_suites` and `test_cases` by globs.
(This PR doesn't change the test command's interface.)

--- 

1. map `test_suite_glob::test_case_glob` to `List[test_suite_path::test_case_glob]` for `target` and `ignore`
2. combine `test_case_globs` in `Mapping[test_case_path, Set[test_case_glob]]` for `target` and `ignore`
3. subtract from target ignored paths
4. build `TestSuite` from target and ignored test case globs for the specific test suite
5. ...

### DIFF
--- a/protostar/commands/test/test_collector.py
+++ b/protostar/commands/test/test_collector.py
@@ -299,18 +299,23 @@ class TestCollector:
         preprocessed = self._preprocess_contract(test_suite_path)
         collected_test_case_names = set(self._collect_test_case_names(preprocessed))
 
-        target_test_case_names: Set[str] = collected_test_case_names.copy()
+        # gather test cases that match any test case glob
+        target_test_case_names: Set[str] = set()
         for test_case_name in collected_test_case_names:
-            for test_case_glob in test_cases_info.test_case_globs:
-                if not fnmatch(test_case_name, test_case_glob):
-                    target_test_case_names.remove(test_case_name)
-                    break
+            if len(test_cases_info.test_case_globs) == 0:
+                target_test_case_names = collected_test_case_names.copy()
+            else:
+                for test_case_glob in test_cases_info.test_case_globs:
+                    if fnmatch(test_case_name, test_case_glob):
+                        target_test_case_names.add(test_case_name)
 
+        # filter out test cases that match any test case glob
         not_ignored_target_test_case_names = target_test_case_names.copy()
         for test_case_name in target_test_case_names:
             for ignored_test_case_glob in test_cases_info.ignored_test_cases:
                 if fnmatch(test_case_name, ignored_test_case_glob):
                     not_ignored_target_test_case_names.remove(test_case_name)
+                    break
 
         return TestSuite(
             test_path=test_suite_path,
--- a/protostar/commands/test/test_collector.py
+++ b/protostar/commands/test/test_collector.py
@@ -1,17 +1,16 @@
 import os
 import re
 from dataclasses import dataclass
+from fnmatch import fnmatch
 from glob import glob
 from logging import Logger
 from pathlib import Path
 from typing import Generator, List, Optional, Pattern, Set, cast
 
-from starkware.cairo.lang.compiler.preprocessor.preprocessor_error import (
-    PreprocessorError,
-)
-from starkware.starknet.compiler.starknet_preprocessor import (
-    StarknetPreprocessedProgram,
-)
+from starkware.cairo.lang.compiler.preprocessor.preprocessor_error import \
+    PreprocessorError
+from starkware.starknet.compiler.starknet_preprocessor import \
+    StarknetPreprocessedProgram
 
 from protostar.commands.test.test_suite import TestSuite
 from protostar.protostar_exception import ProtostarException
@@ -98,10 +97,10 @@ class TestCollector:
         unique_test_suites: Set[TestSuite] = set()
         for glob_target in glob_targets:
             test_suite_filepaths: Set[str] = set()
-            target_test_case: Optional[str] = None
+            target_test_case_glob: Optional[str] = None
             valid_glob = glob_target
             if re.match(r"^.*\.cairo::.*", glob_target):
-                valid_glob, target_test_case = glob_target.split("::")
+                valid_glob, target_test_case_glob = glob_target.split("::")
 
             test_suite_filepaths.update(
                 self._find_test_suite_filepaths_from_glob(valid_glob)
@@ -120,7 +119,7 @@ class TestCollector:
             test_suites: List[TestSuite] = []
             for test_suite in test_suite_filepaths:
                 test_suites.append(
-                    self._build_test_suite(Path(test_suite), target_test_case)
+                    self._build_test_suite_from_test_case_glob(Path(test_suite), target_test_case_glob)
                 )
 
             non_empty_test_suites = list(
@@ -152,6 +151,25 @@ class TestCollector:
             if TestCollector.is_test_suite(Path(filepath).name):
                 results.add(filepath)
         return results
+
+    def _build_test_suite_from_test_case_glob(
+        self, file_path: Path, target_test_case_glob: Optional[str]
+    ) -> TestSuite:
+        preprocessed = self._preprocess_contract(file_path)
+        test_case_names = self._collect_test_case_names(preprocessed)
+        if target_test_case_glob:
+            test_case_names = [
+                test_case_name
+                for test_case_name in test_case_names
+                if fnmatch(test_case_name, target_test_case_glob)
+            ]
+
+        return TestSuite(
+            test_path=file_path,
+            test_case_names=test_case_names,
+            preprocessed_contract=preprocessed,
+            setup_state_fn_name=self._find_setup_state_hook_name(preprocessed),
+        )
 
     @classmethod
     def is_test_suite(cls, filename: str) -> bool:
--- a/protostar/commands/test/test_collector.py
+++ b/protostar/commands/test/test_collector.py
@@ -233,8 +233,6 @@ class TestCollector:
         results: TestCasesDict = {}
 
         for parsed_target in parsed_targets:
-            if not parsed_target.test_suite_glob:
-                continue
             test_suite_paths = self._find_test_suite_paths_from_glob(
                 parsed_target.test_suite_glob
             )
--- a/protostar/commands/test/test_collector.py
+++ b/protostar/commands/test/test_collector.py
@@ -48,12 +48,12 @@ class ParsedTarget:
 
 
 @dataclass
-class TargetTestCasesInfo:
+class TestSuiteInfo:
     test_case_globs: Set[TestCaseGlob]
     ignored_test_case_globs: Set[TestCaseGlob]
 
 
-TargetTestCasesInfoDict = Dict[TestSuitePath, TargetTestCasesInfo]
+TestSuiteInfoDict = Dict[TestSuitePath, TestSuiteInfo]
 
 
 class TestCollectingException(ProtostarException):
@@ -209,13 +209,13 @@ class TestCollector:
             ignored_test_cases_dict,
         )
 
-        target_test_cases_info_dict = self.build_target_test_cases_info_dict(
+        test_suite_info_dict = self.build_test_suite_info_dict(
             filtered_test_cases_dict,
             ignored_test_cases_dict,
         )
 
-        test_suites = self._build_test_suites_from_target_test_cases_info_dict(
-            target_test_cases_info_dict
+        test_suites = self._build_test_suites_from_test_suite_info_dict(
+            test_suite_info_dict
         )
 
         non_empty_test_suites = list(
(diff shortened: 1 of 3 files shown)


kasperski95:
> LGTM, but please reafactor the logic as we discussed. Deduplicating paths, then running collect on each then filtering test case names will be much cleaner flow

It won't be much cleaner because filtering by test case name must be done only for certain test suites. 
a::test_foo*
b::test_bar*
is no the same as
`{a,b} x {test_foo*,test_bar*}`

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +170,156 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+            if test_suite_filepath in ignored_test_cases_mapping:
+                target_test_cases_info.ignored_test_cases = ignored_test_cases_mapping[
+                    test_suite_filepath
+                ]
+
+        return result
+
+    def _find_test_suite_filepaths_from_glob(self, test_suite_glob: str) -> Set[str]:
+        results: Set[str] = set()
+        matches = glob(test_suite_glob, recursive=True)
```
MaksymilianDemitraszek, line 259:
```suggestion
        matches = map(Path, glob(test_suite_glob, recursive=True))
```

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +170,156 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+        for match in matches:
+            path = Path(match)
+            if path.is_dir():
+                results.update(self._find_test_suite_filepaths_in_dir(path))
+            elif path.is_file() and TestCollector.is_test_suite(path.name):
+                results.add(match)
+        return results
+
+    # pylint: disable=no-self-use
+    def _find_test_suite_filepaths_in_dir(self, path: Path) -> Set[str]:
```
MaksymilianDemitraszek, line 269:
```suggestion
    def _find_test_suite_filepaths_in_dir(self, path: Path) -> Set[TestSuiteFilepaths]:
```

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +170,156 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+                test_suite_filepath
+            ]
+            if test_suite_filepath in ignored_test_cases_mapping:
+                target_test_cases_info.ignored_test_cases = ignored_test_cases_mapping[
+                    test_suite_filepath
+                ]
+
+        return result
+
+    def _find_test_suite_filepaths_from_glob(self, test_suite_glob: str) -> Set[str]:
```
MaksymilianDemitraszek, line 257:
```suggestion
    def _find_test_suite_filepaths_from_glob(self, test_suite_glob: str) -> Set[TestSuiteFilepath]:
```

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +170,156 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+            path = Path(match)
+            if path.is_dir():
+                results.update(self._find_test_suite_filepaths_in_dir(path))
+            elif path.is_file() and TestCollector.is_test_suite(path.name):
+                results.add(match)
+        return results
+
+    # pylint: disable=no-self-use
+    def _find_test_suite_filepaths_in_dir(self, path: Path) -> Set[str]:
+        filepaths = set(glob(f"{path}/**/*.cairo", recursive=True))
```
MaksymilianDemitraszek, line 270:
```suggestion
        filepaths = set(glob(str(path /"**" /"*.cairo"), recursive=True))
```

protostar/commands/test/test_collector.py
```diff
@@ -16,17 +18,35 @@
...
+TestSuiteGlob = str
+TestSuiteFilepath = str
+TestCaseGlob = str
+TestSuiteAndCaseGlob = str
+"""e.g. `tests/**/::test_*`"""
+TestCasesMapping = Dict[TestSuiteFilepath, Set[TestCaseGlob]]
+
+
+@dataclass
+class TargetTestCasesInfo:
```
MaksymilianDemitraszek, line 30: This is more a Suite Info, also target is a weird word. Maybe?
```suggestion
class CollectedTestSuite:
```

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +170,156 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+
+    # pylint: disable=no-self-use
+    def build_target_test_cases_info_mapping(
+        self,
+        test_cases_mapping: TestCasesMapping,
+        ignored_test_cases_mapping: TestCasesMapping,
+    ) -> TargetTestCasesInfoMapping:
+        # filter out test suites, which shouldn't be parsed
+        filtered_test_cases_mapping = test_cases_mapping
+        for ignored_target_path in ignored_test_cases_mapping:
```
MaksymilianDemitraszek, line 233: Please move this filtering into separate function

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +170,156 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+            if (
+                len(ignored_test_cases_mapping[ignored_target_path]) == 0
+                and ignored_target_path in filtered_test_cases_mapping
+            ):
+                del filtered_test_cases_mapping[ignored_target_path]
+
+        # create the mapping
```
MaksymilianDemitraszek, line 240:
```suggestion
```

protostar/commands/test/test_collector.py
```diff
@@ -16,17 +18,35 @@
...
+TestCaseGlob = str
+TestSuiteAndCaseGlob = str
+"""e.g. `tests/**/::test_*`"""
+TestCasesMapping = Dict[TestSuiteFilepath, Set[TestCaseGlob]]
+
+
+@dataclass
+class TargetTestCasesInfo:
+    test_case_globs: Set[TestCaseGlob]
+    ignored_test_cases: Set[TestCaseGlob]
```
MaksymilianDemitraszek, line 32:
```suggestion
    test_case_globs: Set[TestCaseGlob]
    ignored_test_case_glob: Set[TestCaseGlob]
```

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +170,156 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+        test_cases_info: TargetTestCasesInfo,
+    ) -> TestSuite:
+        preprocessed = self._preprocess_contract(test_suite_path)
+        collected_test_case_names = set(self._collect_test_case_names(preprocessed))
+
+        # gather test cases that match any test case glob
+        target_test_case_names: Set[str] = set()
+        for test_case_name in collected_test_case_names:
+            if len(test_cases_info.test_case_globs) == 0:
+                target_test_case_names = collected_test_case_names.copy()
```
MaksymilianDemitraszek, line 306: Why we have to copy?

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +170,156 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+            else:
+                for test_case_glob in test_cases_info.test_case_globs:
+                    if fnmatch(test_case_name, test_case_glob):
+                        target_test_case_names.add(test_case_name)
+
+        # filter out test cases that match any test case glob
+        not_ignored_target_test_case_names = target_test_case_names.copy()
```
MaksymilianDemitraszek, line 313: Why we have to copy?

MaksymilianDemitraszek (changes requested):
Logic seems ok, but we need to simplify some of the logic.
I would rename `mapping` -> `dict`

protostar/commands/test/test_collector.py
```diff
@@ -16,17 +18,35 @@
 from protostar.protostar_exception import ProtostarException
 from protostar.utils.starknet_compilation import StarknetCompiler
 
+TestSuiteGlob = str
+TestSuiteFilepath = str
+TestCaseGlob = str
+TestSuiteAndCaseGlob = str
```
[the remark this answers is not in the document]
kasperski95: `Pattern` is even more confusing
https://docs.python.org/3/library/typing.html#typing.Pattern

I'll change it to something else.

protostar/commands/test/test_collector.py
kasperski95: `len(target_test_case_names)` changed during loop execution

protostar/commands/test/test_collector.py
kasperski95: This copy is probably unnecessary.

protostar/commands/test/test_collector.py
```diff
@@ -16,17 +18,35 @@
 from protostar.protostar_exception import ProtostarException
 from protostar.utils.starknet_compilation import StarknetCompiler
 
+TestSuiteGlob = str
+TestSuiteFilepath = str
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: > Glob returns strings rather than Paths.
We should convert them to Path

protostar/commands/test/test_collector.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: Then provide some type abstraction over empty set, knowing that empty set equals any is very contextual

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +187,171 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
         except PreprocessorError as p_err:
             print(p_err)
             raise TestCollectingException("Failed to collect test cases") from p_err
+
+    def collect_from_globs(
+        self,
+        globs: List[str],
```
MaksymilianDemitraszek, line 193: targets, ignored targets

protostar/commands/test/test_collector.py
MaksymilianDemitraszek: Maybe `TargetTestCasesGlobs`?

kasperski95: An instance of `TargetTestCasesGlobs` should use snake case, and the snake version of this class would suggest a list or set of globs.

`CollectedTestSuite` could suggest that this class inherits from `TestSuite`.

I'll change it to `TestSuiteInfo`.

protostar/commands/test/test_collector.py
```diff
@@ -47,8 +67,13 @@ def log(self, logger: Logger):
             else:
                 logger.warning("No cases found")
 
-    def __init__(self, starknet_compiler: StarknetCompiler) -> None:
+    def __init__(
+        self,
+        starknet_compiler: StarknetCompiler,
+        default_test_suite_glob: Optional[str] = None,
```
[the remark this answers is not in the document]
kasperski95: I can move this parameter it to the `collect` function. This must be done in this class because the None value would propagate to the TestSuite::path.

protostar/commands/test/test_suite.py
```diff
@@ -13,3 +13,10 @@ class TestSuite:
     preprocessed_contract: StarknetPreprocessedProgram
     test_case_names: List[str]
     setup_fn_name: Optional[str] = None
+
+    def __hash__(self) -> int:
```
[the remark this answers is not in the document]
kasperski95: Leftover

protostar/commands/test/test_collector.py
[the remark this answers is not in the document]
kasperski95: I changed it to add `*` if no test case is provided to support the following use case `./test_foo.cairo ./test_foo.cairo::test_case_a`

protostar/commands/test/test_collector.py
kasperski95: I do the mapping in the loop below to avoid running the loop twice.

protostar/commands/test/test_collector.py
kasperski95: I'll leave it as it is. Too many brackets.

protostar/commands/test/test_collector.py
```diff
@@ -16,17 +20,56 @@
 from protostar.protostar_exception import ProtostarException
 from protostar.utils.starknet_compilation import StarknetCompiler
 
+TestSuiteGlob = str
+TestSuitePath = Path
+TestCaseGlob = str
+Target = str
+"""e.g. `tests/**/::test_*`"""
+TestCaseGlobsDict = Dict[TestSuitePath, Set[TestCaseGlob]]
```
MaksymilianDemitraszek, line 28: We probably can skip this abstraction at all, as we dissused on Slack

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +191,172 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+
+        # gather test cases that match any test case glob
+        target_test_case_names: Set[str] = set()
+        for test_case_name in collected_test_case_names:
+            if len(test_suite_info.test_case_globs) == 0:
+                target_test_case_names = collected_test_case_names
+            else:
+                for test_case_glob in test_suite_info.test_case_globs:
+                    if fnmatch(test_case_name, test_case_glob):
+                        target_test_case_names.add(test_case_name)
```
MaksymilianDemitraszek, line 345: Can we make this a TestSuiteInfo method?

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +191,172 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
...
+
+        # filter out test cases that match any test case glob
+        not_ignored_target_test_case_names = (
+            target_test_case_names.copy()
+        )  # copy prevents changing lengths of this collection during loop execution
+        for test_case_name in target_test_case_names:
+            for ignored_test_case_glob in test_suite_info.ignored_test_case_globs:
+                if fnmatch(test_case_name, ignored_test_case_glob):
+                    not_ignored_target_test_case_names.remove(test_case_name)
+                    break
```
MaksymilianDemitraszek, line 355: Can we make this a TestSuiteInfo method?

MaksymilianDemitraszek (changes requested):
Good job, close to being merged. I still have some minor ideas for improvement. 

sorry for prolonged process and being a bit too nitpicky 😅 , but it's 300 lines of complex alghoritm, It is worth to give it some extra attention.

protostar/commands/test/test_collector.py
```diff
@@ -151,3 +191,172 @@ def _preprocess_contract(self, file_path: Path) -> StarknetPreprocessedProgram:
         except PreprocessorError as p_err:
             print(p_err)
```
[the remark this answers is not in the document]
kasperski95: You did. I think this print is to print error above Protostar Exception.

```...
milion
lines
of
error
...
[ERROR] Summary
```

protostar/commands/test/test_collector.py
kasperski95: This is a short name for this dict. It affects the name of functions, parameters, variables and makes the name consistent. Renaming this type to more explicit form would make the code too explicit for the scope of this type.

MaksymilianDemitraszek (approved):

PR #323, opened 2022-05-31
Patch pedersen hash with cpp native implementation

THenry14:
This patch decreases tests execution times for complex contracts by ~30%.
We should be able to get rid of this when the change is eventually upstreamed to cairo-lang.


kasperski95:
It looks like this PR freezes E2E tests. I think we should patch it only for the test command.
@MaksymilianDemitraszek WDYT?

THenry14:
@kasperski95 @MaksymilianDemitraszek Thanks for the comments, I have moved the patch to tests only as requested.

It turned out e2e tests were not passing, because there was a problem bundling the crypto extensions binary. I have explicitly added it to protostar.spec file.

protostar/commands/test/__init__.py
```diff
@@ -1 +1,23 @@
+import sys
```
kasperski95, line 1: I think this patch will affect the `build` command because Protostar imports all commands in `protostar_cli.py`. I'd move this code to `TestExecutionEnvironment::invoke_test_case`.

THenry14: But patching this in `TestExecutionEnvironment::invoke_test_case` won't be effective, as original hashing function will already be imported and won't be patched for most invokations in cairo-lang itself (and there are over 11000 calls)

THenry14: Moreover, I have just tried to check how much calls test execution command makes to hashing, and it is ~20 calls, so patching only this won't yield any gains :( 

In comparison, the way I'm doing it right now results in ~30% faster tests (and I have just measured - there are exactly 16900 calls to the function), so I think it actually should affect the build command in order to make the tests run faster..

kasperski95: NVM. We would have to use lazy imports in each command.

THenry14: As discussed offline, such approach would still result in most of the calls being made to "original" hashing function, as other modules/classes import it indirectly, so a more "top-level" patch is desired.

@MaksymilianDemitraszek what do you think?

MaksymilianDemitraszek: I understand @kasperski95 concern, there is some risk connected to messing with cairo-lang code, but crypto-cpp is an offical starknet package. I think we can patch it everyhere. Eventually cairo-lang will support this AFAIK.

@THenry14 In this case can you please move this patch to global init as it affects everything? Sorry causing an extra effort :p

MaksymilianDemitraszek (approved):

protostar/commands/test/__init__.py
THenry14: No problem :+1: done

PR #328, opened 2022-06-01
SSL fix

MaksymilianDemitraszek:


binary_entrypoint.py
```diff
@@ -1,10 +1,14 @@
-import sys
+import sys, os, certifi
 from pathlib import Path
 
 if __name__ == "__main__":
     try:
         from protostar import main
 
+        # Use certifi certs to avoid problems on mac os
+        os.environ["SSL_CERT_FILE"] = certifi.where()
```
kasperski95, line 9: I tried something similar, but I used a different env variable — `REQUESTS_CA_BUNDLE`. Not sure if that worked because I couldn't replicate this issue.

binary_entrypoint.py
```diff
@@ -1,10 +1,14 @@
...
+        print(os.environ["SSL_CERT_FILE"])
```
kasperski95, line 10: Remove `print`.

kasperski95, line 10: Remove `print`.

binary_entrypoint.py
MaksymilianDemitraszek: `REQUESTS_CA_BUNDLE` is env var for requests library, requests bundles it's own certs

PR #330, opened 2022-06-02
Refactor/fix types in cheatable classes

kasperski95:
- use types correctly
- remove unnecessary asserts

PR #332, opened 2022-06-02
Map event selector to event name

kasperski95:
Based on the following PR: https://github.com/software-mansion/protostar/pull/330

- map selector to name
- refine output

Screen Shot 2022-06-02 at 16 46 19

### DIFF
--- a/protostar/commands/test/starkware/forkable_starknet.py
+++ b/protostar/commands/test/starkware/forkable_starknet.py
@@ -92,6 +92,7 @@ class ForkableStarknet(Starknet):
     def fork(self):
         return ForkableStarknet(state=self.cheatable_state.copy())
 
+    # pylint: disable=too-many-arguments
     async def deploy(
         self,
         source: Optional[str] = None,
--- a/protostar/commands/test/starkware/forkable_starknet.py
+++ b/protostar/commands/test/starkware/forkable_starknet.py
@@ -1,12 +1,13 @@
 import copy
-from typing import Dict, Optional, cast
+from typing import Dict, List, Optional, cast
 
 from starkware.cairo.lang.vm.crypto import pedersen_hash_func
 from starkware.starknet.business_logic.state.state import CarriedState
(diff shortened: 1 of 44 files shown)


kasperski95:
> Please merge the changes, I will be easier to review

@MaksymilianDemitraszek merged master

protostar/commands/test/starkware/forkable_starknet.py
```diff
@@ -15,6 +16,13 @@ class CheatableCarriedState(CarriedState):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.pranked_contracts_map: Dict[int, int] = {}
+        self.event_selector_to_name_map: Dict[int, str] = {}
+
+    def update_event_selector_to_name_map(
```
MaksymilianDemitraszek, line 21: Why do we copy this data? Can't we just provide property to ForkableStarknet which maps to this map? This way we would probably be compatible with new deploy syscall

MaksymilianDemitraszek (changes requested):
LGTM but some logic seems unnecessary.

Also I would like to see a test case:
- deploy test contract which deploys other contract and check if all selectors are present inside the map

protostar/commands/test/starkware/forkable_starknet.py
kasperski95: I can move this logic to `ForkableStarknet` if this is what you mean. `event_selector_to_name_map` stores mapping for all contracts whereas `local_event_selector_to_name_map` only for a single contract, that's why I need to copy the map.

Why do we need `self.pranked_contracts_map` and `CheatableCarriedState` then?

MaksymilianDemitraszek: Ok, we will have to investigate it with cairo 0.9 but for now write a test and we can merge this

MaksymilianDemitraszek (approved):

PR #338, opened 2022-06-06
[project_class_refactor::01] Rewrite from scratch

kasperski95:
- 02 -> add tests
- 03 -> integrate


protostar/protostar_toml/protostar_toml_section.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+from typing import Dict, List, Union
+
+
+class ProtostarTOMLSection(ABC):
+    Dict = Dict[str, Union[str, int, bool, List[str]]]
```
MaksymilianDemitraszek, line 6: This is cryptic + i suggest using different name than in standard typing to avoid being misleading,

MaksymilianDemitraszek (approved):

protostar/protostar_toml/protostar_toml_section.py
```diff
@@ -0,0 +1,15 @@
+from abc import ABC, abstractmethod
+from typing import Dict, List, Union
+
+
+class ProtostarTOMLSection(ABC):
+    TOMLCompatibleDict = Dict[str, Union[str, int, bool, List[str]]]
```
MaksymilianDemitraszek, line 6: `Compatible` is a bit weird as it suggests that other dicts aren't compatible. Compatibilty relates more to interfaces not typing. `ParsedProtostarTOML` idk?

PR #339, opened 2022-06-07
Inform user about the constructor limitation

kasperski95:


protostar/commands/test/test_runner.py
```diff
@@ -122,11 +132,17 @@ async def _run_test_suite(
...
-                        exception=err,
+                        exception=ex,
                     )
                 )
+
+    @staticmethod
+    def is_constructor_args_exception(ex: StarkException) -> bool:
+        if not ex.message:
+            return False
+        return "constructor" in ex.message and "__calldata_actual_size" in ex.message
```
MaksymilianDemitraszek, line 148: Won't this overlay with some `deploy_contract` exception?

kasperski95: It won't because`deploy_contract` raises `StarknetRevertableException` and that exception keeps message in `details` property.  Added a test case for that.

MaksymilianDemitraszek: What if i try to deploy it using 0.9 deply syscall handler?

MaksymilianDemitraszek (changes requested):
LGTM but please write a test case covering 0.9 deploy syscall

protostar/commands/test/test_runner.py
kasperski95: Protostar throws `RevertableException` for deploy syscall error.

MaksymilianDemitraszek (approved):

PR #349, opened 2022-06-09
Improve feedback after `upgrade` fail

kasperski95:


MaksymilianDemitraszek:
You forgot to request review, please remember about it, it makes my life easier 🥺

protostar/commands/upgrade/upgrade_command.py
```diff
@@ -93,12 +94,20 @@ def upgrade(self):
             self._pull_tarball()
             self._install_new_version()
             self.cleanup()
-        # pylint: disable=broad-except
-        except (Exception, KeyboardInterrupt, SystemExit) as err:
+        except KeyboardInterrupt:
+            self._rollback()
+            self.cleanup()
```
MaksymilianDemitraszek, line 99: Why not re-raise?

MaksymilianDemitraszek (changes requested):

protostar/commands/upgrade/upgrade_command.py
kasperski95: There's no point of printing details of this exception. It's an interrupt triggered by a user, not an error.

MaksymilianDemitraszek: Maybe then it's worth to add a log about interrupted process
`Interrupting...`
cleanup
`Upgrade interupted`

PR #353, opened 2022-06-09
Use data transformer

kasperski95:


protostar/commands/test/test_execution_environment.py
```diff
@@ -45,33 +49,40 @@ def contract_address(self):
...
 
 class TestExecutionEnvironment:
+    # pylint: disable=too-many-arguments
     def __init__(
         self,
         include_paths: List[str],
         forkable_starknet: ForkableStarknet,
         test_contract: StarknetContract,
         test_context: TestContext,
+        starknet_compiler: StarknetCompiler,
```
MaksymilianDemitraszek, line 59: We should rename it to `CairoCompilerFacade` at some point, current name doesn't make much sense (don't do it here)

MaksymilianDemitraszek (approved):
LGTM

PR #358, opened 2022-06-10
Use data transformer in `mock_call`

kasperski95:
- patched data transformer
- implemented in `mock call`
- removed old `deploy_contract` code 
- moved mock call to a separate file

- [x] remove `deploy_contract`


protostar/utils/data_transformer_facade.py
```diff
@@ -15,7 +16,52 @@ class FunctionNotFoundException(BaseException):
     pass
 
 
+class PatchedDataTransformer(DataTransformer):
```
MaksymilianDemitraszek, line 19: Please create a task to make this a  starknet.py PR in this release

MaksymilianDemitraszek (changes requested):

protostar/utils/data_transformer_facade.py
kasperski95: I'd do it later, once I finish mapping for arguments in events.

protostar/commands/test/test_execution_environment.py
```diff
@@ -263,9 +266,35 @@ def stop_prank(target_contract_address: Optional[int] = None):
...
+                List[int],
+                Dict[
+                    DataTransformerFacade.ArgumentName,
+                    DataTransformerFacade.SupportedType,
+                ],
+            ],
+        ):
             selector = get_selector_from_name(fn_name)
             try:
+                if isinstance(ret_data, Mapping):
```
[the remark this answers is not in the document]
kasperski95: I'll fix it after https://github.com/software-mansion/protostar/pull/360.

protostar/commands/test/cheatcodes/mock_call_cheatcode.py
```diff
@@ -0,0 +1,112 @@
...
+            List[int],
+            Dict[
+                DataTransformerFacade.ArgumentName,
+                DataTransformerFacade.SupportedType,
+            ],
+        ],
+    ):
+        selector = get_selector_from_name(fn_name)
+        if isinstance(ret_data, Mapping):
+            contract_path = self.get_contract_path_from_contract_address(
```
MaksymilianDemitraszek, line 76: Please move this block into `def _transform_from_python`

protostar/commands/test/cheatcodes/mock_call_cheatcode.py
```diff
@@ -0,0 +1,112 @@
...
+                contract_address
+            )
+            if contract_path is None:
+                raise CheatcodeException(
+                    self.name,
+                    (
+                        "Couldn't map the `contract_address` to the `contract_path`.\n"
```
MaksymilianDemitraszek, line 83: What do you think about altering the error a bit? "Couldn't map the `contract_address` to the `contract_path`. Make sure `contract_address` is correct or use a raw cairo output as return data"

protostar/commands/test/starkware/cheatable_state.py
```diff
@@ -119,6 +120,9 @@ def __init__(self, *args, **kwargs):
             AddressType, Dict[SelectorType, List[int]]
         ] = defaultdict(dict)
         self.event_selector_to_name_map: Dict[int, str] = {}
+        self.class_hash_to_contract_path_map: Dict[int, Path] = {}
```
MaksymilianDemitraszek, line 123: Please use type aliases here

protostar/commands/test/starkware/cheatable_state.py
```diff
@@ -119,6 +120,9 @@ def __init__(self, *args, **kwargs):
...
+        self.contract_address_to_contract_path_map: Dict[int, Path] = {}
```
MaksymilianDemitraszek, line 124: Can we merge those two dicts into one as they have the same key. We won't have to rememeber about inserting into two dicts.

kasperski95: Not really. This map is mainly used by syscall `deploy`.

protostar/commands/test/cheatcodes/mock_call_cheatcode.py
```diff
@@ -0,0 +1,112 @@
...
+    ):
+        super().__init__(
+            execute_entry_point_cls=execute_entry_point_cls,
+            tx_execution_context=tx_execution_context,
+            state=state,
+            caller_address=caller_address,
+            contract_address=contract_address,
+            starknet_storage=starknet_storage,
+            general_config=general_config,
+            initial_syscall_ptr=initial_syscall_ptr,
```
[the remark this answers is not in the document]
kasperski95: I don't want to lose type information, especially when overwriting starkware classes.

protostar/commands/test/cheatcodes/mock_call_cheatcode.py
kasperski95: `raw cairo output` might be hard to interpret. I'll change it to sth similar.

kasperski95: Actually, there's no point of suggesting providing a list. Contract address is invalid anyway.

protostar/commands/test/cheatcodes/mock_call_cheatcode.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: It's weird, won't pyright detect it? 😞 If not I agree with you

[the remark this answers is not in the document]
kasperski95: I need to define the constructor types somehow.

I'm moving the data transformer dependency to the cheatcode abstract class anyway.

protostar/commands/test/cheatcodes/mock_call_cheatcode.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+        self.data_transformer_facade = DataTransformerFacade(
+            StarknetCompiler(
+                include_paths=general_config.cheatcodes_cairo_path,
+                disable_hint_validation=True,
```
MaksymilianDemitraszek, line 54: Should we disable hint validation here by default?

kasperski95: Nah, `StarknetCompiler` shouldn't be strict, at least in this place.

MaksymilianDemitraszek: I think it sould be consistent with compilation in `declare` cheatcode, #382 We can revisit it working on this one to ensure consistency. Right now leave it whatever state you want.

kasperski95: > I think everythink it's okay, but we have to adjust it to refactored cheatcodes.
> 
> Also what is current status for, there is no hurry, just asking because I feel a bit lost in all of the changes:
> 
> * `prepare` transformer
> * `deploy_contract` transformer
> * `expect_events`  transformer

I think `deploy_contract` stopped supporting data transformation. I am also lost with that. We shouldn't refactor cheatcodes in this release. Too many changes everywhere. The cheatcode refactor didn't include `deploy_contract`. I need to recreate it. Arrgsgsdgjdf;lghjkldfgjhoai;t

Prepare and expect_events doesn't support data transformation.

kasperski95: Updated mock call to use the refactored approach.
Added data transformer to `deploy_contract`, and prepare cheatcodes.

protostar/commands/test/starkware/cheatable_state.py
```diff
@@ -115,10 +119,48 @@ class CheatableCarriedState(CarriedState):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.pranked_contracts_map: Dict[int, int] = {}
-        self.mocked_calls_map: Dict[
-            AddressType, Dict[SelectorType, List[int]]
-        ] = defaultdict(dict)
+        self.mocked_calls_map: Dict[AddressType, Dict[SelectorType, List[int]]] = {}
```
MaksymilianDemitraszek, line 122: Why not using a defaultdict here? Could avoid extra check in mock call

protostar/commands/test/cheatcodes/cheatcode.py
```diff
@@ -11,30 +11,38 @@
...
 if TYPE_CHECKING:
     from protostar.commands.test.starkware.cheatable_execute_entry_point import (
         CheatableExecuteEntryPoint,
     )
+    from protostar.commands.test.starkware.cheatable_state import CheatableCarriedState
 
 
 class Cheatcode(BusinessLogicSysCallHandler):
+    """
+    #### ACHTUNG:
```
MaksymilianDemitraszek, line 28: Yea, we shouldn't use `__subclassess__`, problem is it has to be loaded before so python knows it exists. Can we just list Cheatcodes explicilty in `execute_entry_point`?

> The connection between `CheatcodeFactory::build` and `ConcreteCheatcode::constructor` is not type-safe.

I wouldn't write such stuff in docstring. Python generally is not typesafe and nobody should rely on any type guarantees. It's unnecessary information which may be confusing

protostar/commands/test/test_execution_environment.py
```diff
@@ -121,6 +121,7 @@ def deploy_in_env(
                 )
```
MaksymilianDemitraszek, line 121: dead code

protostar/commands/test/test_execution_environment.py
```diff
@@ -74,6 +73,7 @@ def __init__(
         self._include_paths = include_paths
         self._test_finish_hooks: Set[Callable[[], None]] = set()
         self._starknet_compiler = starknet_compiler
```
MaksymilianDemitraszek, line 75: Aren't `_starknet_compiler` and `_data_transformer` unused?

MaksymilianDemitraszek (approved):
Don't want to block it so approving but please add missing prepare test and remove dead code before merging. Rest of suggestions is optional

protostar/commands/test/cheatcodes/cheatcode.py
kasperski95: | nobody should rely on any type guarantees

Unless the project uses types quite heavily. If you can't trust types, then the function shouldn't use it or use `Any` type.

Anyway, this will be removed.

protostar/commands/test/test_execution_environment.py
kasperski95: Starknet compiler is used. Data transformer is not.

protostar/commands/test/cheatcodes/cheatcode.py
MaksymilianDemitraszek: Unless the project uses types quite heavily. If you can't trust types, then the function shouldn't use it or use Any type.

Yep exactly, you should store such information using type hints, not docstrings

PR #359, opened 2022-06-10
[data transformer::03] expect events

kasperski95:


kasperski95:
I'll start again once ::02 is merged.

PR #361, opened 2022-06-13
[project_class_refactor::02] Add tests

kasperski95:


MaksymilianDemitraszek:
Can we merge 01 first?

protostar/protostar_toml/protostar_config_section.py
```diff
@@ -0,0 +1,40 @@
...
+@dataclass
+class ProtostarConfigSection(ProtostarTOMLSection):
+    protostar_version: VersionType
+
+    @staticmethod
+    def get_section_name() -> str:
+        return "config"
+
+    @classmethod
+    def from_protostar_toml_reader(
```
MaksymilianDemitraszek, line 21: i think from is a wrong word here. `load_with_protostar_toml_reader`?

kasperski95:
@MaksymilianDemitraszek done

protostar/protostar_toml/protostar_toml_writer.py
```diff
@@ -9,10 +10,23 @@
 )
 from protostar.protostar_toml.protostar_project_section import ProtostarProjectSection
 from protostar.protostar_toml.protostar_toml_section import ProtostarTOMLSection
+from protostar.utils.protostar_directory import VersionManager
 
 
 class ProtostarTOMLWriter:
-    # pylint: disable=no-self-use
+    def save_default(self, path: Path, version_manager: VersionManager):
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: Maybe each section should have classmethod `get_default`?

tests/integration/protostar_toml/protostar_toml_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+from pytest_mock import MockerFixture
+
+from protostar.protostar_toml.protostar_toml_reader import ProtostarTOMLReader
+from protostar.protostar_toml.protostar_toml_writer import ProtostarTOMLWriter
+from protostar.utils.protostar_directory import VersionManager
+
+
+def test_saving_and_reading(tmpdir, mocker: MockerFixture):
+    version_manager_mock = mocker.MagicMock()
```
MaksymilianDemitraszek, line 11: Wouldn't be better if this test loaded each section using `from_protoster_reader_toml` and check values in section object?

MaksymilianDemitraszek (changes requested):
Generally looks nice, good job. Some comments inlcuded.
Aside I think we should alter directory structure a bit, as we have a lot of objects.
Please ass `section` and `io` directories and move readers and section classes accordingly.

protostar/protostar_toml/_conftest.py
```diff
@@ -0,0 +1,15 @@
+from typing import Any
```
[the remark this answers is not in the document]
kasperski95: Test discovery didn't work in VS Code. `conftest.py` caused weird import issues. I needed a file to store a test utils rather than fixtures, so I renamed this file.

protostar/protostar_toml/protostar_config_section_test.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.fixture(name="protostar_config_section_dict")
+def protostar_config_section_dict_fixture() -> ProtostarConfigSection.TOMLCompatibleDict:
+    return {"protostar_version": "0.1.0"}
+
+
+def test_serialization(
+    mocker: MockerFixture,
+    protostar_config_section_dict,
+):
+    protostar_toml_mock = mock_protostar_toml_reader(mocker)(
```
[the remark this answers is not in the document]
kasperski95: I wanted to separate dependencies from configuration arguments.

tests/integration/protostar_toml/protostar_toml_test.py
kasperski95: It would be.

protostar/protostar_toml/_conftest.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: `conftest.py` caused weird import issues wdym by that?

kasperski95:
> Please add section and io directories and move readers and section classes accordingly.
@MaksymilianDemitraszek 
I did that initially, but I dropped this idea. It causes a cyclic dependency between `io` and `section` packages because `ProtostarTOMLWriter` expects specific sections.  And `ProtostarTOMLWriter` expects specific sections to prevent from creating invalid `protostar.toml`.

protostar/protostar_toml/protostar_config_section_test.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: Understood, please change it this code is confusing. You call it `()()` everywhere there is no point in this separation.

MaksymilianDemitraszek:
Weird. What logical structure of modules has to do with dependencies cycle?

protostar/protostar_toml/_conftest.py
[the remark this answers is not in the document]
kasperski95: In the VS Code output panel other there were "coulnd't import ..." in other `conftest.py` files.

kasperski95:
I try to avoid cyclic dependencies at the package/directory level. It doesn't have to cause cyclic dependency errors.

protostar/protostar_toml/_conftest.py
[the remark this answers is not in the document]
kasperski95:
```
__________________________________________________________________ ERROR collecting protostar/cli/argument_parser_facade_test.py __________________________________________________________________
ImportError while importing test module '/protostar/protostar/cli/argument_parser_facade_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../.pyenv/versions/3.7.12/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
protostar/cli/argument_parser_facade_test.py:7: in <module>
    from conftest import BaseTestCommand, FooCommand
E   ImportError: cannot import name 'BaseTestCommand' from 'conftest' (/protostar/protostar/protostar_toml/conftest.py)
_________________________________________________________________________ ERROR collecting protostar/cli/cli_app_test.py __________________________________________________________________________
ImportError while importing test module '/protostar/protostar/cli/cli_app_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../.pyenv/versions/3.7.12/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
protostar/cli/cli_app_test.py:6: in <module>
    from conftest import FooCommand
E   ImportError: cannot import name 'FooCommand' from 'conftest' (/protostar/protostar/protostar_toml/conftest.py)
```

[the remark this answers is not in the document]
kasperski95: I had to create `__init__.py`. Imports in Python are so unintuitive.

Screen Shot 2022-06-15 at 16 52 57

protostar/protostar_toml/io/protostar_toml_reader.py
```diff
@@ -46,15 +45,31 @@ def get_attribute(
...
+        return None
+
+    # pylint: disable=no-self-use
+    def _find_alternative_key(
+        self, base_key: str, raw_dict: Dict[str, Any]
+    ) -> Optional[str]:
+        if base_key in raw_dict:
+            return base_key
+
+        underscored_variant = base_key.replace("-", "_")
```
MaksymilianDemitraszek, line 61: Won't one of this be deprecated soon? I don't rememeber

MaksymilianDemitraszek: If it will please add a comment about it. Also maybe a warning?

protostar/protostar_toml/io/protostar_toml_reader.py
```diff
@@ -1,24 +1,23 @@
...
+
     def __init__(
         self,
-        version_manager: VersionManager,
         protostar_toml_path: Optional[Path] = None,
     ):
         self.path = protostar_toml_path or Path() / "protostar.toml"
-        self._version_manager = version_manager
-        self._cache: Optional[Dict[str, Any]] = None
+        self._cache: Optional[Dict[ProtostarTOMLReader.FlatSectionName, Any]] = None
```
MaksymilianDemitraszek, line 20: `_loaded_config` and `_load_if` would be better name. Cache is misleading, may suggest it's more complicated than it is, that it expires etc.

MaksymilianDemitraszek (approved):

protostar/protostar_toml/io/protostar_toml_reader.py
```diff
@@ -46,15 +45,31 @@ def get_attribute(
...
-            attribute_name = attribute_name.replace("-", "_")
-            if attribute_name not in section:
-                return None
+        alternative_attribute_name = self._find_alternative_key(attribute_name, section)
+
+        if alternative_attribute_name and alternative_attribute_name in section:
+            return section[alternative_attribute_name]
+        return None
+
+    # pylint: disable=no-self-use
```
MaksymilianDemitraszek, line 54: Why not make it a classmethod? You can still call classmethod using self

protostar/protostar_toml/io/protostar_toml_reader.py
```diff
@@ -46,15 +45,31 @@ def get_attribute(
...
+    def _find_alternative_key(
```
MaksymilianDemitraszek, line 55: Why  pylint ignore? staticmethod would be better

protostar/protostar_toml/io/protostar_toml_writer.py
```diff
@@ -12,7 +13,6 @@
 
 
 class ProtostarTOMLWriter:
-    # pylint: disable=no-self-use
     def save(
```
MaksymilianDemitraszek, line 16: I think we path should be a constructor argument, if you want to make it a class.

MaksymilianDemitraszek: Otherwise make this method static

protostar/protostar_toml/io/protostar_toml_reader.py
```diff
@@ -1,24 +1,23 @@
...
 import tomli
 
 from protostar.protostar_toml.protostar_toml_exceptions import (
     NoProtostarProjectFoundException,
 )
-from protostar.utils.protostar_directory import VersionManager
 
 
 class ProtostarTOMLReader:
+    FlatSectionName = str
```
MaksymilianDemitraszek, line 13: Please add comment what FlatSectionName is becuase it's cryptic

MaksymilianDemitraszek (approved):
Some minor nitpicks

protostar/protostar_toml/io/protostar_toml_reader.py
kasperski95: We can add a warning or support both syntaxes. IMO, it doesn't matter.

protostar/protostar_toml/io/protostar_toml_reader.py
kasperski95: `_loaded_config` can be confused with `ProtostarConfigSection`.

> Cache — hardware or software component that stores data so that future requests for that data can be served faster; the data stored in a cache might be the result of an earlier computation or a copy of data stored elsewhere

The name fits the definition of cache. I like this name because I can later use terms `cache hit` and `cache miss`.

> Cache is misleading, may suggest it's more complicated than it is, that it expires etc.

:/

protostar/protostar_toml/io/protostar_toml_reader.py
kasperski95: Static methods extend the interface of this class. I rather keep interface smaller than do a micro-optimization. 

Making it static or not doesn't matter.

protostar/protostar_toml/io/protostar_toml_reader.py
MaksymilianDemitraszek: Ok leave cache, it's still a private field so doesn't matter that much

protostar/protostar_toml/io/protostar_toml_reader.py
MaksymilianDemitraszek: using static is not matter of optimization idk. Is a guarantee that method does not alter object state, which is a VERY important information for developer.

protostar/protostar_toml/io/protostar_toml_writer.py
kasperski95: I created this class because:
- `ProtostarTOMLReader` suggests the existence of this class
- classes are easier to provide as a dependency, therefore it makes testing easier 

`Path` is a configuration, not a dependency argument. It should be provided to save method. 

I don't want to make a static method that isn't pure.
Static methods have to be patched in test files, and patching is more fragile than mocking.

MaksymilianDemitraszek: >I don't want to make a static method that isn't pure.

Why? staticmethod has nothing to do with being pure. It means method does not alter object's state

What do you mean by patching is more fragile than mocking 🤔 ?

kasperski95: Unpure statics have side effects. Side effects are bad because test cases aren't isolated. Therefore, these static methods have to be patched. Usually you don't have to patch pure static methods.


Patching is more fragile because you have to provide a string which can easily be invalidated during refactoring.

protostar/protostar_toml/io/protostar_toml_reader.py
kasperski95: This information comes with the cost of encouraging bad patterns.

I need to log the warning anyway, so this method can't be static anymore.

protostar/protostar_toml/io/protostar_toml_writer.py
MaksymilianDemitraszek: >Unpure statics have side effects. Side effects are bad because test cases aren't isolated. 

I don't see this sentence being an valid argument. Every function with side effects is unpure. Being static has nothing to do with purity, static only means that method does not alter object state. 

I see the issue with patching, but I don't think we should write stupid code only because of pytest quirks. Such invalidation will be easy to spot as test will simply fail(if not it's probably a bad test).

Please change it to staticmethod and use staticmethods whenever method doesn't use object's state.

protostar/protostar_toml/io/protostar_toml_reader.py
MaksymilianDemitraszek: > This information comes with the cost of encouraging bad patterns.
I don't see any bad patterns here, that is a purpose why staticmethod exists.

protostar/protostar_toml/io/protostar_toml_writer.py
kasperski95: I know that static has nothing to do with purity. The point I was trying to make
`Hard to mock stuff` + `Stuff that needs to be mocked` = hard to test code

> I see the issue with patching, but I don't think we should write stupid code only because of pytest quirks

~stupid~ -> testable
~ ~only~

I spend more time writing tests that actual code. Patching is unintuitive, fragile and sometimes doesn't work just because of naming collision.

> Please change it to staticmethod and use staticmethods whenever method doesn't use object's state.

Nope. I'd make it a function. I'll be avoiding statics at all costs.

protostar/protostar_toml/io/protostar_toml_reader.py
kasperski95: I'll do it in one of the next PRs, because this warning is the responsibility of `ArgumentValueFromConfigProvider`.

PR #364, opened 2022-06-14
[project_class_refactor::03] Replace old code in `init`. Refactor `init`.

kasperski95:
- connected `protostar_toml` to `init`
- refactored `init`

---
- [x] mark as ready to review once ::02 is merged


protostar/commands/init/project_creator/_project_creator.py
```diff
@@ -0,0 +1,41 @@
...
+    def save_protostar_toml(self, project_root: Path, libs_path: Path) -> None:
+        self.protostar_toml_writer.save(
+            path=project_root / "protostar.toml",
+            protostar_config=ProtostarConfigSection.get_default(self.version_manager),
+            protostar_project=ProtostarProjectSection(libs_path=libs_path),
+            protostar_contracts=ProtostarContractsSection.get_default(),
+        )
+
+    @abstractmethod
+    def run(self) -> None:
```
MaksymilianDemitraszek, line 40: Instead of using inheritance we could use strategy desing pattern here. But it ok either way just leaving it as an idea no need to change.

protostar/commands/init/init_command.py
```diff
@@ -1,21 +1,27 @@
...
+)
+from protostar.utils.requester import Requester
 
 
 class InitCommand(Command):
-    def __init__(self, script_root: Path, version_manager: VersionManager) -> None:
+    def __init__(
+        self,
+        requester: Requester,
+        new_project_creator: NewProjectCreator,
```
MaksymilianDemitraszek, line 18: I think we shouldn't inject creators here. I think Init command is the one which should be responsible for creating this objects. Can we change it?

MaksymilianDemitraszek (approved):
It's generaly very good, I like the way you implemented it. Some minor comments please addres them before merging

protostar/commands/init/init_command.py
kasperski95: I see the following issues with such approach:

- Creating objects in this class would require passing `ProjectCreator`'s dependencies to `InitCommand`.
- Stategy Pattern can be implemented with DI https://ufukhaciogullari.com/blog/strategy-pattern-with-dependency-injection/
- Objects should be created in the composition root https://blog.ploeh.dk/2011/07/28/CompositionRoot/
  - (I should've created `ProjectCreatorResolver`)
- DI framework will be easier to integrate if we decide to use it in the future
- Testability is better with current approach
- not following open–closed principle

protostar/utils/requester.py
```diff
@@ -0,0 +1,18 @@
+from protostar.utils.log_color_provider import LogColorProvider
+
+
+class Requester:
```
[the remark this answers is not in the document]
kasperski95: I forgot to push the change before merge. I'll do it in a different PR.

PR #366, opened 2022-06-14
Improve handling preprocessing errors in the collecting phase

kasperski95:
A broken test suite shouldn't prevent from running collected tests.


MaksymilianDemitraszek (approved):
Good job

PR #379, opened 2022-06-17
Refactor cheatcodes

MaksymilianDemitraszek:


protostar/commands/test/cheatcodes/cheatcode.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class Cheatcode(BusinessLogicSysCallHandler):
+    @property
+    def cheatable_state(self):
+        return cast(CheatableCarriedState, self.state)
+
+    @property
+    def cheatable_global_config(self):
+        return cast(CheatableStarknetGeneralConfig, self.state)
```
kasperski95, line 17:
```suggestion
        return cast(CheatableStarknetGeneralConfig, self.general_config)
```

protostar/commands/test/cheatcodes/cheatcode.py
```diff
@@ -0,0 +1,35 @@
...
+        ...
+
+
+class CheatcodeFactory():
+    def __init__(self, *args, **kwargs): 
+        self.args = args
+        self.kwargs = kwargs
+
+    def build(self, cheatcode_type: Type[Cheatcode]):
+        return cheatcode_type(*self.args, **self.kwargs)
```
kasperski95, line 35: :/

Deps should be provided in cheat code constructor.

protostar/commands/test/cheatcodes/cheatcode.py
```diff
@@ -0,0 +1,35 @@
...
+    def cheatable_global_config(self):
+        return cast(CheatableStarknetGeneralConfig, self.state)
+
+    @staticmethod
+    @abstractmethod
+    def name() -> str:
+        ...
+
+    @abstractmethod
+    def execute(self) -> Any:
```
kasperski95, line 25: Cheatcodes accept various arguments.

protostar/commands/test/cheatcodes/cheatcode.py
```diff
@@ -0,0 +1,35 @@
+from abc import abstractmethod
+from typing import Any, Type, cast
+
+from starkware.starknet.core.os.syscall_utils import BusinessLogicSysCallHandler
+
+from protostar.commands.test.starkware.cheatable_carried_state import CheatableCarriedState
+from protostar.commands.test.starkware.cheatable_starknet_general_config import CheatableStarknetGeneralConfig
+
+
+class Cheatcode(BusinessLogicSysCallHandler):
```
kasperski95, line 10: A cheatcode is not a SysCallHandler. We should pass a modified `BusinessLogicSysCallHandler` that supports middlewares as a dependency to this class.

protostar/commands/test/cheatcodes/cheatcode.py
MaksymilianDemitraszek: I decided to remove it. There is no point in defininig common execute interface as the interface is not common. unnecessary coupling.

protostar/commands/test/cheatcodes/cheatcode.py
```diff
@@ -0,0 +1,88 @@
...
+)
+
+if TYPE_CHECKING:
+    from protostar.commands.test.starkware.cheatable_execute_entry_point import (
+        CheatableExecuteEntryPoint,
+    )
+
+
+class Cheatcode(BusinessLogicSysCallHandler):
+    def __init__(
```
MaksymilianDemitraszek, line 28: In the future we can hoist this constructor to the parent class `CheatableSyscallHandler` and have `CheatedSyscallHandler` and `Cheatcode` inherit from it, but i don't want to pump this PR

kasperski95: I think Cheatcode should inherit from `CheatableSysCallHandler`.

protostar/commands/test/cheatcodes/declare_cheatcode.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class DeclareCheatcode(Cheatcode):
+    @staticmethod
+    def name() -> str:
+        return "declare"
+
+    @staticmethod
+    def implementation() -> str:
+        return "declare"
```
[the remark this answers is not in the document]
kasperski95: I don't like this approach because:
- `implementation` is not the best name
- it's not type safe
- makes refactoring harder
- potential errors will be catched by unit tests rather than pyright what increases feedback loop

I'd rather use `build` or `execute` approach. 
`build` method can return `Callable[[Any],Any]`

kasperski95 (changes requested):
We need to change `implementation` pattern to something else.

kasperski95:
I'm merging this PR. It's a blocker.

PR #381, opened 2022-06-21
Fix SSL issue for upgrading

MaksymilianDemitraszek:


protostar/commands/upgrade/upgrade_command.py
```diff
@@ -89,9 +88,10 @@ def upgrade(self):
             self.current_version,
             self.latest_version,
         )
+
+        self._pull_tarball()
```
kasperski95, line 92: Why did you move this call out of `try`'s scope?

kasperski95 (approved):
LGTM

protostar/commands/upgrade/upgrade_command.py
MaksymilianDemitraszek: Because i have to pull before backup, and there is nothing to cleanup if it fails.

MaksymilianDemitraszek: backup moves cert location

MaksymilianDemitraszek: good point 😅

PR #386, opened 2022-06-22
Use external for tests, added CI check

MaksymilianDemitraszek:
For the sake of consistency


kasperski95 (approved):
LGTM.
We should consider running (tests) checks in parallel.

PR #392, opened 2022-06-23
Execution resources per test case

kasperski95:
Screenshot 2022-06-24 at 14 28 23


protostar/commands/test/test_cases.py
```diff
@@ -20,13 +21,54 @@ def get_formatted_file_path(self):
...
-        result.append(f"[{log_color_provider.colorize('GREEN', 'PASS')}]")
-        result.append(f"{self.get_formatted_file_path()} {self.test_case_name}")
-        return " ".join(result)
+        first_line_elements: List[str] = []
+        first_line_elements.append(f"[{log_color_provider.colorize('GREEN', 'PASS')}]")
+        first_line_elements.append(
+            f"{self.get_formatted_file_path()} {self.test_case_name}"
+        )
+
+        if self.execution_resources:
```
mkaput, line 33: Where does `self.execution_resources` come from, and what is the difference between it and the `execution_resources` argument?

protostar/utils/log_color_provider.py
```diff
@@ -36,9 +36,9 @@ def get_color(
     def colorize(self, color_name: SupportedColorName, content: str):
         return f"{self.get_color(color_name)}{content}{self.get_color('RESET')}"
 
-    # pylint: disable=no-self-use
-    def bold(self, content: str):
-        return f"{Style.BRIGHT}{content}{Style.RESET_ALL}"
+    @staticmethod
+    def bold(content: Union[str, int]):
```
mkaput, line 40:
```suggestion
    def bold(content: Any):
```

so that it would match `print` or `str.format` signature

tests/integration/testing_output/testing_output_test.py
```diff
@@ -0,0 +1,28 @@
...
+            "test_printing_resource_usage_in_unit_testing_approach",
+            "test_printing_resource_usage_in_integration_testing_approach",
+        ],
+        expected_failed_test_cases_names=[],
+    )
+
+    for passed_test_case in testing_summary.passed:
+        output = str(passed_test_case)
+        assert "steps=" in output
+        assert "pedersen_builtin=" in output
```
mkaput, line 28: it would be nice to assert that range_check is not present in the output I think

mkaput (changes requested):

protostar/commands/test/test_cases.py
kasperski95: `self.execution_resources` refers to `execution_resources` from the 24. There's no difference between them.

https://docs.python.org/3/library/dataclasses.html

mkaput: 🤦🏻 misunderstood the code

PR #395, opened 2022-06-24
Fully migrate to the new cheatcodes registration flow

kasperski95:


protostar/commands/test/cheatcodes/expect_events_cheatcode.py
```diff
@@ -0,0 +1,60 @@
...
+        syscall_dependencies: Cheatcode.SyscallDependencies,
+        starknet: "ForkableStarknet",
+        test_execution_environment: "TestExecutionEnvironment",
+    ):
+        super().__init__(syscall_dependencies)
+        self.starknet = starknet
+        self.test_execution_environment = test_execution_environment
+
+    @property
+    def name(self) -> str:
```
Arcticae, line 28: this could be static, am i right?

Arcticae (approved):
Good job! Those are looking really clean

protostar/commands/test/cheatcodes/expect_events_cheatcode.py
kasperski95: Cheatcodes must implement common interface (abstract class) in order to easily register them (`hint_locals[cheatcode.name] = cheatcode.build()`). Theoretically this property could be static, but there would be a problem with defining abstract Cheatcode class and I using metaclasses just to make propery static would be an overkill.

> “Metaclasses are deeper magic than 99% of users should ever worry about. If you wonder whether you need them, you don’t (the people who actually need them know with certainty that they need them, and don’t need an explanation about why).”

Arcticae: Understandable - just nitpicking

PR #400, opened 2022-06-27
Disable saving PreprocessedProgram in the memory

kasperski95:
In large projects, Protostar was slowing down and then froze. The CPU usage was low, and the memory usage was high.

When I disabled saving the `PreprocessedProgram` in the memory, Protostar run tests at steady pace and used 100% of the CPU.

This PR can potentially fix the race condition issue reported internally by @Arcticae since I wasn't able to replicate it.

NOTE: Modifying codebase during test running will have the impact on test results.

(I think we should extract test case using regex because preprocessing is too slow, and I want to avoid changing too much logic.)


mkaput:
> I think we should extract test case using regex because preprocessing is too slow, and I want to avoid changing too much logic.

Not sure I follow the details here, but I think you could run a custom pass manager with parser, module collector and maybe identifier&struct collectors. That way you could reliably extract all symbols from codebase without having to perform full compilation.

kasperski95:
> 

I think we already do it. In bigger projects, preprocessing takes ~15 seconds per test suite. However in this PR I'll only disable caching preprocessed program in the memory.

```cairo
 def preprocess_contract(
        self, *cairo_file_paths: Path
    ) -> StarknetPreprocessedProgram:
        pass_manager = self.get_starknet_pass_manager()

        try:
            codes = [
                (cairo_file_path.read_text("utf-8"), str(cairo_file_path))
                for cairo_file_path in cairo_file_paths
            ]
            context = PassManagerContext(
                start_codes=[],
                codes=codes,
                main_scope=MAIN_SCOPE,
                identifiers=IdentifierManager(),
            )
            pass_manager.run(context)
            assert isinstance(context.preprocessed_program, StarknetPreprocessedProgram)
            return context.preprocessed_program
```

protostar/commands/test/test_collector.py
```diff
@@ -299,6 +299,7 @@ def _build_test_suite_from_test_suite_info(
         self,
         test_suite_info: TestSuiteInfo,
     ) -> TestSuite:
+
```
mkaput, line 302: ?

mkaput (approved):

PR #402, opened 2022-06-28
Add `account-contract` flags

kasperski95:


mkaput (approved):
Looks good, but I would document that the preferred path is to declare this in TOML instead of passing to command line on each invocation.

kasperski95:
> Looks good, but I would document that the preferred path is to declare this in TOML instead of passing to command line on each invocation.

A user needs to define configuration profiles to use this flag effectively in the TOML file, and I don't want to explain how to do it in the error message.

The configuration file was designed in a such way that ~any argument of any command~ any not required argument of any command can be configured there and I assume a user knows that.
https://docs.swmansion.com/protostar/docs/tutorials/project-initialization#command-configuration

PR #405, opened 2022-06-30
Allow disabling hint validation for tested files

kasperski95:
Added `disable-hint-validation` flag to the test command. It modifies `declare` cheatcode behavior.


website/docs/cli-reference.md
```diff
@@ -110,6 +110,9 @@ A glob or globs to a directory or a test suite, for example:
 Compile as account contract.
 #### `--cairo-path DIRECTORY[]`
 Additional directories to look for sources.
+#### `--disable-hint-validation`
+Disable hint validation in contracts declared by the `declare` cheatcode.
```
MaksymilianDemitraszek, line 114: We may add info about `deploy_contract` here too

protostar/commands/test/test_execution_environment.py
```diff
@@ -180,7 +187,10 @@ def build_cheatcodes(
             internal_calls: List[CallInfo],
         ) -> List[Cheatcode]:
             data_transformer = DataTransformerFacade(self._starknet_compiler)
-            declare_cheatcode = DeclareCheatcode(syscall_dependencies)
+            declare_cheatcode = DeclareCheatcode(
+                syscall_dependencies,
+                disable_hint_validation=self._disable_hint_validation_in_external_contracts,
```
MaksymilianDemitraszek, line 192: Wouldn't it be better to put it inside starknet_general_config? Would be consistent with `cairo_path` as this are both compiler args

MaksymilianDemitraszek: List of such populated arguments is very likely to grow in the future

protostar/commands/test/test_command.py
```diff
@@ -75,6 +75,13 @@ def arguments(self) -> List[Command.Argument]:
                 description="Compile as account contract.",
                 type="bool",
             ),
+            Command.Argument(
+                name="disable-hint-validation",
+                description=(
+                    "Disable hint validation in contracts declared by the `declare` cheatcode.\n"
```
MaksymilianDemitraszek, line 81: Can we add information about `deploy_contract` too? Probably `deploy_contract` is more popular cheatcode than declare.

MaksymilianDemitraszek (changes requested):
Looks good, but can we use starknet_config to populate the flag though? For the sake of consistency with cairo path.

protostar/commands/test/test_execution_environment.py
kasperski95: As discussed, moved cairo_path as a dependency of `DeclareCheatcode`. Removed `CheatableStarknetGeneralConfig` because it was no longer used.

MaksymilianDemitraszek (approved):

kasperski95:
GitHub merged the branch but didn't close this PR. GitHub plz...

PR #408, opened 2022-07-01
Added `store` cheatcode

MaksymilianDemitraszek:
- [x] Add docs

Transformer support will be added in a separate PR


protostar/commands/test/cheatcodes/store_cheatcode.py
```diff
@@ -0,0 +1,80 @@
...
+        target_contract_address: int,
+        variable_name: str,
+        value: List[int],
+        key: Optional[List[int]] = None,
+    ):
+        key = key or []
+        variable_address = self.calc_address(variable_name, key)
+        if target_contract_address == self.contract_address:
+            for i, val in enumerate(value):
+                self.store_local(variable_address + i, val)
```
kasperski95, line 31: Can you move the for loop to the `store_local` and add `store_remote` method?

kasperski95 (approved):
LGTM

protostar/commands/test/cheatcodes/store_cheatcode.py
MaksymilianDemitraszek: Yes we should

PR #417, opened 2022-07-05
Notify user about a new available Protostar version

kasperski95:
- added e2e test for `protostar upgrade`
- made the Protostar check the new version in non-blocking way
- works in offline-mode

Screenshot 2022-07-06 at 14 49 37


protostar/upgrader/upgrade_info_writer_thread.py
```diff
@@ -0,0 +1,44 @@
...
+                )
+
+        except ConnectionError:
+            pass
+
+    def __enter__(self):
+        self._thread.start()
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        pass
```
mkaput, line 44: This code is not waiting for the thread to be joined. Unless overwriting TOML is atomic operation (which I guess isn't), this may break toml contents.

mkaput (changes requested):

protostar/upgrader/upgrade_remote_checker.py
```diff
@@ -0,0 +1,42 @@
...
+        changelog_url = "https://github.com" + response_dict["update_url"]
+        return UpgradeRemoteChecker.Result(
+            latest_version=latest_version,
+            latest_release_tag=latest_release_tag,
+            is_newer_version_available=latest_version
+            > (
+                self._version_manager.protostar_version or VersionManager.parse("0.0.0")
+            ),
+            changelog_url=changelog_url,
+        )
```
mkaput, line 42: 1. This can be a plain function IMHO
2. Would it make sense to make this `async`?

protostar/upgrader/upgrade_toml.py
```diff
@@ -0,0 +1,55 @@
...
+    ProtostarDirectory,
+    VersionManager,
+    VersionType,
+)
+
+
+@dataclass(frozen=True)
+class UpgradeTOML:
+    """
+    Pytonish representation of the upgrade.toml.
```
mkaput, line 17:
```suggestion
    Pythonic representation of the upgrade.toml.
```

protostar/upgrader/upgrade_manager.py
```diff
@@ -0,0 +1,189 @@
...
+    VersionManager,
+    VersionType,
+)
+
+
+class UpgradeManagerException(Exception):
+    pass
+
+
+class UpgradeManager:
```
mkaput, line 23: This class also looks like it could just be a function `upgrade`?

protostar/upgrader/upgrade_info_writer_thread.py
```diff
@@ -0,0 +1,44 @@
+from threading import Thread
+
+# pylint: disable=redefined-builtin
+from requests.exceptions import ConnectionError
+
+from protostar.upgrader.upgrade_remote_checker import UpgradeRemoteChecker
+from protostar.upgrader.upgrade_toml import UpgradeTOML
+
+
+class UpgradeInfoWriterThread:
```
mkaput, line 10: I wonder if this class wouldn't be simpler are more robust if it would check remote. asynchronously in the background (possibly using asyncio features rather than daemon thread), and then perform file write in `__exit__` method, **synchronously** when program ends.

Generally this looks like a thing to be possibly written using `@asynccontextmanager`.

mkaput (changes requested):

mkaput:
I love the fact that you try to cache network query results in a file!

protostar/upgrader/upgrade_remote_checker.py
kasperski95: 1.  Classes that would use such a function would have to provide `VersionManager` and they would have to pass it to that function, which shouldn't be a responsibility of these classes. This is problematic when dependencies changes. 
2. It would.

protostar/upgrader/upgrade_info_writer_thread.py
kasperski95: It could. I'll take a look on contextmanager stuff because I didn't use it before.

protostar/upgrader/upgrade_manager.py
kasperski95: All dependencies of `UpgradeManager` would become dependencies of `UpgradeCommand` (and other classes that could possible use this function) and that generates boilerplate.

protostar/upgrader/upgrade_info_writer_thread.py
mkaput: Yup, that's what I also think and I have written the same in later comment 😅 

Resolving this one

protostar/upgrader/upgrade_info_writer_thread.py
kasperski95: I tried to use asyncio tasks without success. I end up leaving thread hence I left it as a class and didn't use `@asynccontextmanager`.

The task created by the `create_task` didn't start. I think the event loop on the main thread was busy, and the task never had a chance to run.

protostar/upgrader/upgrade_toml.py
```diff
@@ -0,0 +1,65 @@
...
+    ProtostarDirectory,
+    VersionManager,
+    VersionType,
+)
+
+
+@dataclass(frozen=True)
+class UpgradeTOML:
+    """
+    Pythonic representation of the upgrade.toml.
```
MaksymilianDemitraszek, line 18: I think it is a bit cryptic name for a class and suggests it is an upgrade configuration. `LatestVersionCacheTOML`? WDYT?

protostar/upgrader/upgrade_checker.py
```diff
@@ -0,0 +1,80 @@
...
+from logging import Logger
+from typing import Optional
+
+from protostar.upgrader.upgrade_remote_checker import UpgradeRemoteChecker
+from protostar.upgrader.upgrade_toml import UpgradeTOML
+from protostar.utils.log_color_provider import LogColorProvider
+from protostar.utils.protostar_directory import ProtostarDirectory, VersionManager
+
+
+class UpgradeChecker:
```
MaksymilianDemitraszek, line 11:
```suggestion
class LatestVersionChecker:
```
WDYT?

protostar/protostar_cli.py
```diff
@@ -174,28 +197,31 @@ def _check_git_version(self):
             )
 
     async def run(self, args: Any) -> None:
-        logger = self._setup_logger(args.no_color)
+        self._setup_logger(args.no_color)
         has_failed = False
-
         try:
+            await self.upgrade_checker.check_for_upgrades_if_necessary()
```
MaksymilianDemitraszek, line 203:
```suggestion
            await self.upgrade_checker.check_for_upgrades()
```
I think this postfix is not very informative.

protostar/upgrader/upgrade_remote_checker.py
```diff
@@ -0,0 +1,44 @@
...
+    Call a remote endpoint to figure out if the new Protostar version is available.
+    """
+
+    PROTOSTAR_REPO = "https://github.com/software-mansion/protostar"
+
+    @dataclass
+    class Result:
+        latest_release_tag: str
+        latest_version: VersionType
+        is_newer_version_available: bool
```
MaksymilianDemitraszek, line 19: It feels weird that this class compares versions and returns this boolean, but it's a minor thing, don't need to change.

protostar/upgrader/upgrade_toml.py
MaksymilianDemitraszek: Maybe `protostar_cache.toml`? IDK I am open for ideas just `upgrade.toml` feels bad

protostar/upgrader/upgrade_checker.py
```diff
@@ -0,0 +1,80 @@
...
+                        (
+                            "To install the latest Protostar version, run "
+                            f"{bold(colorize('CYAN', 'protostar upgrade'))}."
+                        ),
+                        "",
+                    ]
+                )
+            )
+
+    async def _update_upgrade_toml_if_necessary(
```
MaksymilianDemitraszek, line 62: `_if_necessary` seems unnecessary wordy

protostar/protostar_cli.py
MaksymilianDemitraszek: Wouldn't be better to make this check after the command? First, user wouldn't have to wait for it before he sees what he wants, second it won't dissapear under command output. I think `pip` and `gh` print notifications at the bottom.

protostar/upgrader/upgrade_checker.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if upgrade_toml is None or upgrade_toml.next_check_datetime < datetime.now():
+            self._logger.info("Checking for updates...")
+            result = await self._upgrade_remote_checker.check()
+
+            if result is not None:
+                new_upgrade_toml = UpgradeTOML(
+                    version=result.latest_version,
+                    changelog_url=result.changelog_url,
+                    next_check_datetime=datetime.now() + timedelta(days=3),
```
MaksymilianDemitraszek, line 75: It's nice you did TTL here instead last check time 👍

MaksymilianDemitraszek (changes requested):
It's very nice but we may need to update naming + behaviour.

Aside I think we have to add `protostar disable-notifications` or similar which will set cache to infinite time. Some users may don't want protostar to do this check. Don't have to do this in this PR though. Can you create task for it?

protostar/upgrader/upgrade_info_writer_thread.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import Optional
+
+# pylint: disable=redefined-builtin
+from requests.exceptions import ConnectionError
+
+from protostar.upgrader.upgrade_remote_checker import UpgradeRemoteChecker
+from protostar.upgrader.upgrade_toml import UpgradeTOML
+
+
+class UpgradeInfoWriterThread:
```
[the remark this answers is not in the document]
kasperski95: This class is already removed.

protostar/upgrader/upgrade_remote_checker.py
kasperski95: Actually I think the same. I'll remove it.

protostar/upgrader/upgrade_checker.py
kasperski95: Resolving. Explained the purpose of using "if_necessary" in the previous thread.

protostar/protostar_cli.py
kasperski95: Outvoted. I'll remove it.

MaksymilianDemitraszek: If it's a convention it makes kinda sense to name it this way. It is weird because technically you shouldn't write methods like this. If you need to include such postfix you break single respoinsibility rule 90% of the time.

What do you think about writing it like this:

self.upgrade_checker.load_chached_info() or (await self.upgrade_checker.load_remote_info() )
self.log_new_version_info()

this way you make stuff explicit and don't hide log underneath check.
You can add assert in log to make it secure.

kasperski95: That's better. I'll split the method.

protostar/upgrader/upgrade_info_writer_thread.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: leftovers from not submitted review

protostar/protostar_cli.py
kasperski95: It's a bit more complicated than that. I hide ifs in the `run` method. `run` is less explicit but not confusing.

kasperski95:
@MaksymilianDemitraszek https://github.com/software-mansion/protostar/issues/456

kasperski95:
A user can modify the cached file manually if they need to disable checking for upgrades.

MaksymilianDemitraszek (approved):

protostar/upgrader/latest_version_checker.py
```diff
@@ -0,0 +1,108 @@
...
+                    (
+                        "To install the latest Protostar version, run "
+                        f"{bold(colorize('CYAN', 'protostar upgrade'))}."
+                    ),
+                    "",
+                ]
+            )
+        )
+
+    async def load_local_latest_version_cache_toml(
```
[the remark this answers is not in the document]
kasperski95: There's no reason. Removing it.

PR #425, opened 2022-07-07
Detect account-contracts

kasperski95:
Protostar will try to again compiling the contract if the account contract error is detected.

### DIFF
--- a/protostar/utils/starknet_compilation.py
+++ b/protostar/utils/starknet_compilation.py
@@ -7,14 +7,17 @@ from starkware.cairo.lang.compiler.cairo_compile import get_module_reader
 from starkware.cairo.lang.compiler.constants import MAIN_SCOPE
 from starkware.cairo.lang.compiler.identifier_manager import IdentifierManager
 from starkware.cairo.lang.compiler.preprocessor.pass_manager import (
-    PassManager, PassManagerContext)
-from starkware.cairo.lang.compiler.preprocessor.preprocessor_error import \
-    PreprocessorError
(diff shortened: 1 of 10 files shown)


protostar/utils/starknet_compilation.py
```diff
@@ -67,37 +67,35 @@ def preprocess_contract(
...
             )
-            assert isinstance(assembled, ContractClass)
-            return assembled
         except PreprocessorError as err:
-            err_message = err.message
-            if isinstance(err_message, str) and "account_contract" in err_message:
-                raise ProtostarException(
-                    err_message.replace("account_contract", "account-contract")
-                ) from err
+            if isinstance(err.message, str) and "account_contract" in err.message:
```
MaksymilianDemitraszek, line 81: Can we add info about it to documentation? I think it's worth mentioning

MaksymilianDemitraszek (changes requested):

protostar/utils/starknet_compilation.py
kasperski95: Should user care about that? What's should be in the docs?

MaksymilianDemitraszek: That we automatically detect that, users can come from starknet-compiler which requires this flag.

MaksymilianDemitraszek: I know, but it exists, please provide it as a ::tip or something

MaksymilianDemitraszek (approved):

PR #432, opened 2022-07-08
Speed up test collecting phase (x6)

kasperski95:


protostar/utils/starknet_compilation.py
```diff
@@ -39,6 +40,40 @@ def get_starknet_pass_manager(self) -> PassManager:
             disable_hint_validation=self.disable_hint_validation,
         )
 
+    def get_main_identifiers_in_file(self, cairo_file_path: Path) -> List[str]:
+        pass_manager = self.get_starknet_pass_manager()
```
MaksymilianDemitraszek, line 44: Unfortunately this is not enough 😞 , counterexample:
```
func test_a{
    syscall_ptr : felt*, range_check_ptr, pedersen_ptr : HashBuiltin*
}():
    assert 0 = 0
    return ()
end

@external
func test_b{
    syscall_ptr : felt*, range_check_ptr, pedersen_ptr : HashBuiltin*
}():
    assert 0 = 0
    return ()
end
```

What if we included a few more passes:
- "directives_collector"
- "struct_collector"
- "preprocessor"

Then we could just rely on a ABI and one less starknet coupling to maintain. What would be an impact of that if not that big maybe it is worth? If it is too big we probably have to filter functions ids based on the external decorator, but I am not sure if it will cover everything, I would prefer to avoid it tbh.

MaksymilianDemitraszek (changes requested):

MaksymilianDemitraszek (approved):

PR #434, opened 2022-07-11
Added exit_first flag

Radinyn:
Added exit_first flag


protostar/commands/test/test_results_queue.py
```diff
@@ -1,17 +1,26 @@
...
+from protostar.commands.test.test_cases import TestCaseResult, PassedTestCase
 
 if TYPE_CHECKING:
     import queue
 
 
 class TestResultsQueue:
-    def __init__(self, shared_queue: "queue.Queue[TestCaseResult]") -> None:
+    def __init__(
+        self, shared_queue: "queue.Queue[TestCaseResult]", failed: ValueProxy
```
kasperski95, line 12: `failed` is a bit too ambiguous name. Maybe `has_failed_or_broken_result_shared_value`? (It's a bit long, but descriptive.)

Btw: https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/shared-values/

protostar/commands/test/test_scheduler.py
```diff
@@ -28,9 +29,16 @@ def run(
...
     ):
         with multiprocessing.Manager() as manager:
-            test_results_queue = TestResultsQueue(manager.Queue())
+            test_results_queue = TestResultsQueue(
+                manager.Queue(),
+                manager.Value(
+                    ctypes.c_bool,
+                    (len(test_collector_result.broken_test_suites) > 0) and exit_first,
+                ),
+            )
```
kasperski95, line 41: It's hard to tell what's the purpose of this value. Please pass it as a keyword argument.

protostar/commands/test/testing_live_logger.py
```diff
@@ -14,11 +14,23 @@
...
         self._no_progress_bar = no_progress_bar
         self.testing_summary = testing_summary
+        self.exit_first = exit_first
+
+    def exit_before_log(self, test_collector_result: "TestCollector.Result") -> None:
+        self.testing_summary.log(
+            logger=self._logger,
+            collected_test_cases_count=test_collector_result.test_cases_count,
+            collected_test_suites_count=len(test_collector_result.test_suites),
+        )
```
kasperski95, line 33:
```suggestion
    def log_testing_summary(self, test_collector_result: "TestCollector.Result") -> None:
        self.testing_summary.log(
            logger=self._logger,
            collected_test_cases_count=test_collector_result.test_cases_count,
            collected_test_suites_count=len(test_collector_result.test_suites),
        )
```

The name of this function is misleading. Now you can reuse this function in line 76.

protostar/commands/test/test_scheduler.py
```diff
@@ -42,6 +50,11 @@ def run(
                 for test_suite in test_collector_result.test_suites
             ]
 
+            # A test case was broken
+            if exit_first and test_results_queue.failed():
+                self._live_logger.exit_before_log(test_collector_result)
+                return
```
kasperski95, line 56: I'd move this check out of `TestLogger` to the `TestCommand` because at this point broken results come from `TestCollector`. You can skip instantiating  `TestScheduler`.

protostar/commands/test/test_results_queue.py
```diff
@@ -1,17 +1,26 @@
...
     def get(self) -> TestCaseResult:
         return self._shared_queue.get(block=True, timeout=1000)
 
     def put(self, item: TestCaseResult) -> None:
+        if not isinstance(item, PassedTestCase):
+            self._failed.value = True
         self._shared_queue.put(item)
+
+    def failed(self) -> bool:
+        return self._failed.value
```
kasperski95, line 26:
```suggestion
    @property
    def has_failed_or_broken_result(self) -> bool:
        return self._failed.value
```

Compare:
- `queue.failed` — (how can a queue fail?)
- `queue.has_failed_or_broken_result`

kasperski95 (changes requested):
Logic seems good. Mostly comments about the code style.

protostar/commands/test/test_scheduler.py
Radinyn: It has to be here because we need the logger instanced to make the output look good - it returns instantly anyway.

kasperski95 (approved):

protostar/commands/test/test_command.py
```diff
@@ -88,6 +88,12 @@ def arguments(self) -> List[Command.Argument]:
                 type="bool",
                 description="Disable progress bar.",
             ),
+            Command.Argument(
+                name="exit-first",
+                short_name="x",
+                type="bool",
+                description="Exit at the first failure.",
```
MaksymilianDemitraszek, line 95:
```suggestion
                description="Exit instantly on first broken or failed test",
```

protostar/commands/test/test_results_queue.py
```diff
@@ -1,17 +1,28 @@
...
+    ) -> None:
         self._shared_queue = shared_queue
+        self._any_failed_or_broken_shared_value = any_failed_or_broken_shared_value
 
     def get(self) -> TestCaseResult:
         return self._shared_queue.get(block=True, timeout=1000)
 
     def put(self, item: TestCaseResult) -> None:
+        if not isinstance(item, PassedTestCase):
+            self._any_failed_or_broken_shared_value.value = True
```
MaksymilianDemitraszek, line 24: I think it is a bit confusing that queue sets a global value. Generally you don't expect that whent you put something on a queue it will change some global flag :p. 

Generally I think we should avoid introducing global state when multiprocessing at all cost, it makes code hard to reason about. 

What do you think about instead of introducing this shared value we would throw an custom error in logger if exit_first flag is set and terminate processes accordingly? 

I know it is weird that logger would handle this as it supposed to log stuff, but I think we will need to gravitate in the future towards making it more of a like results collector with wider responsibilites. You can play around with naming if you want a bit but generally my point is we should keep it in the main process.

Does that make sense?

MaksymilianDemitraszek (changes requested):
That's a very good job. I think we should reconsider introducing a new shared state as it may not scale very well.

protostar/commands/test/test_results_queue.py
Radinyn: I don't think that flag is particularly controversial, personally I don't find it too different from things like queue.empty(), after all we are only keeping information about it's current state. Changing the logger into something more elaborate might introduce more confusion than a single state variable.

MaksymilianDemitraszek: It's generally best to avoid shared state when dealing with multiprocessing. In this case it does not much mess as it is only one flag, but it forces you to keep it in mind when you work on this code. What if we increased number of variables (which is very probable to happen in the future)? We are not in rush or anything so I would prefer to do this in a future-proof, stateless way.

I think would be nice to have 
```
class TestResultsCoordinator
    live_logger: TestLiveLogger
    
    def listen():
        handles exit first
        passes to live logger
```
or something similar to avoid confision you mentioned.

Also setting it when you put on the queue is definitely incorrect, this one has to be changed, it is a drity side effect nobody expects.

Radinyn: I don't think it is possible to create the TestResultsCoordinator without completely overhauling the TestingLiveLogger as we still need the queue's elements to check whether we found a failed/broken one or not.  I can rewrite the Logger into a Collector (with a separate logger that has nothing to do with the queue or with a logging function) if you want but I just realised that this is what our Queue actually is at the moment - a stateful Collector - perhaps this is something worth embracing, something here has to store the state after all.

MaksymilianDemitraszek: As we discussed on slack, resolving

MaksymilianDemitraszek (approved):

website/docs/cli-reference.md
```diff
@@ -113,6 +113,8 @@ Additional directories to look for sources.
 #### `--disable-hint-validation`
 Disable hint validation in contracts declared by the `declare` cheatcode or deployed by `deploy_contract` cheatcode.
 
+#### `-x` `--exit-first`
+Exit instantly on first broken or failed test
```
mkaput, line 117: immediately 😉

mkaput: wow I'm impressed.

PR #438, opened 2022-07-11
Added command timing

Radinyn:
Added command timing, because command classes are abstracted I added logging to the ProtostarCLI object destructor.

### DIFF
--- a/protostar/protostar_cli.py
+++ b/protostar/protostar_cli.py
@@ -2,7 +2,8 @@
 import sys
 from logging import INFO, Logger, StreamHandler, getLogger
 from pathlib import Path
-from typing import Any
+from typing import Any, Optional
+import time
 
 from protostar.cli import CLIApp, Command
 from protostar.commands import (
@@ -80,8 +81,12 @@ class ProtostarCLI(CLIApp):
         protostar_toml_writer: ProtostarTOMLWriter,
         protostar_toml_reader: ProtostarTOMLReader,
         requester: InputRequester,
+        start_time: float = 0.0,
+        logger: Optional[Logger] = None,
     ) -> None:
         self.project = project
+        self.start_time = start_time
+        self.logger = logger
 
         super().__init__(
             commands=[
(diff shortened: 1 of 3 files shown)


protostar/protostar_cli.py
```diff
@@ -80,8 +81,12 @@ def __init__(
         protostar_toml_writer: ProtostarTOMLWriter,
         protostar_toml_reader: ProtostarTOMLReader,
         requester: InputRequester,
+        start_time: float = 0.0,
```
MaksymilianDemitraszek, line 84: Why do we have to make this one being custom value? Can't we just run `= time.perf_counter()` in constructor?

protostar/protostar_cli.py
```diff
@@ -80,8 +81,12 @@ def __init__(
...
+        logger: Optional[Logger] = None,
```
MaksymilianDemitraszek, line 85: Why do logger is optional?

protostar/protostar_cli.py
```diff
@@ -185,7 +192,14 @@ async def run(self, args: Any) -> None:
...
                 print(err.details)
-            logger.error(err.message)
+            assert self.logger is not None
+            self.logger.error(err.message)
             sys.exit(1)
         except KeyboardInterrupt:
             sys.exit(1)
+
+    def __del__(self):
+        if self.logger is not None:
```
MaksymilianDemitraszek, line 202: I think use of destuctor is a bit awkward, there is not that much control when it runs. What if we introduced an method and called it explicitly from main function after `asyncio.run(protostar_cli.run(parser.parse()))` finishes? Is it possible?

MaksymilianDemitraszek (changes requested):
Good job, couple nitpicks

protostar/protostar_cli.py
MaksymilianDemitraszek: Wouldn't be better to just pass `SilentLogger` then or something? This way you have to assert and if it away everywhere which is messy as we log a lot

MaksymilianDemitraszek: You can keep argument optional but i think you should set default then to keep property not an optional one

PR #445, opened 2022-07-13
Auto add issue to TODOs when opened

Arcticae:


.github/workflows/assign_issues_to_project.yml
```diff
@@ -0,0 +1,19 @@
...
+  assign_one_project:
+    runs-on: ubuntu-latest
+    name: Assign to Project
+    steps:
+    - name: Assign new issues to the project board
+      uses: srggrs/assign-one-project-github-action@1.2.1
+      if: github.event.action == 'opened'
+      with:
+        project: 'https://github.com/orgs/software-mansion/projects/2'
+        column_name: 'TODO'
```
MaksymilianDemitraszek, line 19: I wonder if this shouldn't be backlog.

MaksymilianDemitraszek (approved):

.github/workflows/assign_issues_to_project.yml
MaksymilianDemitraszek: I mean not every issue we open is TODO, maybe it should move to TODO if somebody has been assigned to it?

Arcticae: IMHO i think we should at least look at it before we throw it into the abyss of backlog :P

MaksymilianDemitraszek: maybe that's a good point

PR #458, opened 2022-07-13
Added std capture and it's corresponding flag

Radinyn:


MaksymilianDemitraszek (changes requested):
Very good jobs, requesting a few tweaks in comments

protostar/commands/test/test_cases.py
```diff
@@ -22,6 +22,7 @@ def get_formatted_file_path(self):
 class PassedTestCase(TestCaseResult):
     test_case_name: str
     execution_resources: Optional[ExecutionResourcesSummary]
+    logs: str
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: And as @kasperski95 we have logger, it is name collision

protostar/commands/test/test_cases.py
```diff
@@ -17,13 +19,19 @@ class TestCaseResult:
...
 
 @dataclass(frozen=True)
 class PassedTestCase(TestCaseResult):
     test_case_name: str
     execution_resources: Optional[ExecutionResourcesSummary]
+    captured_setup_stdout: str
+    captured_test_stdout: str
 
-    def __str__(self) -> str:
+    def log(self, include_stdout_section: bool = False) -> str:
```
mkaput, line 34: maybe?

```suggestion
    def display(self, include_stdout_section: bool = False) -> str:
```

by the name `log` I thought I can send custom logs from testing code with this function

protostar/commands/test/test_cases.py
```diff
@@ -75,14 +105,36 @@ def __str__(self) -> str:
...
+                    "[setup]:\n"
+                    f"{log_color_provider.colorize('GRAY', self.captured_setup_stdout)}\n"
+                )
+
+            if self.captured_test_stdout:
+                result.append(
+                    "[test]:\n"
+                    f"{log_color_provider.colorize('GRAY', self.captured_test_stdout)}\n"
+                )
+
```
mkaput, line 137: this section is repeated in PassedTestCase

protostar/commands/test/test_cases.py
```diff
@@ -17,13 +19,19 @@ class TestCaseResult:
     def get_formatted_file_path(self):
         return log_color_provider.colorize("GRAY", str(self.file_path))
 
+    @abstractmethod
+    def log(self, _: bool = False) -> str:
```
mkaput, line 23:
```suggestion
    def log(self, include_stdout_section: bool = False) -> str:
```

and please use the same argument name in each override.

you are breaking `kwargs` passing here. parameter names are parts of function contract in Python, following snippet would crash here:

```python
result = get_random_test_case_result()
result.log(include_stdout_section=True)
# or
result.log(_=True)
```

website/docs/cli-reference.md
```diff
@@ -110,14 +110,16 @@ Additional directories to look for sources.
 Disable hint validation in contracts declared by the `declare` cheatcode or deployed by `deploy_contract` cheatcode.
 
 #### `-x` `--exit-first`
-Exit instantly on first broken or failed test
+Exit immediately on first broken or failed test
```
mkaput, line 113: could you please move this change to separate PR?

mkaput (changes requested):

protostar/commands/test/test_cases.py
```diff
@@ -17,13 +19,47 @@ class TestCaseResult:
...
 
+    @abstractmethod
+    def display(self, include_stdout_section: bool = False) -> str:
+        ...
+
+
+@dataclass(frozen=True)
+class FinishedTestCase(TestCaseResult):
+    captured_setup_stdout: str
+    captured_test_stdout: str
```
mkaput, line 30: I think that it might be nice to have this model generalized over a possibility of having 0:* stdouts captured in test case. This way, it would be natural to write generic code over `setup` and `test` stages.

I'm thinking about following data model:

```python
class PassedTestCase:
    stdouts: List[OutputCapture]

@dataclass
class OutputCapture
    # typing pseudocode for brevity
    stream: "stdout" | "stderr" # possible future addition
    stage: str # "setup" or "test" but can be arbitrary
    data: bytes
```

It is also trivial to express interleaving of stdout/stderr ✖️ setup/test/whatever captures here. You just insert `OutputCapture`s for each captured line, one by one as encountered` to the list.

mkaput: I do wonder if we actually have to capture `setup` stdout ? I haven't ever needed such feature

mkaput: discussed privately, let's leave current implementation and consider this refactoring if we stumble upon a need for it

protostar/commands/test/test_cases.py
MaksymilianDemitraszek: Problem is we use `log` a lot as a name for such functions, I am worried about consistency. Agreed `display` is a better name though.

MaksymilianDemitraszek: Oh i didn't see nvm then

protostar/commands/test/test_cases.py
```diff
@@ -7,27 +8,53 @@
 )
 from protostar.commands.test.test_environment_exceptions import ReportedException
 from protostar.protostar_exception import UNEXPECTED_PROTOSTAR_ERROR_MSG
-from protostar.utils.log_color_provider import log_color_provider
+from protostar.utils.log_color_provider import log_color_provider, SupportedColorName
+
+
+def _get_formatted_stdout(
```
mkaput, line 14: nitpick: could you please move these underscored functions to bottom of the file, so that the meat of this module stays on top? :)

protostar/commands/test/test_command.py
```diff
@@ -122,6 +128,7 @@ async def test(
         no_progress_bar=False,
         fast_collecting=False,
         exit_first=False,
+        stdout_on_success=False,
```
mkaput, line 131:
```suggestion
        stdout_on_success: bool = False,
```

protostar/commands/test/test_command.py
```diff
@@ -97,6 +97,11 @@ def arguments(self) -> List[Command.Argument]:
                 type="bool",
                 description="Exit immediately on first broken or failed test",
             ),
+            Command.Argument(
+                name="stdout-on-success",
+                type="bool",
+                description="Also print captured stdout for passed test cases.",
```
mkaput, line 103:
```suggestion
                description="Print captured standard output for passed test cases.",
```

mkaput (changes requested):

protostar/commands/test/test_command.py
Radinyn: This "also" is here to show that the default behaviour is that we always print for failed test cases

protostar/commands/test/test_command.py
Radinyn: We don't do that anywhere else, is it necessary here?

mkaput (approved):

PR #460, opened 2022-07-14
Fix data transformer in expect events

kasperski95:
Data transformer didn't work in `expect_events` when a contract was deployed in the __setup__ function.

This problem also exists in the `mock_call` however, this fix can't be applied to `mock_call`.


mkaput (approved):

kasperski95:
Cancelling. It could be fixed by the following PR: https://github.com/software-mansion/protostar/pull/461

PR #464, opened 2022-07-14
Make all exceptions inherit from Exception instead of BaseException

mkaput:


MaksymilianDemitraszek (approved):

kasperski95 (approved):
> Derive exceptions from Exception rather than BaseException. Direct inheritance from BaseException is reserved for exceptions where catching them is almost always the wrong thing to do.

protostar/starknet/execution_environment.py
```diff
@@ -49,3 +58,25 @@ def set_cheatcodes(cheatcode_factory: CheatcodeFactory):
...
+    """
+
+    if isinstance(ex, ReportedException):
+        return ex
+
+    if isinstance(ex, VmException):
+        return _walk_inner_exc_chain_for_reported_ex(ex.inner_exc)
+
+    if isinstance(ex, HintException):
+        return _walk_inner_exc_chain_for_reported_ex(ex.inner_exc)
```
kasperski95, line 77:
```suggestion
    if isinstance(ex, (VmException, HintException)):
        return _walk_inner_exc_chain_for_reported_ex(ex.inner_exc)
```

kasperski95 (approved):
LGTM

Radinyn (approved):

mkaput:
I have analysed use cases here, and I have come to a conclusion that it's actually OK for `ReportedException` to be descendant of `BaseException` directly. The missing semantics that we need to document is to inform users that this exception is to be raised from cheatcodes and its purpose is to interrupt the Cairo VM. I think we may consider renaming `ReportedException` so something akin to `CheatcodeInterrupt` (in similar fashion to Python's `KeyboardInterrupt`). This is a topic for future PR.

I will effectively revert all changes that needed review since initial commit, so I am going to merge this now.

Issue #485, opened 2022-07-18
Refactor `StarknetCompiler`


MaksymilianDemitraszek:
Recent changes made a bit of mess in `StarknetCompiler` responsibilities, It should be altered to make its behaviour more predicatble.

MaksymilianDemitraszek:
StarknetCompiler is deprecated

PR #489, opened 2022-07-18
Improved default collection strategy

MaksymilianDemitraszek:
Performance boost compared to `--fast-collection` around 30%
Performance boost compared to default strategy ~10fold

It deprecates fast-collecting, I will remove it in a separate PR, currently only added an assert

### DIFF
--- a/protostar/commands/test/test_command.py
+++ b/protostar/commands/test/test_command.py
@@ -137,7 +137,7 @@ class TestCommand(Command):
         logger = getLogger()
 
         include_paths = self._build_include_paths(cairo_path or [])
-        assert not fast_collecting
+        assert not fast_collecting, "`--fast-collecting` is deprecated, use default strategy"
         with ActivityIndicator(log_color_provider.colorize("GRAY", "Collecting tests")):
             test_collector_result = TestCollector(
                 StarknetCompiler(
--- a/protostar/commands/test/test_command.py
+++ b/protostar/commands/test/test_command.py
@@ -137,19 +137,14 @@ class TestCommand(Command):
         logger = getLogger()
 
         include_paths = self._build_include_paths(cairo_path or [])
-
+        assert not fast_collecting
         with ActivityIndicator(log_color_provider.colorize("GRAY", "Collecting tests")):
-            factory = (
-                StarknetPassManagerFactory
-                if fast_collecting
-                else TestCollectorPassManagerFactory
-            )
             test_collector_result = TestCollector(
                 StarknetCompiler(
                     config=CompilerConfig(
                         disable_hint_validation=True, include_paths=include_paths
                     ),
-                    pass_manager_factory=factory,
+                    pass_manager_factory=TestCollectorPassManagerFactory,
                 ),
                 config=TestCollector.Config(fast_collecting=fast_collecting),
             ).collect(
--- a/protostar/commands/test/test_command.py
+++ b/protostar/commands/test/test_command.py
@@ -10,6 +10,7 @@ from protostar.commands.test.test_scheduler import TestScheduler
 from protostar.commands.test.testing_live_logger import TestingLiveLogger
 from protostar.commands.test.testing_summary import TestingSummary
 from protostar.utils.compiler.pass_managers import (
+    StarknetPassManagerFactory,
     TestCollectorPassManagerFactory,
 )
 from protostar.utils.log_color_provider import log_color_provider
@@ -138,12 +139,17 @@ class TestCommand(Command):
         include_paths = self._build_include_paths(cairo_path or [])
 
         with ActivityIndicator(log_color_provider.colorize("GRAY", "Collecting tests")):
+            factory = (
+                StarknetPassManagerFactory
+                if fast_collecting
+                else TestCollectorPassManagerFactory
+            )
             test_collector_result = TestCollector(
                 StarknetCompiler(
                     config=CompilerConfig(
                         disable_hint_validation=True, include_paths=include_paths
                     ),
-                    pass_manager_factory=TestCollectorPassManagerFactory,
+                    pass_manager_factory=factory,
                 ),
                 config=TestCollector.Config(fast_collecting=fast_collecting),
             ).collect(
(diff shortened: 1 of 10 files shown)


tests/integration/compilation/collector_contract.cairo
```diff
@@ -0,0 +1,42 @@
+%lang starknet
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: const is included already

tests/integration/compilation/collector_contract.cairo
```diff
@@ -0,0 +1,49 @@
...
+
+    func increase_twice(amount : felt):
+    end
+
+    func get_balance() -> (res : felt):
+    end
+end
+
+namespace test_utils:
+    func internal_function{
```
kasperski95, line 23: Should the collector collect function if `@external` decorator is used here?

```cairo
namespace token {
  @external
  func test_transfering_ownership...
}

```

protostar/utils/compiler/pass_managers.py
```diff
@@ -38,19 +71,27 @@ def build(config: "CompilerConfig") -> PassManager:
...
+        manager.add_stage(
+            "module_collector",
+            ModuleCollector(
+                read_module=read_module,
+                additional_modules=[],
+            ),
+        )
+        collector_fac: Callable[
+            [PassManagerContext], Visitor
+        ] = lambda _: TestCollectorPreprocessor()
```
mkaput, line 86: by adding a `context` argument to `TestCollectorPreprocessor.__init__` you could omit this lambda trickery

protostar/utils/compiler/pass_managers.py
```diff
@@ -102,3 +143,61 @@ def visit_CodeElementFunction(self, elm: CodeElementFunction):
         if attr is not None:
             self.add_abi_storage_var_types(elm=attr)
             return
+
+
+class TestCollectorStage(VisitorStage):
```
mkaput, line 148: This stage is strictly bound to a particular Visitor. The pattern we did in Cairo compiler was to override `__init__` constructor to forcefully specify the visitor factory and `modify_ast` arguments.

protostar/utils/compiler/pass_managers.py
```diff
@@ -102,3 +143,61 @@ def visit_CodeElementFunction(self, elm: CodeElementFunction):
...
+
+class TestCollectorPreprocessor(Visitor):
+    def __init__(self):
+        super().__init__()
+        self.abi: AbiType = []
+
+    """
+    This preprocessor generates simpler and more limited ABI in exchange for performance.
+    ABI includes only function types with only names.
+    """
```
mkaput, line 173: order is wrong here 😅

protostar/utils/starknet_compilation.py
```diff
@@ -57,12 +58,15 @@ def build_codes(*cairo_file_paths: Path) -> List[Tuple[str, str]]:
 
     def preprocess_contract(
         self, *cairo_file_paths: Path
-    ) -> StarknetPreprocessedProgram:
+    ) -> Union[StarknetPreprocessedProgram, TestCollectorPreprocessedProgram]:
```
mkaput, line 61: It seems kinda counterintuitive that the `compiler` class does both compilation and/or test collection. And it's kinda bad that this broad scope leaks in types.

Would it make sense to move test collection outside this class? It seems like the building blocks are all static here so this should be easily doable.

mkaput (changes requested):

protostar/utils/starknet_compilation.py
MaksymilianDemitraszek: I have a separate task to refactor this StarknetCompiler class because it became a mess. I want to remove stuff releated to `fast-collection` first though.

MaksymilianDemitraszek: I want to keep size of the PR managable, can we move this request to #485 ?

protostar/utils/compiler/pass_managers.py
mkaput: You could remove these arguments from `__init__` and provide them directly here. Also you should rather call super() here instead of setting these fields manually

mkaput (changes requested):
Unresolved one comment.

protostar/commands/test/test_command.py
```diff
@@ -131,33 +137,35 @@ async def test(
...
         include_paths = self._build_include_paths(cairo_path or [])
-
+        factory = (
+            StarknetPassManagerFactory
+            if safe_collecting
+            else TestCollectorPassManagerFactory
+        )
+        assert (
+            not fast_collecting
+        ), "`--fast-collecting` is deprecated, use default strategy"
```
kasperski95, line 154: You can remove `fast-collecting` flag completely in this PR. The only thing left is `Command.Argument` and :::info in docs.

tests/integration/compilation/test_collector_compilation_test.py
```diff
@@ -0,0 +1,34 @@
...
+from pytest_mock import MockerFixture
+
+from protostar.utils.compiler.pass_managers import (
+    ProtostarPassMangerFactory,
+    TestCollectorPassManagerFactory,
+)
+from protostar.utils.starknet_compilation import CompilerConfig, StarknetCompiler
+
+
+def expected_collected_functions_oracle(path: Path):
```
kasperski95, line 12: oracle? What does it mean?

kasperski95 (approved):

mkaput (approved):

mkaput:
good job 👏🏻

tests/integration/compilation/test_collector_compilation_test.py
MaksymilianDemitraszek: https://en.wikipedia.org/wiki/Test_oracle

protostar/commands/test/test_command.py
MaksymilianDemitraszek: It's also a logic in TestCollector and StarknetCompilation, that's why I postponed it

PR #493, opened 2022-07-19
Performance tests

Arcticae:


protostar/commands/test/test_runner.py
```diff
@@ -127,7 +127,6 @@ async def _run_test_suite(
                 self.user_contracts_compiler,
                 test_contract,
             )
-
```
mkaput, line 130: Why was this line removed? 😀

tests/e2e/performance/test_performance_testing.py
```diff
@@ -1,55 +1,504 @@
...
+        from starkware.cairo.common.cairo_builtins import HashBuiltin
+        {imports}
+        {source}
+    """,
+        cases,
+    )
+
+
+@pytest.fixture(name="tmp_test_dir")
+def make_test_dir() -> Generator[Path, None, None]:
```
mkaput, line 56: Could you please use pattern 
- Fixture=`X`
- Function name=`X_fixture`

Here?

tests/e2e/performance/test_performance_testing.py
```diff
@@ -1,55 +1,504 @@
...
+    test_dir_path = SCRIPT_DIRECTORY / "tmp_test_files"
```
mkaput, line 57: Don't we have already a functionality for this? 🤨

tests/e2e/performance/test_performance_testing.py
```diff
@@ -1,55 +1,504 @@
...
+
+
+@pytest.fixture(name="protostar_cairo_path")
+def get_protostar_cairo_path(request) -> Path:
+    return (Path(request.node.fspath).parent.parent.parent.parent / "cairo").absolute()
+
+
+@contextmanager
+def timing(expected, minimum=None):
+    start_timer = time()
```
mkaput, line 75: https://docs.python.org/3/library/time.html I think this might be more precise thing to use for benchmarking

mkaput (commented):
Why not using https://pytest-benchmark.readthedocs.io/en/latest/index.html instead of manually writing benchmarking logic?

In the future it might be nice to let us maintain a living histogram of Protostars performance somewhere and let tests automatically download expected performance metrics instead of hardcoding these.

tests/e2e/performance/test_performance_testing.py
Arcticae: I don't think so - if it exists you can point me to one

Arcticae:
@mkaput 
> Why not using https://pytest-benchmark.readthedocs.io/en/latest/index.html instead of manually writing benchmarking logic?
> 
> In the future it might be nice to let us maintain a living histogram of Protostars performance somewhere and let tests automatically download expected performance metrics instead of hardcoding these.

That one doesn't support async functions - if it did i'd consider it

tests/e2e/performance/test_performance_testing.py
Arcticae: I actually think that the code is not that fast anyway, so timer resolution isn't critical here

mkaput: Right.

tests/e2e/performance/test_performance_testing.py
mkaput: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

mkaput:
> That one doesn't support async functions - if it did i'd consider it.

You can always wrap async code in [`asyncio.run`](https://docs.python.org/3/library/asyncio-task.html#asyncio.run) 😄 and you get async to sync integration. In this case, I think that'd be the way to go :)

> As for harcoding - it's a rare operation since i gave some breathing space for perf anyway, so a rise in time would mean we really want to fix the perf, not the time value i require here.

Iiii'm not entirely sure that this would be a meaningful metric for us (just a threshold when action is urgent). I think we might benefit more from having a review bot which says "This PR speeds up/slows down Protostar by X%". Such bot would either be executed when PR is ready for review, or for example just ran weekly (in that case it could even try to bisect and try to find offending commit in case of a slow-down by specified threshold).

MaksymilianDemitraszek (approved):
Very nice

pyproject.toml
```diff
@@ -53,11 +54,12 @@ lint = "pylint protostar"
 local_static_check = ["format", "lint", "type_check", "custom_checks"]
 profile = "python -m cProfile -o recent_profiling.prof protostar.py"
 show_prof = "snakeviz recent_profiling.prof"
-test = ["test_unit", "test_integration", "build", "test_e2e"]
+test = ["test_unit", "test_integration", "build", "test_e2e", "test_performance"]
```
kasperski95, line 57:
```suggestion
test = ["test_unit", "test_integration", "test_performance", "build", "test_e2e"]
```
I'd move performance tests before "build" to make the feedback loop faster.

kasperski95 (approved):

Arcticae:
Waiting for checks and going to merge.

PR #496, opened 2022-07-19
Add event ABI in default cheatcode implementation

### LINKED ISSUE #495: [BUG] Not able to record events in "unit_testing" approach
## Protostar version

| Name       | Version              |
| ---------- | -------------------- |
| Protostar  | `PROTOSTAR_VERSION`  |
| Cairo-lang | `CAIRO_LANG_VERSION` |
| Git        | `GIT_VERSION`        |

## Current Behavior
self = <protostar.commands.test.cheatcodes.expect_events_cheatcode.ExpectEventsCheatcode object at 0x110f1d848>
raw_expected_event = {'data': {'amount': 20, 'current_balance': 123}, 'name': 'increase_balance_called'}

    def _convert_raw_expected_event_to_expected_event(
        self,
        raw_expected_event: RawExpectedEventType,
    ):
    
        name: str
        data: Optional[
            Union[
                List[int],
                Dict[
                    DataTransformerFacade.ArgumentName,
                    DataTransformerFacade.SupportedType,
                ],
            ]
        ] = None
        from_address: Optional[int] = None
        if isinstance(raw_expected_event, str):
            name = raw_expected_event
        else:
            name = raw_expected_event["name"]
            if "data" in raw_expected_event:
                raw_data = raw_expected_event["data"]
                if isinstance(raw_data, collections.Mapping):
                    assert (
                        name in self.state.event_name_to_contract_abi_map
>                   ), "Couldn't map event name to the contract path with that event"
E                   AssertionError: Couldn't map event name to the contract path with that event

## Expected Behavior
It should not crash

Arcticae:
fixes #495

### DIFF
--- a/protostar/commands/test/cheatcodes/declare_cheatcode.py
+++ b/protostar/commands/test/cheatcodes/declare_cheatcode.py
@@ -6,7 +6,7 @@ from typing import Any, Callable
 from starkware.python.utils import from_bytes
 from starkware.starknet.business_logic.internal_transaction import InternalDeclare
 from starkware.starknet.testing.contract import DeclaredClass
-from starkware.starknet.testing.contract_utils import EventManager, get_abi
+from starkware.starknet.testing.contract_utils import get_abi
 
 from protostar.starknet.cheatcode import Cheatcode
 from protostar.utils.starknet_compilation import StarknetCompiler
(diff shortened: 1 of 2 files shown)


protostar/commands/test/cheatcodes/declare_cheatcode.py
```diff
@@ -6,7 +6,7 @@
 from starkware.python.utils import from_bytes
```
kasperski95, line 6: Please add the following test case for `expect_events` cheatcode:
`test_data_transformer_works_in_unit_testing`

kasperski95 (changes requested):

PR #507, opened 2022-07-21
Reorganise documentation

mkaput:
1. This PR reorganizes guides pages into separate sections, one level up.
2. Testing section has been split into many smaller pages.
3. Each cheatcode has its dedicated page now.
4. Sidebar is fully autogenerated from filesystem.
5. Replaced all absolute URL links with relative Markdown links, so that Docusaurus can maintain valid links througout documentation.
6. Added redirects to old URLs so that we do not break any existing blogposts.


website/docs/tutorials/07-testing/02-cheatcodes/03-expect-events.md
```diff
@@ -0,0 +1,72 @@
...
+                    ]],
+                    "from_address": NotRequired[int]
+                },
+            )],
+        ) -> None: ...
+```
+Compares expected events with events in the StarkNet state. You can use this cheatcode to test whether a contract emits specified events. Protostar compares events after a test case is completed. Therefore, you can use this cheatcode in any place within a test case.
+
+:::tip
+You can provide `"data"` as a dictionary to leverage [data transformer](/docs/tutorials/guides/testing#data-transformer).
```
kasperski95, line 30: Broken link

website/docs/tutorials/07-testing/02-cheatcodes/01-mock-call.md
```diff
@@ -0,0 +1,106 @@
+# `mock_call`
+
+```python
+def mock_call(contract_address: int, fn_name: str, ret_data: Union[List[int], Dict]) -> Callable: ...
+```
+
+Mocks all calls to function with the name `fn_name` of a contract with an address `contract_address` until the returned callable is called. Mocked call returns data provided in `ret_data`. Mock works globally, for all of the contracts, not only the testing contract.
+
+:::tip
+You can provide constructor arguments as a dictionary to leverage [data transformer](/docs/tutorials/guides/testing#data-transformer).
```
kasperski95, line 10: broken link

website/docs/tutorials/07-testing/02-cheatcodes/06-prepare.md
```diff
@@ -0,0 +1,33 @@
...
+    return ()
+end
+
+```
+:::info
+You can prepare multiple contracts from one `DeclaredContract`.
+:::
+
+:::tip
+You can provide `constructor_calldata` as a dictionary to leverage [data transformer](/docs/tutorials/guides/testing#data-transformer).
```
kasperski95, line 32: broken link

kasperski95 (commented):
Looks neat.

Cheatcodes order is weird. We should sort them by name or importance (which needs to be defined).

Arcticae:
Let me know when it's deployed - i will update my article then :)

mkaput:
> Cheatcodes order is weird. We should sort them by name or importance (which needs to be defined).

I have preserved existing order, but overall I am big 👍🏻 for lexical sorting. Doing so.

kasperski95 (approved):

mkaput:
@MaksymilianDemitraszek said:
> I think drop-down menu being also an overview section is very confusing. Can we make overview sections explicit? Screenshot 2022-07-22 at 10 15 56

mkaput:
> @MaksymilianDemitraszek said:
> 
> > I think drop-down menu being also an overview section is very confusing. Can we make overview sections explicit? Screenshot 2022-07-22 at 10 15 56

Docusaurus seems to have such convention to have overview pages. But there's not a problem for me to create separate page (_Fundamentals_) for Testing. More tricky is the layout for cheatcodes, as personally I like how it looks.

MaksymilianDemitraszek:
It is kinda nice, but I would never find the page if i didn't know it has to exist somewhere 😅

PR #509, opened 2022-07-21
Removed --stdout-on-success flag, made printing default behaviour

Radinyn:
I'd like to keep optional printing functionality within the TestCaseResult class, as it is something that may be needed later.

### DIFF
--- a/protostar/commands/test/testing_live_logger.py
+++ b/protostar/commands/test/testing_live_logger.py
@@ -68,7 +68,9 @@ class TestingLiveLogger:
                             else "GREEN"
                         )
 
-                        progress_bar.write(test_case_result.display(include_stdout_section=True))
+                        progress_bar.write(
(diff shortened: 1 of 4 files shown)


protostar/commands/test/testing_live_logger.py
```diff
@@ -71,12 +68,7 @@ def log(
                             else "GREEN"
                         )
 
-                        progress_bar.write(
-                            test_case_result.display(
-                                self.stdout_on_success
-                                or isinstance(test_case_result, FailedTestCase)
-                            )
-                        )
+                        progress_bar.write(test_case_result.display(True))
```
kasperski95, line 71:
```suggestion
                        progress_bar.write(test_case_result.display(include_stdout_section=True))
```
It's easier to read in that way.

kasperski95 (approved):

PR #526, opened 2022-07-25
Isolate state between fuzz test runs

mkaput:


tests/integration/fuzzing/fuzzing_test.py
```diff
@@ -42,3 +42,18 @@ async def test_non_felt_parameter(run_cairo_test_runner: RunCairoTestRunnerFixtu
...
+@pytest.mark.asyncio
+async def test_state_is_isolated(run_cairo_test_runner: RunCairoTestRunnerFixture):
+    testing_summary = await run_cairo_test_runner(
+        Path(__file__).parent / "state_isolation_test.cairo"
+    )
+
+    assert_cairo_test_cases(
+        testing_summary,
+        expected_passed_test_cases_names=[
+            "test_storage_var",
```
kasperski95, line 56: This test makes sense if the function "test_storage_var" is executed more than once. Can you add somehow an assertion that there were at least 2 fuzzing steps?

mkaput: Hmm that's interesting point of view. It would definitely make sense to have something like this, though take into account that fuzzing doesn't make sense if run once. I think we won't really benefit here from such condition.

kasperski95: Yep, fuzzing doesn't make sense when there's one step.

Right now, it's obscure to me how fuzzing can be configured or how many steps are executed by default. It would be nice to explicitly configure fuzzing for the sake of test clarity.

Approving to not block you.

kasperski95 (approved):

PR #534, opened 2022-07-27
Added reporting slowest tests

Radinyn:


protostar/commands/test/slowest_test_formatting.py
```diff
@@ -0,0 +1,35 @@
...
+        row.append(f"{log_color_provider.colorize('GRAY', test_case.file_path)}")
+        row.append(test_case.test_case_name)
+        row.append(f"{test_case.execution_time:.2f} s")
+
+        rows.append(row)
+
+    widths = [max(map(len, col)) for col in zip(*rows)]
+    return "\n".join(
+        "  ".join((val.ljust(width) for val, width in zip(row, widths))) for row in rows
+    )
```
kasperski95, line 35: This could be a part of `TestingSummary` class.

protostar/commands/test/test_command.py
```diff
@@ -188,4 +197,19 @@ async def test(
...
+                    report_slowest_tests_count
+                    and (len(testing_summary.failed) + len(testing_summary.passed)) > 0
+                ):
+                    logger.info(log_color_provider.bold("Slowest test cases:"))
+                    print(
+                        get_formatted_slow_tests(
+                            testing_summary, report_slowest_tests_count
+                        ),
+                        end="\n\n",
+                    )
```
kasperski95, line 211: We need to keep `TestCommand` from growing. If you move the logic to the `TestingSummary` you can pass slowest_tests_count to that class and move this logic there.

tests/e2e/test_testing.py
```diff
@@ -209,3 +209,28 @@ def test_print_only_setup(protostar, copy_fixture):
...
+def test_report_slowest(protostar, copy_fixture):
+    copy_fixture("basic.cairo", "./src")
+    copy_fixture("proxy_contract.cairo", "./src")
+    copy_fixture("test_proxy.cairo", "./tests")
+    copy_fixture("test_failed.cairo", "./tests")
+    copy_fixture("test_print_passed.cairo", "./tests")
+
+    result = protostar(
+        ["test", "tests", "--report-slowest-tests", "999999"], ignore_exit_code=True
+    )
```
kasperski95, line 224: Integration tests checking:
- the order of displaying the slowest tests
- if the amount of the slowest tests actually limits the output
would be nice.

(E2E tests are slow and should be use for testing the core functionality.)

protostar/commands/test/test_command.py
```diff
@@ -136,6 +144,7 @@ async def test(
         safe_collecting: bool = False,
         exit_first: bool = False,
         seed: Optional[int] = None,
+        report_slowest_tests_count: int = 0,
```
kasperski95, line 147: We should generally avoid a first word that could be interpreted as a verb when naming variables:
```suggestion
        slowest_tests_to_report_count: int = 0,
```

kasperski95 (commented):
Screenshot 2022-07-28 at 10 42 51

- I'd move the slowest tests section above [INFO] Test suites. In the console, the mose important information should be at the bottom.
- Green color should be reserved for representing success state (e.g. test pass). I'd change the color of "captured stdout" and color of the enumeration to the "CYAN" and remove the color from the enumeration
- For the enumeration, I'd use format `1.` instead of `[1]` and removed the tab between enumeration and a test case name
- Time should be in the same brackets with steps e.g. `(time=..., steps=...)`
- The slow test list item should have the same or similar look to the test case title (2.31 s) -> gray `(time=2.31s)`

protostar/commands/test/testing_summary.py
```diff
@@ -155,3 +173,39 @@ def _get_preprocessed_core_testing_summary(
...
+
+    def _get_formatted_slow_tests(self, count: int) -> str:
+
+        slowest = self._get_slowest_list(self.failed + self.passed, count)  # type: ignore
+
+        rows: List[List[str]] = []
+        for i, test_case in enumerate(slowest, 1):
+            row: List[str] = []
+            row.append(
+                # pylint: disable=line-too-long
```
Arcticae, line 195: let's just extract color to a variable instead of turning pylint off here

Arcticae (changes requested):
Just a thought - could this be done as a boolean flag with optional parameter of number of slowest tests to log? We could get i.e. `min(tests_ran, 10)` as a default value

kasperski95:
> Just a thought - could this be done as a boolean flag with optional parameter of number of slowest tests to log? We could get i.e. min(tests_ran, 10) as a default value

The same functionality can be achieved with one flag as it is now.

protostar/commands/test/testing_summary.py
```diff
@@ -155,3 +173,42 @@ def _get_preprocessed_core_testing_summary(
...
+        self,
+        failed_and_passed_list: List[Union[PassedTestCase, FailedTestCase]],
+        count: int,
+    ) -> List[Union[PassedTestCase, FailedTestCase]]:
+        lst = sorted(
+            failed_and_passed_list, key=lambda x: x.execution_time, reverse=True
+        )
+        return lst[: min(count, len(lst))]
+
+    def _get_formatted_slow_tests(self, count: int) -> str:
```
kasperski95, line 187: The name suggests returning a list of strings and that this function is a getter.

```suggestion
    def _format_slow_test_cases_list(self, count: int) -> str:
```

tests/integration/testing_timing/testing_timing_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+import pytest
+
+from tests.integration.conftest import (
+    RunCairoTestRunnerFixture,
+    assert_cairo_test_cases,
+)
+
+
+def assert_in_order(logs: str) -> None:
```
kasperski95, line 12:
```suggestion
def assert_execution_times_are_in_desc_order(logs: str) -> None:
```

tests/integration/testing_timing/testing_timing_test.py
```diff
@@ -0,0 +1,62 @@
...
+    assert {1, 2, 3, 4} == get_one_digit_test_indices(test4)
+
+    # Specyfing a number too big should have no effect, because of potential skipped tests.
+    test5 = testing_summary._get_formatted_slow_tests(5)
+    assert_in_order(test5)
+    assert {1, 2, 3, 4} == get_one_digit_test_indices(test5)
+
+    # Zero should yield no result
+    test0 = testing_summary._get_formatted_slow_tests(0)
+    assert test0 == ""
```
kasperski95, line 62: Consider splitting asserts to test cases where `testing_summary` is passed as a fixture.

tests/integration/testing_timing/testing_timing_test.py
```diff
@@ -0,0 +1,62 @@
...
+    assert {1, 2, 3} == indices3
+
+    test4 = testing_summary._get_formatted_slow_tests(4)
+    assert_in_order(test4)
+    assert {1, 2, 3, 4} == get_one_digit_test_indices(test4)
+
+    # Specyfing a number too big should have no effect, because of potential skipped tests.
+    test5 = testing_summary._get_formatted_slow_tests(5)
+    assert_in_order(test5)
+    assert {1, 2, 3, 4} == get_one_digit_test_indices(test5)
```
kasperski95, line 58:
```suggestion
    assert "4. " in test5
    assert "5. " not in test5 
```

protostar/commands/test/testing_summary.py
```diff
@@ -61,6 +61,24 @@ def log(
                 + str(self.testing_seed.value)
             )
 
+    def log_slowest(
```
kasperski95, line 64:
```suggestion
    def log_slowest_test_case(
```
Public methods should be more explicit. Somebody could think that this method logs THE slowest test case.

kasperski95 (changes requested):
Try to be more explicit when naming functions and variables. Quite often I had to analyze the code to figure out what's going on instead of just reading the name.

Requesting changes because the integration test is too hard to read/maintain/fragile.

tests/integration/testing_timing/testing_timing_test.py
Radinyn: There are color issues with that approach + it's possible (though unlikely) that this will interfere with the test timing.
I believe that set comparison is more bullet-proof anyway.

protostar/commands/test/testing_summary.py
```diff
@@ -155,3 +173,42 @@ def _get_preprocessed_core_testing_summary(
...
+                )
+            )
+
+            rows.append(row)
+
+        widths = [max(map(len, col)) for col in zip(*rows)]
+        return "\n".join(
+            "  ".join((val.ljust(width) for val, width in zip(row, widths)))
+            for row in rows
+        )
```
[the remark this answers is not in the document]
Radinyn: Is it necessary to make it a dataclass? Most of the time when we deal with text like this (for example in ```test_cases.py```) we only use strings.

tests/integration/testing_timing/testing_timing_test.py
kasperski95:
```suggestion
    assert {1, 2, 3, 4} == get_first_digit_test_indices(test5)
```
I consider it unnecessarily complicated, but this is just a test. Your call.

tests/integration/testing_timing/testing_timing_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+async def test_timing_equal_to_the_number_of_tests(
+    testing_summary: TestingSummary, no_color_log_color_provider: LogColorProvider
+):
+    test4 = testing_summary._format_slow_test_cases_list(4, no_color_log_color_provider)
+    assert_exec_times_in_desc_order(test4)
+    assert {1, 2, 3, 4} == get_one_digit_test_indices(test4)
+
+
+async def test_timing_more_than_tests(
```
kasperski95, line 61: `test_timing_...` -> `test_output_when_slowest_tests_count...`

protostar/commands/test/testing_summary.py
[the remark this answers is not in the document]
kasperski95: It's not necessary. The end goal is to simplify the code. Maybe from this code a helper function `format_as_table` could be extracted.

I'll not block this PR if you leave this as it is though.

kasperski95 (approved):

protostar/commands/test/testing_summary.py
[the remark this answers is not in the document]
kasperski95: It's not necessary. The end goal is to simplify the code. Maybe from this code a helper function `format_as_table` could be extracted.

I'll not block this PR if you leave this as it is though.

tests/e2e/test_testing.py
```diff
@@ -209,3 +209,28 @@ def test_print_only_setup(protostar, copy_fixture):
...
+
+    assert "Slowest test cases" in result
```
mkaput, line 226: you are not testing that the tests themselves have been printed out

protostar/commands/test/test_command.py
```diff
@@ -108,6 +108,12 @@ def arguments(self) -> List[Command.Argument]:
                 type="int",
                 description="Set a seed to use for all fuzz tests.",
             ),
+            Command.Argument(
+                name="report-slowest-tests",
+                type="int",
+                description="Print slowest tests at the end.",
+                default=0,
+            ),
```
mkaput, line 116: Why is this a configurable argument? IMHO people will not use this feature not even know that it exists, because it's disabled by default

tests/integration/testing_timing/testing_timing_test.py
```diff
@@ -0,0 +1,67 @@
...
+from protostar.utils.log_color_provider import LogColorProvider
+from tests.integration.conftest import RunCairoTestRunnerFixture
+
+
+def assert_exec_times_in_desc_order(logs: str) -> None:
+    raw_exec_times = [row.split()[3] for row in logs.split("\n")]
+    exec_times = [
+        float(exec_time.lstrip("(time=").rstrip("s)")) for exec_time in raw_exec_times
+    ]  # "(time={NUMBER}s)" -> {NUMBER}
+    assert exec_times == sorted(exec_times, reverse=True)
```
mkaput, line 15: I think this assertion may be faulty if there are two tests with different names but exactly the same timings. This assertion may not check that tests are sorted by name as expected.

Why not simply asserting that the logged output matches a regular expression?

PR #540, opened 2022-07-28
Max fuzz examples config

kasperski95:
The hypothesis exceeds the threshold when the test fails. Max examples parameter doesn't affect the simplification process: 

> If it finds an example where it doesn’t, it takes that example and cuts it down to size, simplifying it until it finds a much smaller example that still causes the problem

Screenshot 2022-07-29 at 13 26 55

**Question:**
- [x] Should `max_fuzz_examples` be renamed to `fuzz_examples`? Is there a case in which Hypothesis runs fewer examples?


kasperski95:
@MaksymilianDemitraszek The purpose of this flag is controlling fuzzing quality/performance tradeoff from the CLI. `max_examples` is the Hypothesis configuration parameter. @Radinyn is working on `assume` cheatcode.

protostar/commands/test/starkware/test_execution_state.py
```diff
@@ -6,24 +6,29 @@
...
 from protostar.starknet.forkable_starknet import ForkableStarknet
 from protostar.utils.starknet_compilation import StarknetCompiler
-from protostar.commands.test.test_output_recorder import OutputRecorder
 
 
 @dataclass
 class TestExecutionState(ExecutionState):
+    @dataclass
+    class Config:
+        max_fuzz_examples: int
```
MaksymilianDemitraszek, line 19: It is a value only related to `FuzzTestExecutionEnvironment` we shouldn't keep it in `TestExecutionState` config

MaksymilianDemitraszek: We can make a config that inherits from this config which includes max_fuzz_examples

protostar/commands/test/environments/test_execution_environment.py
```diff
@@ -22,22 +23,23 @@
...
-from protostar.commands.test.cheatcodes.reflect.cairo_struct import (
-    CairoStructHintLocal,
-)
 
 
-class TestExecutionEnvironment(
-    ExecutionEnvironment[Optional[ExecutionResourcesSummary]]
-):
+@dataclass
+class TestExecutionResult:
```
MaksymilianDemitraszek, line 29: Same as for config, we should introduce proper hierarchy here

MaksymilianDemitraszek (changes requested):

protostar/commands/test/test_cases.py
```diff
@@ -29,6 +29,7 @@ class PassedTestCase(TestCaseResult):
     test_case_name: str
     execution_resources: Optional[ExecutionResourcesSummary]
     captured_stdout: Dict[OutputName, str] = field(default_factory=dict)
+    fuzz_runs_count: int = 1
```
mkaput, line 33: I wonder if it wouldn't be more correct to make this

```suggestion
    fuzz_runs_count: Optional[int] = None
```

and keep it `None` for non-fuzz tests. The reason is that with current model, we lost an ability to track 1 fuzz test execution -- which although is not desirable, is totally possible if the test fails in second run.

protostar/commands/test/test_environment_exceptions.py
```diff
@@ -29,6 +29,7 @@ class ReportedException(BaseException):
 
     def __init__(self, *args: object) -> None:
         self.metadata: List[ExceptionMetadata] = []
+        self.execution_info: Dict[str, Union[int, str]] = {}
```
mkaput, line 32: Why not reusing `metadata` field here? We could add metadata objects like `class FuzzRunsMetadata(ExceptionMetadata, ExecutionInfo)` and you could write `isinstance(..., ExecutionInfo)` in the loop in `FailingTestCase`.

BTW I'm starting to dislike the difference of handling succeeding and failing test and the duplication of logic which is a result of it...

mkaput (changes requested):
I can accept this to be merged for the release but please address these two issues in subsequent PR as soon as possible.

protostar/commands/test/test_environment_exceptions.py
kasperski95: I created a new field because to print `fuzz_runs` next to `steps` and other execution properties etc. Statistics are easier to interpret when we know the number of observations.

> BTW I'm starting to dislike the difference of handling succeeding and failing test and the duplication of logic which is a result of it...

I don't like it, too. I created the task for that, and I plan to refactor after the release. https://github.com/software-mansion/protostar/issues/544

mkaput: > I created a new field because to print `fuzz_runs` next to `steps` and other execution properties etc. Statistics are easier to interpret when we know the number of observations.

Yeah, this part I got. The thing I am not fully liking is the fact that there is a new field `execution_info` added to the exception, instead of reusing the `metadata` field which is already existing.

Take into account that exception metadata must be pickleable and this property is probably easier to manage if all metadata-like things are stored in uniform way.

protostar/commands/test/test_command.py
```diff
@@ -108,6 +108,15 @@ def arguments(self) -> List[Command.Argument]:
                 type="int",
                 description="Set a seed to use for all fuzz tests.",
             ),
+            Command.Argument(
+                name="max-fuzz-examples",
```
mkaput, line 112: Wouldn't it be more consistent if all fuzzing-related options had single prefix `--fuzz-*` ?

mkaput (changes requested):

protostar/commands/test/test_environment_exceptions.py
kasperski95: Ok, I'll change to use `metadata` property and use `instanceof` in `FailedTestCase` to determine how to print metadata.

protostar/commands/test/environments/fuzz_test_execution_environment.py
```diff
@@ -43,10 +44,24 @@ def is_fuzz_test(function_name: str, state: TestExecutionState) -> bool:
...
 
+@dataclass
+class FuzzTestExecutionResult(TestExecutionResult):
+    fuzz_runs_count: int
+
+
 class FuzzTestExecutionEnvironment(TestExecutionEnvironment):
+    @dataclass
+    class FuzzConfig:
+        max_fuzz_examples: int = 100
```
mkaput, line 55: could this class be moved out to the top of this module, so that the stance `FuzzTestExecutionEnvironment.FuzzConfig` could be simplified to just `FuzzConfig`? without repetitive `Fuzz` part

protostar/commands/test/test_scheduler.py
```diff
@@ -26,7 +26,7 @@ def __init__(
     def run(
         self,
         test_collector_result: "TestCollector.Result",
-        config: TestRunner.Config,
+        config: TestRunner.FuzzConfig,
```
mkaput, line 29:
```suggestion
        fuzz_config: TestRunner.FuzzConfig,
```

protostar/commands/test/test_runner.py
```diff
@@ -162,7 +171,9 @@ async def _run_test_suite(
                         execution_resources=execution_result.execution_resources,
                         execution_time=time.perf_counter() - start_time,
                         captured_stdout=new_execution_state.output_recorder.get_captures(),
-                        fuzz_runs_count=execution_result.fuzz_runs_count,
+                        fuzz_runs_count=execution_result.fuzz_runs_count
+                        if isinstance(execution_result, FuzzTestExecutionResult)
+                        else None,
```
mkaput, line 176: strange indentation, may be worth extracting a variable here.

protostar/commands/test/environments/fuzz_test_execution_environment.py
```diff
@@ -43,10 +44,24 @@ def is_fuzz_test(function_name: str, state: TestExecutionState) -> bool:
...
+
     def __init__(self, state: TestExecutionState):
         super().__init__(state)
         self.initial_state = state
+        self._fuzz_config = FuzzTestExecutionEnvironment.FuzzConfig()
+
+    def set_fuzz_config(self, fuzz_config: FuzzConfig) -> Self:
+        self._fuzz_config = fuzz_config
+        return self
```
mkaput, line 64: could the value of this property be passed via constructor? even at the cost of adding extra keyword arguments here and there. I don't like exposing a mutable-like semantics here.

mkaput (changes requested):

protostar/commands/test/test_environment_exceptions.py
kasperski95: Discussed privately. It will be refactored in https://github.com/software-mansion/protostar/issues/544.

MaksymilianDemitraszek (approved):

protostar/commands/test/environments/fuzz_test_execution_environment.py
kasperski95: It could be passed through constructor however:
- it breaks the factory (TestExecEnv i FuzzTestExecEnv) would accept different argument and the factory
- I try to avoid passing runtime configuration through a constructor because the class can't be instantiated before configuration is known therefore the class can't be injected as a dependency. This forces a developer using Builders as a workaround. In the constructor I tend to pass only dependencies, and I am not sure if the configuration is a dependency. Anyway, in this case it doesn't matter — I'll change it.

PR #546, opened 2022-07-29
Add more feedback for commands

Arcticae:


protostar/cli/command.py
```diff
@@ -45,6 +45,12 @@ def directory(arg: str) -> Path:
         def copy_with(self, **changes) -> "Command.Argument":
             return replace(self, **changes)
 
+    @dataclass(frozen=True)
+    class PrintedOutputs:
+        entry: str
+        exit_success: str
+        exit_error: str
+
```
mkaput, line 53: I don't fully understand the reasoning behind having this class and the `outputs` field... Wouldn't just inlining these messages in `self._logger.***(...)` calls just work?

mkaput (changes requested):

mkaput (approved):

PR #550, opened 2022-07-29
Add section about `fuzz-max-examples`

kasperski95:
- [x] merge after https://github.com/software-mansion/protostar/pull/540


website/docs/tutorials/07-testing/03-fuzzing/README.md
```diff
@@ -179,6 +179,12 @@ amount = 340282366920938463463374607431768211456
 11:41:48 [INFO] Seed:        4258368192
 ```
 
+:::info
+By default, Protostar tries to fail a test case within 100 examples. The more complex code, the more examples are needed to find uncommon bugs. Use `--max-fuzz-examples` parameter to change the number of examples.
```
MaksymilianDemitraszek, line 183: I am not sure about word `example` but I have nothing better in my mind:
- input examples
- approaches
- inputs

MaksymilianDemitraszek (approved):

website/docs/tutorials/07-testing/03-fuzzing/README.md
kasperski95: It's a hypothesis term though.

mkaput, line 183: please put a link to CLI reference somewhere here

mkaput (changes requested):
I don't think having this just as a suggestion box in a tutorial will help discoverability of this configuration argument -- people won't learn that they can adjust this.

I would rather make this a separate section after `interpreting results` part.

mkaput (approved):

PR #551, opened 2022-08-01
Assume and reject cheatcodes

Radinyn:


website/docs/tutorials/07-testing/02-cheatcodes/assume_and_reject.md
```diff
@@ -0,0 +1,49 @@
+# `assume and reject`
```
mkaput, line 1: we have an unspoken rule one cheatcode == one page here

website/docs/tutorials/07-testing/02-cheatcodes/assume_and_reject.md
```diff
@@ -0,0 +1,49 @@
...
+Those two cheatcodes can be used **only** in fuzz tests.
```
mkaput, line 2:
```suggestion
:::warning
This cheatcode is only available in [fuzz tests](put a link to fuzz testing docs here).
:::
```

tests/e2e/test_fuzzing.py
```diff
@@ -41,3 +41,21 @@ def test_fuzzing(protostar, copy_fixture):
...
+    )
+
+    assert (
+        """
+[falsifying example]:
+a = 6
+b = 3
+""".rstrip()
+        in result
+    )
```
mkaput, line 61: IMHO using an integration test for this is enough here. E2E tests are slow, no need to add more of them unnecessarily

protostar/commands/test/fuzzing/exceptions.py
```diff
@@ -1,5 +1,13 @@
...
 
 class FuzzingError(ProtostarException):
     pass
+
+
+class HypothesisRejectException(ReportedException):
+    """
+    Hypothesis uses exceptions to handle `hypothesis.reject()` and `hypothesis.assume()`.
+    This class is used to smuggle them through the Cairo VM.
+    """
```
mkaput, line 13: Hypothesis may end up using `UnsatisfiedAssumption` to smuggle some information somewhen, so we should play it safe, and not reimplement their logic for rejections ourselves.

I think we should:
1. Store catched `UnsatisfiedAssumption` inside `HypothesisRejectException`
2. `throw HypothesisRejectException(unsatisifed_assumption_ex) from unsatisifed_assumption_ex` instead of `from None`
3. re-raise catched `unsatisifed_assumption_ex` in fuzzing loop

mkaput (changes requested):

protostar/commands/test/environments/fuzz_test_execution_environment.py
```diff
@@ -258,3 +275,24 @@ class RunsCounter:
...
+
+
+class FuzzTestCaseCheatcodeFactory(TestCaseCheatcodeFactory):
+    def __init__(
+        self,
+        state: TestExecutionState,
+        expect_revert_context: ExpectRevertContext,
+        finish_hook: Hook,
+    ):
+        super().__init__(state, expect_revert_context, finish_hook)
```
kasperski95, line 287: You could pass `TestCaseCheatcodyFactory` in the constructor. By doing so, we don't have to update `FuzzTestCaseCheatcodeFactory` everytime `TestCaseCheatcodeFactory` changes.

This approach breaks the following rules:
- Open/Closed principle
- composition over inheritance
- DRY

website/docs/tutorials/07-testing/02-cheatcodes/reject.md
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    # ...
+
+    return ()
+end
+```
+:::warning
+This cheatcode is only available in [fuzz tests](../fuzzing).
+:::
```
mkaput, line 26: it's definitely more important fragment than the example and the statement that reject is equivalent to assume. I would put this box after the `Marks example as bad` statement

mkaput (changes requested):

website/docs/tutorials/07-testing/02-cheatcodes/assume.md
```diff
@@ -0,0 +1,25 @@
+# `assume`
+```python
+def assume(condition: bool) -> None:
+```
+`assume(condition)` works similarly to `assert condition` but instead of marking the example as failed it marks it as bad, thus preventing fuzz tests from using it as a falsifying example.
+
```
kasperski95, line 6: - **it** marks **it**
- it looks similar to hypothesis docs however without context this description is hard to interpret
-  works similarly to `assert condition` — in Cairo asserts are used a bit differently

```suggestion
`assume(condition)` skips testing against given fuzzing arguments if the `condition` is `False`.
```

website/docs/tutorials/07-testing/02-cheatcodes/reject.md
```diff
@@ -0,0 +1,30 @@
+# `reject`
+```python
+def reject() -> None:
+```
+Marks example as bad, thus preventing fuzz tests from using it as a falsifying example.
```
kasperski95, line 5: `bad` and `example` are unknown terms in the Protostar's context

protostar/starknet/cheatable_cairo_vm.py
```diff
@@ -1,58 +1,7 @@
...
-
-        # Decode.
-        instruction = self.decode_current_instruction()
-
-        # Run.
-        self.run_instruction(instruction)
+# pylint: disable=pointless-string-statement
+"""
+CairoVM asserts in `vm_exceptions.py` that all exceptions come from a file named `virtual_machine_base.py`.
+"""
```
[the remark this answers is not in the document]
mkaput: discussed privately

protostar/starknet/cheatable_cairo_vm.py
```diff
@@ -4,7 +4,7 @@
 
 class CheatableVirtualMachine(VirtualMachine):
     """
-    VirtualMachine with modified step function that captures execution locals of a hint.
+    `VirtualMachine` with modified `step` function that builds cheatcodes created with `DelayedBuilder`.
```
mkaput, line 7: Could we please avoid touching this file entirely in this PR? We don't change any logic here, so it would be great to keep this PR separated from this file.

mkaput (changes requested):

protostar/commands/test/environments/fuzz_test_execution_environment.py
kasperski95: This is also a useless constructor. You can remove it to temporarily fix the DRY problem.

Refactor can be done in a different PR.

mkaput (approved):

PR #564, opened 2022-08-02
Simplify cheatcodes imports

mkaput:


kasperski95 (approved):
I always wanted to do it in such way, but I've always been told to use absolute paths without a reasonable to me explanation.

mkaput:
@MaksymilianDemitraszek 👀

MaksymilianDemitraszek:
@kasperski95 I don't think what you say is true, if I'm wrong please link relevant discussion because I don't rememeber such thing.

Only discussion I can remeber about the subject is this comment:
https://github.com/software-mansion/protostar/pull/224#discussion_r859898453
It is about something completely different (explicit imports).

MaksymilianDemitraszek (approved):

kasperski95:
I don't have @MaksymilianDemitraszek specifically on my mind.

- https://github.com/software-mansion/protostar/pull/8#pullrequestreview-874381836
- https://github.com/software-mansion/protostar/pull/11#discussion_r799617222

And a few more times somewhere, but I don't want to waste more time on searching that. We also had internal discussion or discussions during our calls. I talk about it with @Solpatium privately. It was 3 against 1 so I gave up.

PR #565, opened 2022-08-02
Deploying by a contract name [1/2] — refactor feature dependencies

kasperski95:
- refactored `ProjectCompiler`
 - moved `project_compiler.py` out of `commands/build` 
 - extracted `ProjectCairoPathBuilder`
 - extracted `ContractWriter`
 - changed `ProjectCompiler` methods to be smaller and more specific


protostar/protostar_toml/protostar_contracts_section.py
```diff
@@ -50,3 +51,22 @@ def to_dict(self) -> "ProtostarTOMLSection.ParsedProtostarTOML":
...
+
+    def get_contract_names(self) -> List[str]:
+        return list(self.contract_name_to_paths.keys())
+
+    def get_relative_contract_source_paths(self, contract_name: str) -> List[Path]:
+        self.assert_contract_is_defined(contract_name)
+        source_paths = self.contract_name_to_paths[contract_name]
+        return source_paths
+
+    def assert_contract_is_defined(self, contract_name: str):
```
mkaput, line 63:
```suggestion
    def check_contract_is_defined(self, contract_name: str):
```

assert suggest it will raise `AssertionError` and can be turned off

protostar/compiler/compiled_contract_writer.py
```diff
@@ -0,0 +1,45 @@
+import json
+from pathlib import Path
+
+from starkware.starknet.services.api.contract_class import ContractClass
+
+
+class CompiledContractWriter:
```
mkaput, line 7: this could be a plain function

protostar/compiler/compiled_contract_writer.py
```diff
@@ -0,0 +1,45 @@
...
+            mode="w",
+            encoding="utf-8",
+        ) as output_abi_file:
+            json.dump(self._contract.abi, output_abi_file, indent=4, sort_keys=True)
+            output_abi_file.write("\n")
+
+    @staticmethod
+    def _create_output_dir(output_dir: Path):
+        if not output_dir.exists():
+            output_dir.mkdir(parents=True)
```
mkaput, line 45:
```suggestion
        output_dir.mkdir(parents=True, exist_ok=True)
```

and you could inline this function

protostar/compiler/compiled_contract_writer.py
```diff
@@ -0,0 +1,45 @@
...
+        with open(
+            Path(output_dir, f"{self._contract_name}.json"), mode="w", encoding="utf-8"
+        ) as output_file:
+            json.dump(
+                self._contract.Schema().dump(self._contract),
+                output_file,
+                indent=4,
+                sort_keys=True,
+            )
+            output_file.write("\n")
```
mkaput, line 27: the pattern of opening a file X and writing Y into it as pretty formatted JSON with trailing whitespace is duplicated here

protostar/compiler/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .project_compiler import (
+    CompilationException,
+    SourceFileNotFoundException,
+    ProjectCairoPathBuilder,
+    ProjectCompiler,
+)
```
mkaput, line 6: Do we need this? It's not a request, but personally I tend to skip writing collective `__init__` modules in non-library Python projects 😃

protostar/commands/build/build_command.py
```diff
@@ -53,11 +52,13 @@ async def run(self, args):
             log_color_provider.colorize("GRAY", "Building projects' contracts")
         ):
             try:
-                self._project_compiler.compile(
-                    output_dir=args.output,
-                    relative_cairo_path=args.cairo_path,
-                    disable_hint_validation=args.disable_hint_validation,
+                self._project_compiler.set_config(
```
mkaput, line 55: Could this mutability be avoided? It suggests that the config can be changed during compilation, which I guess would result in crashes.

protostar/compiler/project_compiler.py
```diff
@@ -0,0 +1,136 @@
...
+ContractSourcePath = Path
+ContractIdentifier = Union[ContractName, ContractSourcePath]
+
+
+class ProjectCompiler:
+    @dataclass
+    class Config:
+        relative_cairo_path: List[Path]
+        debugging_info_attached: bool = False
+        hint_validation_disabled: bool = False
```
mkaput, line 32: Could this be renamed to `ProjectCompilerConfig` and put outside the `ProjectCompiler` class? I find it unergonomic to use nested classes because they are hard to find when grepping symbols in the source code/autocompletion.

mkaput (changes requested):

protostar/compiler/__init__.py
kasperski95: I like to use this approach from the following reasons:
- changes in the module structure don't propagate across the codebase
- making imports nicer
```python
from protostar.compiler import CompilationException, ProjectCompiler
```
- defining module's interface

protostar/commands/build/build_command.py
kasperski95: I'll make it a parameter of `compile*` functions instead.

protostar/compiler/project_compiler.py
kasperski95: I tend to use nested classes because they don't have to be imported. I can put it outside.

protostar/compiler/compiled_contract_writer.py
kasperski95: In this case, it could. Something feels off when mixing paradigms.

It's easier to find a file that contains a class rather than a list of functions. Naming functions in the same way as files is confusing because a function and a filename use the same casing.

protostar/compiler/project_compiler.py
mkaput: do you manage imports manually when writing code? because I'm used to have good auto importer running in my code editor 😅

protostar/compiler/__init__.py
mkaput: 👍🏻

protostar/compiler/compiled_contract_writer.py
mkaput: You can still define single public module function, and name the module the same way. That's what I have seen in the wild.

> Something feels off when mixing paradigms.

To me, using classes too much in Python/JS is mixing paradigms.

mkaput (approved):

protostar/compiler/compiled_contract_writer.py
kasperski95: > You can still define single public module function, and name the module the same way. That's what I have seen in the wild.

I did it when I used the functional paradigm. Auto import functionality confused the function with the module, if they were named in the same way. Also, I couldn't patch the function, if it was imported from the `__init__.py` (barrel file) from the same reason.

> To me, using classes too much in Python/JS is mixing paradigms.

Functional programming is more common in Python, but it doesn't mean that this is the proper tool for the job.

> -  Object-Oriented Programming is the tool best suited for defining how we cross architectural boundaries with polymorphism and plugins
> - Functional programming is the tool we use to push data to the boundaries of our applications
> - and Structured programming is the tool we use to write algorithms

https://khalilstemmler.com/articles/software-design-architecture/full-stack-software-design/

PR #575, opened 2022-08-03
Validate `protostar_version` from the toml file 

Arcticae:


Arcticae:
@mkaput can you take a look? seems like fuzzing is failing here

protostar/utils/protostar_directory.py
```diff
@@ -58,30 +63,33 @@ def protostar_version(self) -> Optional[VersionType]:
...
+        return VersionManager.parse(version_s)
+
+    @property
+    def toml_breaking_versions(self) -> List[VersionType]:
+        return [
+            VersionManager.parse(v)
+            for v in self.pyproject_toml["tool"]["protostar"]["toml_breaking_versions"]
+        ]
+
+    def version_range_has_breaking_changes(self, low: VersionType, high: VersionType):
```
kasperski95, line 88: Methods and functions should start with a noun.

kasperski95 (changes requested):

protostar/utils/protostar_directory.py
```diff
@@ -58,30 +63,33 @@ def protostar_version(self) -> Optional[VersionType]:
...
             with open(path, "r", encoding="UTF-8") as file:
-                version_s = tomli.loads(file.read())["tool"]["poetry"]["version"]
-                return VersionManager.parse(version_s)
-        except FileNotFoundError:
-            getLogger().warning("Couldn't read Protostar version")
-            return None
+                self._pyproject_toml_dict = tomli.loads(file.read())
+                return self._pyproject_toml_dict
+        except FileNotFoundError as err:
+            raise ProtostarException("Couldn't read Protostar version") from err
```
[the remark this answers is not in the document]
Arcticae: This is not a problem since working directory will contain `pyproject.toml` anyway

protostar/utils/protostar_directory.py
Arcticae: this does start with a noun

MaksymilianDemitraszek: He probably meant verb.

Renaming it to `has_version_range_breaking_changes` won't hurt imho

MaksymilianDemitraszek: Also please add `-> bool` here

protostar/utils/protostar_directory.py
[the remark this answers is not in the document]
kasperski95: It is if you want for example to test protostar in the "access" project. To do so, you have to navigate to that project (I usually keep them in the playground directory that is git ignored), and that triggers the error.

protostar/protostar_cli.py
```diff
@@ -296,3 +297,40 @@ async def run(self, args: Any) -> None:
 
         if has_failed:
             sys.exit(1)
+
+    def _check_protostar_toml_compatibility(self, args: Any):
+        if args.command in ["init", "upgrade"]:
+            return
```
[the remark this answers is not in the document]
kasperski95: `ProtostarTOMLReader` is used before running a command. See `start.py`.

mkaput:
> mkaput

#582. In the meantime, feel free to restart the job. The flaky test is not flaky enough, it seems; but still this is a randomized test so this won't happen on each run.

protostar/utils/protostar_directory.py
```diff
@@ -58,30 +63,33 @@ def protostar_version(self) -> Optional[VersionType]:
...
+    def cairo_version(self) -> VersionType:
+        version_s = self.pyproject_toml["tool"]["poetry"]["dependencies"]["cairo-lang"]
+        return VersionManager.parse(version_s)
+
+    @property
+    def toml_breaking_versions(self) -> List[VersionType]:
+        return [
+            VersionManager.parse(v)
+            for v in self.pyproject_toml["tool"]["protostar"]["toml_breaking_versions"]
+        ]
```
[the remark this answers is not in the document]
kasperski95: TBH, I don't see how it makes adding migration scripts easier. The list will grow and grow.

protostar/protostar_cli.py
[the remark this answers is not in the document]
Arcticae: I'm going to create a function for this

protostar/utils/protostar_directory.py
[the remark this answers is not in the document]
Arcticae: i think `python ./path/to/your/protostar/binary_entrypoint.py test` will work just fine, won't it?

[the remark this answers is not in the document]
Arcticae: We probably should have a solution for this (something like `yarn --cwd`, or possibly a flag like --protostar-toml="~/protostar.toml")

protostar/protostar_toml/protostar_toml_version_checker.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(
+        self,
+        protostar_toml_reader: ProtostarTOMLReader,
+        version_manager: VersionManager,
+    ):
+        self._protostar_toml_reader = protostar_toml_reader
+        self._version_manager = version_manager
+
+    def run(self, args: Any):
```
MaksymilianDemitraszek, line 19:
```suggestion
    def run(self, command: str):
```

protostar/protostar_toml/protostar_toml_version_checker.py
```diff
@@ -0,0 +1,48 @@
...
+        if args.command in self.SKIP_FOR:
+            return
+
+        declared_version_str = self._protostar_toml_reader.get_attribute(
+            "config", "protostar_version"
+        )
+
+        assert (
```
MaksymilianDemitraszek, line 27: Can we use ProtostarException instead of assert? 

I know it is not supposed to happen but it is kinda user input validation, I think oldest version of protostar don't have this field

protostar/composition_root.py
```diff
@@ -158,10 +161,14 @@ def build_di_container(script_root: Path):
...
         ),
     ]
+    toml_version_checker = ProtostarTOMLVersionChecker(
+        protostar_toml_reader=protostar_toml_reader, version_manager=version_manager
+    )
 
     protostar_cli = ProtostarCLI(
         commands=commands,
         latest_version_checker=latest_version_checker,
+        toml_version_checker=toml_version_checker,
```
kasperski95, line 171: This name is confusing. It suggests checking the version of any `toml`. I'd change all occurrences of `toml_version_checker`.

```suggestion
        protostar_toml_version_checker=protostar_toml_version_checker,
```

protostar/protostar_toml/protostar_toml_version_checker.py
```diff
@@ -0,0 +1,48 @@
...
+        self,
+        protostar_toml_reader: ProtostarTOMLReader,
+        version_manager: VersionManager,
+    ):
+        self._protostar_toml_reader = protostar_toml_reader
+        self._version_manager = version_manager
+
+    def run(self, args: Any):
+        if args.command in self.SKIP_FOR:
+            return
```
kasperski95, line 21: `ProtostarTOMLVersionChecker` should be uncoupled from IO (CLI in this case).
I'd move this command to `ProtostarCLI`.

kasperski95: You can modify `ProtostarCLI` to accept commands explicitly, e.g.

```python

def __init__(self, init_command: InitCommand, ...)
```
(This still breaks the open-close principle, however `ProtostarCLI` is more related to commands than `ProtostarTOMLVersionChecker`. To do it properly, the check should be in the command code. It would require a bit more code.)


Then you can constraint command name
```python
        if args.command in [self._init_command.name, self._upgrade_command.name]:
            # run checker
```

protostar/protostar_toml/protostar_toml_version_checker.py
kasperski95: This is going to happen when we remove the config section. The error here should tell the user to upgrade the Protostar.

protostar/protostar_toml/protostar_toml_version_checker.py
```diff
@@ -0,0 +1,48 @@
...
+        declared_version_str = self._protostar_toml_reader.get_attribute(
+            "config", "protostar_version"
+        )
+
+        assert (
+            declared_version_str
+        ), "No protostar_version attribute available in protostar.toml"
+
+        # Version from protostar.toml
+        declared_version = VersionManager.parse(declared_version_str)
```
kasperski95, line 32:
```suggestion
        version_from_protostar_toml = VersionManager.parse(declared_version_str)
```

protostar/protostar_toml/protostar_toml_version_checker.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # Version from pyproject.toml
+        last_supported_protostar_toml_version = (
```
kasperski95, line 35: Comments are hard to maintain and easily can become outdated.

```suggestion
        last_supported_protostar_toml_version = (
```

protostar/protostar_toml/protostar_toml_version_checker_test.py
```diff
@@ -0,0 +1,74 @@
...
+        ProtostarTOMLVersionChecker(protostar_toml_reader, version_manager).run(args)
+    assert "is not compatible with provided protostar.toml" in exception.value.message
+
+
+@pytest.mark.parametrize(
+    "last_supported_protostar_toml_version, protostar_toml_version",
+    (
+        ("1.0", "1.1"),
+        ("2.1", "2.2"),
+        ("1.0.0", "1.0.1"),
```
kasperski95, line 42: It's worth to test edges, e.g. `("1.0", "1.0")`

kasperski95 (changes requested):
The business logic looks good.

protostar/protostar_toml/protostar_toml_version_checker.py
kasperski95: Business logic should be independent of the IO code.

protostar/protostar_toml/protostar_toml_version_checker.py
```diff
@@ -0,0 +1,47 @@
...
+        self._version_manager = version_manager
+
+    def run(self):
+        declared_version_str = self._protostar_toml_reader.get_attribute(
+            "config", "protostar_version"
+        )
+
+        if not declared_version_str:
+            raise ProtostarException(
+                "No protostar_version attribute available in protostar.toml.\n"
```
kasperski95, line 28: `protostar_version` can exist in the `protostar.toml` but Protostar doesn't know how to load it

```suggestion
                "Couldn't load `protostar_version`.\n"
                "`protostar.toml` could've been created by a newer version of Protostar".
```

protostar/protostar_cli.py
```diff
@@ -82,6 +89,10 @@ async def _run_command_from_args(self, args: Any) -> None:
         if args.version:
             self._version_manager.print_current_version()
             return
+
+        if args.command not in [InitCommand.name, UpgradeCommand.name]:
```
kasperski95, line 93: It works, but I don't understand why (`self` is passed to the name property).

kasperski95 (approved):

PR #576, opened 2022-08-04
Composition root

kasperski95:
- moved instantiation from `protostar_cli.py` to `composition_root.py`
- refactored:
  - `start.py`
  - `start_test.py`
  - `protostar_cli.py` 
- moved cli_reference_docs snapshot test to `generate_reference_docs_test.py` and update `poe test_integration`

### DIFF
--- a/protostar/protostar_cli_test.py
+++ b/protostar/protostar_cli_test.py
@@ -57,7 +57,7 @@ async def test_should_fail_due_to_old_git(
     protostar_cli: ProtostarCLI, mocker: MockerFixture
 ):
     logger_mock = mocker.MagicMock(Logger)
-    protostar_cli.logger = logger_mock
+    protostar_cli._logger = logger_mock
     logger_mock.error = mocker.MagicMock()
     # pylint: disable=protected-access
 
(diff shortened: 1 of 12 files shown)


Arcticae (approved):

protostar/start.py
```diff
@@ -1,37 +1,56 @@
...
 
 def main(script_root: Path):
-    protostar_cli = ProtostarCLI.create(script_root)
+    container = build_di_container(script_root)
+    arg_parser = _build_parser(container.protostar_cli, container.protostar_toml_reader)
+    args = _parse_args(arg_parser)
+    _run_protostar(container.protostar_cli, args, arg_parser)
 
+
+def _build_parser(
```
MaksymilianDemitraszek, line 26: I never seen `_` prefix for functions only for class methods. Can we don't use it?

MaksymilianDemitraszek (approved):

protostar/protostar_cli_test.py
```diff
@@ -28,94 +30,92 @@ def version_manager_fixture(mocker: MockerFixture, git_version: str):
     return version_manager
 
 
+@pytest.fixture(name="logger")
+def logger_fixture(mocker: MockerFixture):
+    return mocker.MagicMock()
```
mkaput, line 35: Do we want to hide logs from tests? I have personally used Protostar output a lot while debugging integration tests

mkaput (approved):

protostar/protostar_cli_test.py
MaksymilianDemitraszek: +1

PR #577, opened 2022-08-04
Add a test for unbounded integers strategy

mkaput:


tests/integration/fuzzing/strategies_test.cairo
```diff
@@ -14,6 +14,18 @@ func test_integers{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt):
     return ()
 end
 
+@external
+func test_integers_unbounded{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt):
+    %{
+        given(
+            a = strategy.integers(max_value=2**15),
+            b = strategy.integers(min_value=2**15),
```
kasperski95, line 22: Can you move `2**15` to a variable so that it's not a magic number? Shouldn't `PRIME` be used?

```python
        PRIME = 2**251 + 17 * 2**192 + 1
```

kasperski95 (approved):

tests/integration/fuzzing/strategies_test.cairo
mkaput: No, PRIME is the worst choice here actually 😄 I'll introduce a variable here and document the requirements.

tests/integration/fuzzing/fuzzing_test.py
```diff
@@ -98,7 +98,7 @@ async def test_max_fuzz_runs_less_or_equal_than_specified(
 async def test_strategies(
     run_cairo_test_runner: RunCairoTestRunnerFixture,
 ):
-    fuzz_max_examples = 10
+    fuzz_max_examples = 100
```
kasperski95, line 101: It makes fuzz tests slower though. (@Radinyn complained about that.)

kasperski95 (approved):

tests/integration/fuzzing/fuzzing_test.py
mkaput: Yes, but this is needed for correctness. With `10`, fuzzer wasn't able to explore large numbers :(

mkaput: I future PR I till chop this test into many ones, which only the tests that need such high max examples value will have one.

PR #589, opened 2022-08-05
Add cwd as default target for testing

Arcticae:


mkaput:
I actually start to thinking that we should have more sophisticated filtering scheme, like only test files matching `(^test_.*\.cairo)|(.*_test\.cairo)$` or something like this.

kasperski95 (approved):
LGTM

> I wonder if it wouldn't be better for discoverability to just remove default value and make this arg required. Not a change request or something just a thought

`jest` and `pytest` works in this way and CLI should be convenient to use.


> I actually start to thinking that we should have more sophisticated filtering scheme, like only test files matching (^test_.*\.cairo)|(.*_test\.cairo)$ or something like this.

`TestCollector` uses hard-coded filtering schemes

mkaput (approved):

Issue #590, opened 2022-08-05
[BUG] Fuzzing fails because of python's 'ulimit' on macos 12.2.1


JorikSchellekens:
## Protostar version

| Name       | Version              |
| ---------- | -------------------- |
| Protostar  | 0.3.0 |
| Cairo-lang | ^0.9.1 |
| Git        | 2.37.1       |

## Current Behavior

Fuzzing a test which doesn't satisfy `expect_revert` for the following condition causes too many files to be opened.

## Steps to Reproduce

1. run a fuzzer test on the following:
```
@external
func test_safe_cast_fuzz {
    range_check_ptr, bitwise_ptr: BitwiseBuiltin*
  } (val: felt):

  %{
      if (ids.val % 2 == 0 and ids.val > (2**125 - 1) * 2):
          expect_revert()
  %}

  return ()
end
```
## Possible Solution

I checked my laptop's ulimit which is set to unlimited so doesn't seem to be caused by the OS, might be python?

## Crash log

```
[UNEXPECTED_EXCEPTION] tests/lib/test_math_q64.cairo
Traceback (most recent call last):
  File "hypothesis/core.py", line 729, in _execute_once_for_engine
  File "hypothesis/core.py", line 667, in execute_once
  File "hypothesis/executors.py", line 47, in default_new_style_executor
  File "hypothesis/core.py", line 663, in run
  File "protostar/commands/test/environments/fuzz_test_execution_environment.py", line 240, in inner
  File "asyncio/events.py", line 762, in new_event_loop
  File "asyncio/events.py", line 660, in new_event_loop
  File "asyncio/unix_events.py", line 51, in __init__
  File "asyncio/selector_events.py", line 57, in __init__
  File "selectors.py", line 511, in __init__
OSError: [Errno 24] Too many open files

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "protostar/commands/test/test_runner.py", line 96, in run_test_suite
  File "protostar/commands/test/test_runner.py", line 157, in _run_test_suite
  File "protostar/commands/test/environments/factory.py", line 36, in invoke_test_case
  File "protostar/commands/test/environments/fuzz_test_execution_environment.py", line 155, in invoke
  File "protostar/commands/test/environments/fuzz_test_execution_environment.py", line 222, in to_thread
  File "concurrent/futures/thread.py", line 57, in run
  File "protostar/commands/test/environments/fuzz_test_execution_environment.py", line 151, in test_thread
[...]

PR #594, opened 2022-08-08
Add a test for #590 to our testing suite

mkaput:
cc #590

### DIFF
--- a/tests/integration/fuzzing/issue_590_test.cairo
+++ b/tests/integration/fuzzing/issue_590_test.cairo
@@ -1,0 +1,13 @@
+%lang starknet
+
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
+
+@external
+func test_safe_cast_fuzz{range_check_ptr, bitwise_ptr : BitwiseBuiltin*}(val : felt):
+    %{
+        if (ids.val % 2 == 0 and ids.val > (2**125 - 1) * 2):
+            expect_revert()
+    %}
+
+    return ()
+end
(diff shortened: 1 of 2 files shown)


tests/integration/fuzzing/issue_590_test.cairo
```diff
@@ -0,0 +1,13 @@
...
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
+
+@external
+func test_safe_cast_fuzz{range_check_ptr, bitwise_ptr : BitwiseBuiltin*}(val : felt):
+    %{
+        if (ids.val % 2 == 0 and ids.val > (2**125 - 1) * 2):
+            expect_revert()
+    %}
+
+    return ()
```
kasperski95, line 12: I don't understand this test case. Confusing name and magic numbers.

kasperski95 (commented):
I am not sure if we should add a test case if the problem couldn't be replicated.

Maybe it's worth testing fuzzing in the integration mode instead. Contracts can be declared/deployed from `__setup__` or from test cases. The root of that issue is the following error:

`OSError: [Errno 24] Too many open files`

tests/integration/fuzzing/issue_590_test.cairo
mkaput: I have renamed the test to match issue number, but I cannot help here more, because I also do not understand what is going on here and I would love to learn what codepaths are crashing here.

kasperski95 (approved):

PR #600, opened 2022-08-10
Declare command wallet support

Arcticae:
### Code
- [x] Accept legacy network names, add a deprecation warning about them

### Tests
- [x] Test for network argument correctness
- [x] Test for signing deploy transactions, possibly verifying the signature value?
- [x] Test for providing custom signer class
- [x] Test for default signer class usage
- [x] Test parsing & providing various formats of  private key
- [x] Test behavior when account does not exist, or format is invalid

### Docs
- [x] Guidance with pythonnpath in the docs (alignment with cairo-path)
- [x] Guidance with `BaseSigner` interface compliance, for the `signer_class` parameter

### Misc 
- [x] Make issues for fixing stuff when starknet.py updates the `ChainId` type, signing is required, and `StarkCurveSigner` supports signing `Declare`s. Add references in the code 
- [x] Make issue for considering a different signer instantiation logic
- [x] Self-review


protostar/commands/declare/declare_command.py
```diff
@@ -12,7 +16,7 @@
 )
 
 
-class DeclareCommand(Command):
+class DeclareCommand(Command, SignableCommandMixin, NetworkCommandMixin):
```
kasperski95, line 19: Let me quote you from hour ago:
`composition > inheritance`

protostar/starknet_gateway/network_config.py
```diff
@@ -40,32 +47,36 @@ def build(
...
-        if starkware_network_name not in NETWORKS:
+    def from_starknet_network_name(cls, starkware_network_name: str) -> "NetworkConfig":
+        if starkware_network_name not in NetworkConfig.get_starknet_networks():
             raise UnknownStarkwareNetworkException()
 
         contract_explorer_search_url_mapping = {
-            "alpha-goerli": "https://goerli.voyager.online/contract",
-            "alpha-mainnet": "https://voyager.online/contract",
+            TESTNET: "https://goerli.voyager.online/contract",
+            MAINNET: "https://voyager.online/contract",
```
kasperski95, line 59: Starknet CLI uses `alpha-goerli` etc.

At the other hand, I can never memorize these names, so maybe it's better to change them.

protostar/commands/declare/declare_command.py
Arcticae: it's a mixin. it's not inheritance, inheritance is only a mechanism here to add functionality.

protostar/cli/signable_command_mixin.py
```diff
@@ -0,0 +1,86 @@
...
+                type="path",
+            ),
+            Command.Argument(
+                name="signer-class",
+                description="Custom signer class module path.",
+                type="str",
+            ),
+            Command.Argument(
+                name="pythonpath-ext",
+                description="PYTHONPATH extensions, for custom code roots registration. You can extend PYTHONPATH with "
```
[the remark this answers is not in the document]
Arcticae: it's all doable with pythonpath.

protostar/starknet_gateway/network_config.py
Arcticae: in my mind it's better to keep things similar to starknet.py

protostar/commands/declare/declare_command.py
Arcticae: i can mix the functionality in with some other mechanism - no problem, if you consider that an issue.

Arcticae: https://en.wikipedia.org/wiki/Mixin

kasperski95: As long as mixin doesn't have deps it should be ok.

protostar/starknet_gateway/network_config.py
mkaput: didn't @Radinyn write related functions somewhere in deploy cheatcode impl?

protostar/cli/signable_command_mixin.py
```diff
@@ -0,0 +1,86 @@
...
+        signer = None
+        if args.pythonpath_ext:
+            sys.path.extend(args.pythonpath_ext)
+        if args.signer_class:
+            *module_names, class_name = args.signer_class.split(".")
+            module = ".".join(module_names)
+            signer_module = importlib.import_module(module)
+            signer_class = getattr(signer_module, class_name)
+
+            signer = signer_class(**signer_args)
```
mkaput, line 52: it would be nice to ensure that signer is fulfilling the signer protocol, at least by checking that it is an instance of some base class (I don't know what contracts we have here).

protostar/cli/signable_command_mixin.py
```diff
@@ -0,0 +1,86 @@
...
+                description="Custom signer class module path.",
+                type="str",
+            ),
+            Command.Argument(
+                name="pythonpath-ext",
+                description="PYTHONPATH extensions, for custom code roots registration. You can extend PYTHONPATH with "
+                "the same result as well.",
+                type="str",
+                is_array=True,
+            ),
```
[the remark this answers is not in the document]
mkaput: also I find that an ability to extend PYTHONPATH from CLI may end up being tempting to craft horrible development/CI flows with strange dependencies to directories outside repository.

maybe we should keep PYTHONPATH aligned with our CAIRO PATH, so `src` and `tests` etc. directories are included by default, and users should put the Signer class in a file in src directory?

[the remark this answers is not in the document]
Arcticae: on the other hand, not providing this option might be limiting for some users. i will make cairo-path aligned with pythonpath and we can give recommendations in the docs.

protostar/starknet_gateway/network_config.py
kasperski95: Maybe we should support both naming styles and deprecate the StaknetCLI's names in the future versions.

Arcticae: I kind of don't understand why we would do that, do we have a version released which does anything with networks?

kasperski95: The `deploy` command was released in version 0.2.1, and it supported network name similarly to Starknet CLI.

mkaput (commented):
looks good

protostar/cli/network_command_mixin.py
```diff
@@ -60,9 +60,9 @@ def get_network_config(args) -> NetworkConfig:
         )
 
     @staticmethod
-    def get_gateway_client(args) -> GatewayFacade:
+    def get_gateway_client(args) -> GatewayClient:
         network_config = NetworkCommandMixin.get_network_config(args)
         return GatewayClient(
             net=network_config.gateway_url,
-            chain=network_config.chain_id,
+            chain=network_config.chain_id,  # type: ignore
```
mkaput, line 78:
```suggestion
            chain=network_config.chain_id,
```

why?

protostar/starknet_gateway/gateway_facade.py
```diff
@@ -137,16 +138,17 @@ async def declare(
             nonce=nonce,
             version=0,
             signature=[],
-        )
+        )  # type: ignore
 
-        signature = signer.sign_transaction(unsigned_tx) if signer else []
+        # TODO(arcticae): Remove the if/else, when signing is made compulsory
```
mkaput, line 143:
```suggestion
        # TODO(arcticae): Remove the if/else, when signing is made mandatory
```

protostar/cli/network_command_mixin.py
Arcticae: i should explain in the comment, right. This one is expecting chain_id of `StarknetChainId` type, although starknet.py will accept any value (numeric included) and will work just fine

protostar/starknet_gateway/network_config.py
Arcticae: done

protostar/cli/network_command_mixin.py
Arcticae: In fact this is not even used in the client. The constructor is going to get updated pretty soon (0.10)

docs_generator/reference_docs_generator.py
```diff
@@ -34,7 +34,6 @@ def generate_cli_reference_markdown(self) -> str:
 
         return "\n".join(result)
 
-    # pylint: disable=no-self-use
```
Arcticae, line 37: no-self-use was removed in some version of pylint. let me know if i should restore it, it's available with optional extension

protostar/cli/signable_command_mixin.py
```diff
@@ -0,0 +1,170 @@
...
+        except ValueError as v_err:
+            raise ProtostarException(
+                f"Invalid account address format ({args.account_address}). Please provide hex-encoded number."
+            ) from v_err
+
+        return signer
+
+    @staticmethod
+    def _validate_signer_interface(signer_class):
+        if not issubclass(signer_class, BaseSigner):
```
Arcticae, line 143: i am wondering if there's any reason we should be worried that somehow the imported class extends the wrong `BaseSigner`? I'm guessing that there's no possibility to import i.e. a different version of starknet.py in the same run, thus making interface different. @mkaput i would appreciate your insight on this

protostar/commands/migrate/migrate_command.py
```diff
@@ -3,21 +3,22 @@
...
 from protostar.utils.input_requester import InputRequester
 from protostar.utils.log_color_provider import LogColorProvider
 
 
-class MigrateCommand(Command):
-    GATEWAY_URL_ARG_NAME = "gateway-url"
-    NETWORK_ARG_NAME = "network"
-
+class MigrateCommand(Command, NetworkCommandMixin, SignableCommandMixin):
+    # pylint: disable=too-many-arguments
```
Arcticae, line 17: i'd remove this check, pretty useless TBH

protostar/commands/test/fuzzing/strategy_selector_test.py
```diff
@@ -32,7 +32,7 @@ def test_pointer_parameter():
     with pytest.raises(
         FuzzingError,
         match=re.escape(
-            "Parameter 'x' cannot be fuzzed: " "Type felt* cannot be fuzzed."
+            "Parameter 'x' cannot be fuzzed: " + "Type felt* cannot be fuzzed."
```
Arcticae, line 35: implicit string concatenation, warning by pylint

protostar/commands/test/test_collector.py
```diff
@@ -104,7 +102,7 @@ def __init__(
                 broken_test_suites or []
             )
             self.test_cases_count = sum(
-                [len(test_suite.test_case_names) for test_suite in test_suites]
+                len(test_suite.test_case_names) for test_suite in test_suites
```
Arcticae, line 105: also warning by pylint, no need for the list construction here

tests/e2e/conftest.py
```diff
@@ -170,9 +170,12 @@ def _protostar(args: List[str], ignore_exit_code=False) -> str:
     return _protostar
 
 
-@pytest.fixture(name="devnet_gateway_url", scope="module")
-def devnet_gateway_url_fixture(devnet_port: int):
+@pytest.fixture(name="devnet_gateway_url", scope="function")
```
Arcticae, line 173: had to change devnet scopes to function. something weird happens when sharing the same devnet instance between migrator tests

Arcticae: Maybe we could parallelize those?

Arcticae: I heard from the guys that cov has problem with this, so maybe it's not the way to go

docs_generator/reference_docs_generator.py
mkaput: that's fine, imho this lint was to aggressive

protostar/commands/migrate/migrate_command.py
mkaput: 👍🏻 but plz do this in separate PR

protostar/cli/network_command_mixin.py
mkaput: OK.

tests/integration/migrator/conftest.py
```diff
@@ -21,15 +23,21 @@ def project_root_path_fixture(shared_datadir: Path) -> Path:
...
     )
+    return builder
 
-    migrator_builder.set_network(devnet_gateway_url)
 
-    return migrator_builder
+@pytest.fixture(name="gateway_facade")
+def gateway_facade_fixture(devnet_gateway_url, project_root_path):
+    client = GatewayClient(devnet_gateway_url, chain=StarknetChainId.TESTNET)
+    return GatewayFacade(project_root_path=project_root_path, gateway_client=client)
```
mkaput, line 40: what do you think about making `Migrator.Builder` accept `gateway` via `set_gateway` method and therefore let tests forget about having to import `gateway_facade` fixture?

from application code perspective this is indifferent as far as I see

mkaput (changes requested):
Does `poetry.lock` need to be updated in this PR? 😃 It's not obvious from the PR scope.

docs_generator/reference_docs_generator.py
kasperski95: https://swmansion.slack.com/archives/C02MBPPR2HY/p1653408429798169

protostar/cli/signable_command_mixin.py
mkaput: I think that if somebody is importing stuff from external `starknet.py` module than ours, then they are highly probable to use something incompatible with our instance and thus we should reject such objects.

protostar/cli/signable_command_mixin_test.py
```diff
@@ -0,0 +1,117 @@
...
+    network_config = mocker.MagicMock()
+    network_config.chain_id = StarknetChainId.TESTNET.value
+
+    args.signer_class = None
+    args.private_key_path = str(pkey_file_factory("0x123"))
+    args.account_address = "0x123"
+
+    signer = SignableCommandMixin.get_signer(args, network_config, logger)
+    assert isinstance(signer, PatchedStarkCurveSigner)
+    # TODO: Test signing
```
mkaput, line 65: missing code?

website/docs/tutorials/06-deploying/01-cli.md
```diff
@@ -30,6 +30,37 @@ Transaction hash: 0x3c6c...
...
+
+As of cairo v0.10 you will be able to sign declare transactions. 
+Our CLI will also provide 2 ways of providing the signature:
+
+### 1. Using the default signer from starknet.py - [StarkCurveSigner](https://starknetpy.readthedocs.io/en/latest/signer.html#module-starknet_py.net.signer)
+
+This way requires you to pass a private key (for signing) and account contract's address (to fetch the nonce). 
+
+2 options are used for this:
+- `private-key-path` - a path to the file containing hex-encoded private key
```
mkaput, line 43: how to obtain a keypair for signing? I think it would be nice to explain this in our docs or at least refer to source.

website/docs/tutorials/06-deploying/01-cli.md
```diff
@@ -30,6 +30,37 @@ Transaction hash: 0x3c6c...
...
+- `account-address` - your account contract's address (hex-encoded as well) on the appropriate network
+
+Alternatively, if you prefer not to store private key in a file, we check for `ACCOUNT_PRIVATE_KEY` environment variable, and use it if it's available.   
+It should be in the same hex-encoded format, like all the options above.
+
+### 2. Using a custom signer class
+
+You can provide a custom signer class which conforms to [BaseSigner](https://starknetpy.readthedocs.io/en/latest/signer.html#starknet_py.net.signer.BaseSigner) Abstract Base Class specification.
```
mkaput, line 51:
```suggestion
You can provide a custom signer class which inherits from [BaseSigner](https://starknetpy.readthedocs.io/en/latest/signer.html#starknet_py.net.signer.BaseSigner) abstract class. 
```

website/docs/tutorials/06-deploying/01-cli.md
```diff
@@ -30,6 +30,37 @@ Transaction hash: 0x3c6c...
...
+- `private-key-path` - a path to the file containing hex-encoded private key
+- `account-address` - your account contract's address (hex-encoded as well) on the appropriate network
+
+Alternatively, if you prefer not to store private key in a file, we check for `ACCOUNT_PRIVATE_KEY` environment variable, and use it if it's available.   
+It should be in the same hex-encoded format, like all the options above.
+
+### 2. Using a custom signer class
+
+You can provide a custom signer class which conforms to [BaseSigner](https://starknetpy.readthedocs.io/en/latest/signer.html#starknet_py.net.signer.BaseSigner) Abstract Base Class specification. 
+This way of signing requires you to write a class in python, which signs the transaction in a way that is suitable to you.
```
mkaput, line 52:
```suggestion
This way of signing requires you to write a class in Python, which signs the transaction in a way that is suitable to you.
```

website/docs/tutorials/06-deploying/01-cli.md
```diff
@@ -30,6 +30,37 @@ Transaction hash: 0x3c6c...
...
+After writing such class, simply use `signer_class` argument in the CLI for `declare` command to use that class instead of the default one.
+Usage of this way of signing is exclusive with the default signer strategy.
+
+:::caution
+The custom signer class must not take any arguments in the constructor, since we don't pass any args on instantiation.
+:::
+
+You might wonder where you should put this class - we synchronize `PYTHONPATH` with project's `cairo_path`s, so you can put this near to your cairo contracts source code.
+Modules that are dependencies of protostar (like starknet.py or cairo-lang) should be available for importing by default.
```
mkaput, line 61:
```suggestion
Modules that are dependencies of Protostar (like `starknet_py` or `cairo-lang`) should be available for importing by default.
```

website/docs/tutorials/06-deploying/01-cli.md
```diff
@@ -30,6 +30,37 @@ Transaction hash: 0x3c6c...
...
+You can provide a custom signer class which conforms to [BaseSigner](https://starknetpy.readthedocs.io/en/latest/signer.html#starknet_py.net.signer.BaseSigner) Abstract Base Class specification. 
+This way of signing requires you to write a class in python, which signs the transaction in a way that is suitable to you.
+After writing such class, simply use `signer_class` argument in the CLI for `declare` command to use that class instead of the default one.
+Usage of this way of signing is exclusive with the default signer strategy.
+
+:::caution
+The custom signer class must not take any arguments in the constructor, since we don't pass any args on instantiation.
+:::
+
+You might wonder where you should put this class - we synchronize `PYTHONPATH` with project's `cairo_path`s, so you can put this near to your cairo contracts source code.
```
mkaput, line 60:
```suggestion
The Python file containing this class can be put next to Cairo source code.
Protostar synchronizes `PYTHONPATH` with project's `cairo_path`.
```

website/docs/tutorials/06-deploying/02-migrations/README.md
```diff
@@ -80,6 +80,10 @@ If you build the project, Protostar will print migration logs in the command lin
   transaction_hash     0x1234...
 [INFO] Migration completed
 ```
+## Signing the migration
+You can sign the migration's transactions by providing appropriate arguments to the CLI of the command. See signing-related documentation [here](../01-cli.md#signing-a-declaration).
+
+[//]: # (TODO: Add more info when signing invokes in migrations is supported)
```
mkaput, line 86:
```suggestion
<!-- TODO: Add more info when signing invokes in migrations is supported. -->
```

mkaput (changes requested):

tests/integration/migrator/conftest.py
Arcticae: seems reasonable

Arcticae: i will make the builder store the intermediate value so calls can be executed in any order.

protostar/cli/signable_command_mixin.py
Arcticae: I am wondering if such situation would even be possible in python's module system, for 2 modules of the same name to coexist

website/docs/tutorials/06-deploying/01-cli.md
Arcticae: To be honest i don't even know what to refer to here XD

Arcticae: I can link to argentx, that's probably the most universal one, apart from that there is no clear demonstration on how to obtain the keys

protostar/cli/signable_command_mixin.py
```diff
@@ -0,0 +1,157 @@
...
+                account_address=args.account_address,
+                key_pair=key_pair,
+                chain_id=network_config.chain_id,
+            )  # FIXME(arcticae): Change the default signer to starknet.py one, when it supports signing declare txs
+        except ValueError as v_err:
+            raise ProtostarException(
+                f"Invalid account address format ({args.account_address}). Please provide hex-encoded number."
+            ) from v_err
+
+        return signer
```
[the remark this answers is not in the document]
Arcticae: this is a copy-paste and it's temporary until starknet.py implements the signer compliant with new specs. not worth the effort to do anything with this

tests/conftest.py
```diff
@@ -19,12 +32,20 @@ def run_devnet(devnet: List[str], port: int) -> subprocess.Popen:
...
+    if proc.stdout:
+        appended_stdout = f"stdout: {proc.stdout.read()}"
+
+    raise RuntimeError(
+        f"Devnet failed to start on port {port}" + (appended_stdout or "")
+    )
 
 
-@pytest.fixture(name="devnet_port", scope="session")
+@pytest.fixture(name="devnet_port", scope="function")
```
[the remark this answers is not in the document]
Arcticae: no

protostar/cli/network_command_mixin.py
```diff
@@ -0,0 +1,80 @@
...
+        # TODO(arcticae): Remove in the future version, with the legacy names formats support
+        if args.network and is_legacy_network_name(args.network):
+            logger.warning(
+                f"{args.network} is a legacy network name parameter and it won't be supported in future versions"
+            )
+
+        if args.gateway_url and not args.chain_id:
+            raise ProtostarException(
+                f"Argument `{CHAIN_ID_ARG_NAME}` is required when `{GATEWAY_URL_ARG_NAME}` is provided"
+            )
```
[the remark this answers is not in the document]
Arcticae: it's a dependency of GatewayClient

tests/conftest.py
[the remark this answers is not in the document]
kasperski95: Maybe it's worth creating different variants of devnet fixtures: e.g.
- `session_devnet_port`
- `module_devnet_port` etc.

Integration tests will be much slower with the `function` scope.

protostar/cli/signable_command_mixin.py
```diff
@@ -0,0 +1,157 @@
...
+
+from protostar.cli import Command
+
+from protostar.protostar_exception import ProtostarException
+from protostar.starknet_gateway import NetworkConfig
+
+PRIVATE_KEY_ENV_VAR_NAME = "ACCOUNT_PRIVATE_KEY"
+
+
+class PatchedStarkCurveSigner(BaseSigner):
```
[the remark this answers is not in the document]
Arcticae: it's going to get removed from here anyway

protostar/cli/signable_command_mixin.py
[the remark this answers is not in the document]
kasperski95: You can add a comment pointing to the original file.

protostar/cli/network_command_mixin.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @staticmethod
+    def get_network_config(args: Any, logger: Logger) -> NetworkConfig:
+        NetworkCommandMixin.validate_network_command_args(args, logger)
+        return NetworkConfig.build(
+            args.gateway_url, args.network, chain_id=args.chain_id
+        )
+
+    @staticmethod
+    def get_gateway_client(args: Any, logger: Logger) -> GatewayClient:
```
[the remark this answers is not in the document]
Arcticae: logger was supposed to be made static from what i've heard? So that's not going to be an issue anymore soon.

protostar/cli/network_command_mixin.py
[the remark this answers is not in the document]
Arcticae: BTW Not anymore in 0.10 for starknet.py - so that's going to get required only for default signer usage scenario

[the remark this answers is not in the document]
kasperski95: Isn't that going to be removed soon? I heard that the same chain_id is used by devnet and testnet so that could be the default value.

protostar/composition_root.py
```diff
@@ -151,14 +149,14 @@ def build_di_container(script_root: Path, start_time: float = 0):
...
-        DeclareCommand(gateway_facade_builder, logger),
+        DeployCommand(logger=logger, project_root_path=project_root_path),
+        DeclareCommand(logger=logger, project_root_path=project_root_path),
         MigrateCommand(
             migrator_builder=Migrator.Builder(
                 migrator_execution_environment_builder=MigratorExecutionEnvironment.Builder(),
-                gateway_facade_builder=GatewayFacade.Builder(project_root_path),
                 project_root_path=project_root_path,
             ),
+            project_root_path=project_root_path,
```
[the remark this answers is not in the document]
Arcticae: We can use a fixture for those. If that's the price for getting rid of a builder i can take it.

protostar/cli/network_command_mixin.py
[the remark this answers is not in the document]
Arcticae: What solution do you propose here to make this logic modular? This solution quite nicely decouples this logic from commands (only command needs to know about it's components, not the other way around)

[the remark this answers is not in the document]
kasperski95: `log_color_provider` can be used as a global variable.

Logger however is going to be abstracted and adapted to Protostar needs (to avoid interchangeably using logger and `print` function).

`getLogger()` would make sense if it would be used for internal needs, not as a way of presenting results to the user.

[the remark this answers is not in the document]
kasperski95: > What solution do you propose here to make this logic modular? This solution quite nicely decouples this logic from commands (only command needs to know about it's components, not the other way around)

Composition

tests/integration/deploying/declare_command_test.py
```diff
@@ -2,31 +2,83 @@
...
+    args.signer_class = None
+    args.account_address = "123"
+    args.private_key_path = None
+    args.contract = compiled_contract_filepath
+    args.gateway_url = devnet_gateway_url
+    args.network = None
+    args.token = None
+    args.wait_for_acceptance = True
+
+    response = await declare_command.run(args)
```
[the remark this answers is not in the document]
Arcticae: done

protostar/cli/signable_command_mixin_test.py
Arcticae: done

website/docs/tutorials/06-deploying/01-cli.md
Arcticae: done

protostar/cli/network_command_mixin.py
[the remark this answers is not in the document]
Arcticae: Yeah it'd be reasonable to provide a default one, but i suppose it could lead to signatures being incorrect. It could be hard to trace back though, since there's no direct mechanism for verifying this parameter in the contract, only the whole signature

protostar/cli/signable_command_mixin.py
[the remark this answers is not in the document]
Arcticae: done

protostar/cli/network_command_mixin.py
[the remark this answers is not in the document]
Arcticae: done

tests/conftest.py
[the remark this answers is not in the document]
Arcticae: can i deal with that in another PR?

protostar/cli/signable_command_mixin.py
mkaput: > I am wondering if such situation would even be possible in python's module system, for 2 modules of the same name to coexist

I don't know, but current code it capable of dealing with both situations :)

website/docs/tutorials/06-deploying/01-cli.md
```diff
@@ -30,6 +30,39 @@ Transaction hash: 0x3c6c...
 https://goerli.voyager.online/contract/0x038cc...
 ```
 
+### Signing a declaration
+
+As of cairo v0.10 you will be able to sign declare transactions. 
+Our CLI will also provide 2 ways of providing the signature:
```
mkaput, line 36:
```suggestion
Since Cairo v0.10 declare transactions can be signed.
Protostar offers two ways of providing the signature:
```

tests/conftest.py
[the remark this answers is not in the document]
kasperski95: Can you create a task for it?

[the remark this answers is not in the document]
Arcticae: https://github.com/software-mansion/protostar/issues/642

kasperski95 (approved):

PR #601, opened 2022-08-12
Merge unsigned() and signed() search strategies into felts() strategy

### LINKED ISSUE #597: Merge felt strategies into one
This requires further thinking about.

Currently, we have two felt producing strategies: `unsigned()` and `signed()`. I have an impression that there are no use cases for the `signed()` strategy. We may be better to merge these two into the `felt()` strategy, which works the same way as `unsigned()` and push #579 as the default so that range checks work out of the box.

mkaput:
This PR replaces the `unsigned()` and `signed()` fuzzing search strategies with a new (and now default) search strategy called simply `felts()`.

From Cairo code perspective, both strategies observed the same range of values, due to the modulo arithmetic of felts. The only difference between these two strategies laid in a method how input values were picked.

I did a bunch of benchmarks and I have concluded that exploring felts in range `MIN_FELT...MAX_FELT` made Hypothesis pick more variadic set of values. `signed()` explored almost always around 30% more distinct values than `unsigned()`.

Therefore, the `felts()` strategy has the same behaviour as former `signed()`.

fix #597


kasperski95 (approved):
LGTM However, this is a breaking change. IMO, unsigned and signed strategies should be deprecated rather than removed.

PR #608, opened 2022-08-12
Add `call` migration cheatcode doc

kasperski95:
Merge after https://github.com/software-mansion/protostar/pull/618

### DIFF
--- a/website/docs/tutorials/06-deploying/02-migrations/call.md
+++ b/website/docs/tutorials/06-deploying/02-migrations/call.md
@@ -11,7 +11,7 @@ class ContractCallResult(NamedTuple):
     ...
     ...
 ```
-Calls a StarkNet contract without affecting the state.
+Calls a StarkNet contract without affecting the StarkNet's state.
 
 :::tip
 You can provide `inputs` as a dictionary to use [data transformer](./README.md#data-transformer).
--- a/website/docs/tutorials/06-deploying/02-migrations/call.md
+++ b/website/docs/tutorials/06-deploying/02-migrations/call.md
@@ -5,7 +5,10 @@ def call(
     contract_address: int,
     function_name: str,
     inputs: list[int] | dict[str, Any] | None = None,
-) -> NamedTuple:
+) -> ContractCallResult:
+
+class ContractCallResult(NamedTuple):
+    ...
     ...
 ```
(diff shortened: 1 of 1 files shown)


website/docs/tutorials/06-deploying/02-migrations/call.md
```diff
@@ -0,0 +1,50 @@
...
+
+```python
+def call(
+    contract_address: int,
+    function_name: str,
+    inputs: Optional[Union[List[int], Dict[str, Any]]] = None,
+) -> NamedTuple:
+    ...
+```
+Calls a StarkNet contract without affecting the state.
```
mkaput, line 11: whose state is not affected? Mine or called contract's?

website/docs/tutorials/06-deploying/02-migrations/call.md
```diff
@@ -0,0 +1,50 @@
+# `call`
+
+```python
+def call(
+    contract_address: int,
+    function_name: str,
+    inputs: Optional[Union[List[int], Dict[str, Any]]] = None,
```
mkaput, line 7: Could we use modern Python's syntax for typing here? I think it would be much more readable:

```suggestion
    inputs: list[int] | dict[str, Any] | None = None,
```

website/docs/tutorials/06-deploying/02-migrations/call.md
```diff
@@ -0,0 +1,50 @@
...
+) -> NamedTuple:
```
mkaput, line 8:
```suggestion
) -> ContractCallResult:

class ContractCallResult(NamedTuple):
    ...
```

wdyt?

mkaput (changes requested):

website/docs/tutorials/06-deploying/02-migrations/call.md
kasperski95: Starknet state I guess. I copied this sentence from Starknet's docs.

https://docs.starknet.io/docs/CLI/commands/#starknet-call

mkaput: OK. But I think it would still be nice to clarify this

mkaput (approved):

PR #619, opened 2022-08-16
Add initial activity indicator and fix execution time

kasperski95:
- improved initial responsiveness (the initialization time is the same though)
- fixed execution time

https://user-images.githubusercontent.com/9629577/184952214-0358b953-0dd4-49a9-9b32-6dc00de46136.mov

### DIFF
--- a/binary_entrypoint.py
+++ b/binary_entrypoint.py
@@ -7,30 +7,36 @@ from threading import Thread
 from time import perf_counter, sleep
 
 import certifi
+from colorama.ansitowin32 import StreamWrapper
 
 SCRIPT_ROOT = Path(__file__).parent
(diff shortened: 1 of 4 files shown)


mkaput (changes requested):
I am 99% percent sure this will break non-TTY outputs. We should only display this indicator in interactive environment.

At least we should check https://docs.python.org/3/library/os.html?highlight=isatty#os.isatty, but colorama seems to do more on top of it: https://github.com/tartley/colorama/blob/cd653d75be52f4d8c3953eb6942fe597375f8b97/colorama/ansitowin32.py#L49-L59

mkaput (approved):

PR #617, opened 2022-08-16
Added format command

Radinyn:

PR #627, opened 2022-08-17
`protostar format` in docs

Radinyn:
#617

### DIFF
--- a/website/docs/tutorials/08-formatting.md
+++ b/website/docs/tutorials/08-formatting.md
@@ -7,7 +7,7 @@ sidebar_label: Formatting
 Protostar comes with a built-in Cairo source code formatter.
 
 ```shell
-protostar format
+protostar format --help
 ```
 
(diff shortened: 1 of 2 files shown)


mkaput:
I would write a short guide about formatting code in order to make this feature more visible.

mkaput (approved):

website/docs/tutorials/08-formatting.md
```diff
@@ -0,0 +1,13 @@
+---
+sidebar_label: Formatting
+---
+
+# Formatting
+
+Protostar comes with a built-in Cairo source code formatter.
+
+```shell
+protostar format
```
MaksymilianDemitraszek, line 10: If i run it like this what files it will format? Would be nice to include such infromation

Radinyn: That information is included withing CLI Reference, I don't think we need to duplicate it, perhaps
```suggestion
protostar format --help
```
is more adequate?

PR #629, opened 2022-08-18
Deploy by contract name — Migrator deploy cheatcode

kasperski95:
- `deploy_contract` supports deploying by a contract name
- added `MaybeTmpDirectory` to store the potential output of migration
  - directory is created before the script runs to fail quickly if there's a problem with creating such directory 
- refactored remaining migration tests to instantiate Migrator in a one place


protostar/migrator/maybe_tmp_directory.py
```diff
@@ -0,0 +1,36 @@
...
+
+from protostar.protostar_exception import ProtostarException
+
+
+class MaybeTmpDirectory(AbstractContextManager):
+    def __init__(self, path: Path):
+        self._path = path
+
+    def __enter__(self) -> None:
+        self._create_dir()
```
mkaput, line 14: I would return `self._path` here so that usage places could do:

```python
with MaybeTmpDirectory(tmp_path / "tmp_dir") as output_path:
    ...
```

protostar/migrator/maybe_tmp_directory.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _create_dir(self):
+        if self._path.exists():
+            raise DirectoryExistsException(self._path)
+        self._path.mkdir()
+
+    def __exit__(self, __exc_type, __exc_value, __traceback) -> Union[bool, None]:
+        if self._is_directory_empty():
+            rmdir(self._path)
```
mkaput, line 23:
```suggestion
            self._path.rmdir()
```

https://docs.python.org/3/library/pathlib.html#pathlib.Path.rmdir

protostar/migrator/maybe_tmp_directory.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _is_directory_empty(self) -> bool:
+        return len(listdir(self._path)) == 0
```
mkaput, line 26:
```suggestion
        return any(self._path.iterdir())
```

https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir

protostar/migrator/maybe_tmp_directory.py
```diff
@@ -0,0 +1,36 @@
+from contextlib import AbstractContextManager
+from os import listdir, rmdir
+from pathlib import Path
+from typing import Union
+
+from protostar.protostar_exception import ProtostarException
+
+
+class MaybeTmpDirectory(AbstractContextManager):
```
mkaput, line 9: turbo nitpicky, but I find use of `@contextmanager` decorator probably would yield easier to read code here

protostar/migrator/migrator_datetime_state.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def get_output_stem(self) -> str:
+        assert self._datetime is not None
+        migration_file_stem = self._migration_file_path.stem
+        prefix = self.get_datetime_prefix(self._datetime)
+        return f"{prefix}_{migration_file_stem}"
+
+    def get_compilation_output_path(self) -> Path:
+        migrations_dir_path = self._migration_file_path.parent
+        return migrations_dir_path / self.get_output_stem()
```
mkaput, line 26: I find this abstraction a bit strange. What's the reason for this being a stateful class that has deferred initialization (`update_to_now` method) instead of just being a regular function that takes a path and DateTime object?

mkaput (changes requested):

protostar/migrator/migrator_datetime_state.py
kasperski95: I need to pass the prefix to the migrator cheatcode(s) and synchronize that prefix between migration history file name and migration output directory.

I don't see how function would solve this problem.

mkaput: If the timestamp of the migration was determined earlier/higher in abstraction, wouldn't it be possible to pass the prefix/path as a stateless data object?

mkaput: discussed privately

protostar/migrator/maybe_tmp_directory.py
```diff
@@ -1,29 +1,21 @@
...
-        self._path = path
+@contextmanager
+def MaybeTmpDirectory(path: Path):
+    if path.exists():
+        raise DirectoryExistsException(path)
 
-    def __enter__(self) -> None:
-        self._create_dir()
+    path.mkdir()
+    yield path
```
mkaput, line 14: I think you have to put this in `try...finally` in order to remove the directory in case of exceptions

protostar/migrator/maybe_tmp_directory.py
```diff
@@ -1,29 +1,21 @@
-from contextlib import AbstractContextManager
-from os import listdir, rmdir
+from contextlib import AbstractContextManager, contextmanager
```
mkaput, line 1:
```suggestion
from contextlib import contextmanager
```

protostar/migrator/maybe_tmp_directory.py
```diff
@@ -1,29 +1,21 @@
...
 from typing import Union
 
 from protostar.protostar_exception import ProtostarException
 
 
-class MaybeTmpDirectory(AbstractContextManager):
-    def __init__(self, path: Path):
-        self._path = path
+@contextmanager
+def MaybeTmpDirectory(path: Path):
```
mkaput, line 9:
```suggestion
def maybe_temporary_directory(path: Path):
```

mkaput (changes requested):

protostar/migrator/maybe_tmp_directory.py
```diff
@@ -0,0 +1,28 @@
+from contextlib import contextmanager
+from pathlib import Path
+
+from protostar.protostar_exception import ProtostarException
+
+
+@contextmanager
+def create_maybe_tmp_directory(path: Path):
```
mkaput, line 8:
```suggestion
def create_build_directory(path: Path):
```

I didn't like the `create maybe` phrase. And I think a more specific name could be better.

protostar/migrator/maybe_tmp_directory.py
```diff
@@ -0,0 +1,28 @@
+from contextlib import contextmanager
```
mkaput, line 1: file name does not match function name

mkaput (approved):

protostar/migrator/maybe_tmp_directory.py
kasperski95: If the file name matches the function name, it:
- breaks patching
- autoimport imports a module rather than a function (at least in VS Code)

protostar/migrator/maybe_tmp_directory.py
kasperski95: Renamed to `create_output_directory`

protostar/migrator/maybe_tmp_directory.py
mkaput: 🤨

mkaput (approved):

PR #639, opened 2022-08-22
Fixed hex as inputs

Radinyn:


protostar/cli/command.py
```diff
@@ -32,6 +32,11 @@ def directory(arg: str) -> Path:
                 assert pth.is_dir(), f'"{str(pth)}" is not a valid directory path'
                 return pth
 
+            @staticmethod
+            def int(arg: str) -> int:
+                # Enable hex (and other) int representations
```
kasperski95, line 37: Please convert this comment to a test case(s).

protostar/cli/command_test.py
```diff
@@ -19,3 +19,10 @@ def test_directory_type(tmpdir):
 def test_directory_type_checks_existence():
     with pytest.raises(AssertionError):
         Command.Argument.Type.directory("!@#$")
+
+
+def test_int_type():
+    ints = ["123", "0xf", "0b001", "0o444", "0xDEADC0DE"]
+    for i in ints:
+        result = Command.Argument.Type.int(i)
+        assert isinstance(result, int)
```
kasperski95, line 28: It could be a parametrized test. You could also check if all representation are converted to the same value.

kasperski95 (approved):

protostar/cli/command_test.py
mkaput: could we also convert one e2e test to use this feature?

mkaput (changes requested):

tests/e2e/test_testing.py
```diff
@@ -232,7 +232,7 @@ def test_seed_and_max_fuzz_examples_impact_testing_results(
     protostar: ProtostarFixture, copy_fixture
 ):
     copy_fixture("test_fuzz.cairo", "./tests")
-    seed = str(213742)
+    seed = hex(213742)
```
mkaput, line 235: I don't want to support hex ints for seed. It should only allow decimal values.

mkaput (changes requested):

protostar/cli/command.py
```diff
@@ -32,6 +34,14 @@ def directory(arg: str) -> Path:
                 assert pth.is_dir(), f'"{str(pth)}" is not a valid directory path'
                 return pth
 
+            @staticmethod
+            def felt(arg: str) -> int:
+                try:
+                    [output] = cast_to_felts([arg])
+                except ValueError as ex:
+                    assert False, str(ex)
```
mkaput, line 42: @kasperski95 is there a need for raising `AssertionError` here? If not they I think @Radinyn we can remove this try..except clause

protostar/commands/deploy/deploy_command.py
```diff
@@ -82,7 +82,7 @@ def arguments(self) -> List[Command.Argument]:
                     # pylint: disable=line-too-long
                     "[Read more about representing Cairo data types in the CLI.](https://www.cairo-lang.org/docs/hello_starknet/more_features.html#array-arguments-in-calldata)"
                 ),
-                type="int",
+                type="felt",
                 is_array=True,
             ),
             Command.Argument(
```
mkaput, line 88: argument `salt` could also be a `felt` I think @kasperski95 correct me if I'm wrong.

mkaput (changes requested):

protostar/commands/deploy/deploy_command.py
kasperski95: Screenshot 2022-08-22 at 14 11 44

protostar/cli/command.py
mkaput: By 👍🏻 reaction I understand we can remove try..except

mkaput (approved):

protostar/cli/command.py
kasperski95: It's better to throw the exception.

We need to refine this code in the future by introducing `ArgumentValidationError` in order to tell the user which argument caused the error.

kasperski95 (approved):

Issue #638, opened 2022-08-22
install.sh script should not allways install the last version of protostar, regardless of the specified branch


tdelabro:
As a StarkNet smart contract developer, I want to install a past version of protostar.
The `install.sh` script always uses the latest release and version, regardless of the branch we are using in the curl command.

`curl -L https://raw.githubusercontent.com/software-mansion/protostar/master/install.sh | bash`
and
`curl -L https://raw.githubusercontent.com/software-mansion/protostar/v0.3.0/install.sh | bash`
both resolve to `v0.3.1`.

This is highly counterintuitive and should not be the behavior.

kasperski95:
Related https://github.com/software-mansion/protostar/issues/570

kasperski95:
Reopening. https://github.com/software-mansion/protostar/issues/570 should allow specifying a specific version from the Protostar CLI.

PR #649, opened 2022-08-23
Add option to specify install version

### LINKED ISSUE #570: Capability to specify the version to install/upgrade
As a StarkNet smart contract developer, I want to **pin the exact Protostar version to install, or update to**, so that **the devenv setup is reproductible, and kept stable**.

For instance, installing Protostar in our CI will always install the latest version available, and our code wasn't migrated to Cairo 0.9.1 at the time, so all our tests start failing.

It would be very powerful to be able to specify the desired version at install, and upgrade, by looking at the expected version from the configuration file.

WDYT?

EdgarBarrantes:
Adds option `-v` to install script, in the lack of if, it will install latest version.

On specifying a non existent version, it will fail gracefully:

```bash
$ ./install.sh -v 0.3.33
Installing protostar
Retrieving 0.3.33 version from https://github.com/software-mansion/protostar...
Version 0.3.33 not found
```

Fixes #570 #638

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,12 +35,6 @@ To install the tool, run:
 curl -L https://raw.githubusercontent.com/software-mansion/protostar/master/install.sh | bash
 ```
 
-If you want to specify a version:
-
-```shell
-curl -L https://raw.githubusercontent.com/software-mansion/protostar/master/install.sh | bash -s -- -v 0.3.2
-```
-
 ---
 
 ## Development
--- a/README.md
+++ b/README.md
@@ -34,6 +34,13 @@ To install the tool, run:
 ```shell
 curl -L https://raw.githubusercontent.com/software-mansion/protostar/master/install.sh | bash
 ```
+
+If you want to specify a version:
+
+```shell
+curl -L https://raw.githubusercontent.com/software-mansion/protostar/master/install.sh | bash -s -- -v 0.3.2
+```
+
 ---
 
 ## Development
--- a/install.sh
+++ b/install.sh
@@ -38,9 +38,9 @@ while getopts ":v:" opt; do
 done
 
 if [ -n "$VERSION" ]; then
-  DESIRED_VERSION="tag/v${VERSION}"
+  REQUESTED_REF="tag/v${VERSION}"
 else
-  DESIRED_VERSION="latest"
+  REQUESTED_REF="latest"
   VERSION="latest"
 fi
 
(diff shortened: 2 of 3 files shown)


kasperski95:
Related https://github.com/software-mansion/protostar/pull/650

README.md
```diff
@@ -34,6 +34,13 @@ To install the tool, run:
 ```shell
 curl -L https://raw.githubusercontent.com/software-mansion/protostar/master/install.sh | bash
 ```
+
+If you want to specify a version:
```
MaksymilianDemitraszek, line 38: We may include it in docs, for readme seems like a too niche information

install.sh
```diff
@@ -21,18 +21,47 @@ case $PLATFORM in
...
+      ;;
+    :)
+      echo "Option -$OPTARG requires an argument." >&2
+      exit 1
+      ;;
+  esac
+done
+
+if [ -n "$VERSION" ]; then
+  DESIRED_VERSION="tag/v${VERSION}"
```
mkaput, line 41: I think word `requested` is more adequate here instead of `desired`. Wdyt?

https://wikidiff.com/requested/desired

mkaput, line 41:
```suggestion
  REQUESTED_REF="tag/v${VERSION}"
```

this variable is holding git reference instead of version

mkaput (changes requested):
Looks good! I have minor comments regarding wording here 😄

install.sh
EdgarBarrantes: I agree, thanks for the suggestion!

mkaput:
@EdgarBarrantes are you working on the documentation part as @MaksymilianDemitraszek asked in review? We can do this by ourselves and merge this PR if you wish 😄 This is a great contribution!

EdgarBarrantes:
I wasn't, but I just added them, let me know if they make sense, thank you! :smile: 
And thanks for helping integrate this so quickly!

mkaput (approved):
❤️ ❤️ ❤️

EdgarBarrantes:
@kasperski95 want me to rebase to run the tests you just merged? Thanks a lot for that btw.

kasperski95:
@EdgarBarrantes There's no need, but it would be nice to see passing tests.

EdgarBarrantes:
There we go :crossed_fingers:
Thank you!

kasperski95 (approved):

PR #650, opened 2022-08-23
`install.sh` tests

kasperski95:


tests/e2e/test_install_sh/dist/protostar/info/pyproject.toml
```diff
@@ -0,0 +1,159 @@
+[tool.poetry]
```
mkaput, line 1: would it make sense to use a symlink here instead of coping the file?

tests/e2e/test_install_sh.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def expect_detected_shell(self, shell_name: str) -> None:
+        self.expect(f"Detected your preferred shell is {shell_name}")
+
+    def send(self, value: str) -> None:
+        self._process.sendline(value)
+
+
+def run_testing_harness(home_path: Path, shell: str) -> ScriptTestingHarness:
+    command = f"bash ./install__testing_harness.sh {home_path} {shell}"
```
mkaput, line 43:
```suggestion
    command = f"bash ./install_testing_harness.sh {home_path} {shell}"
```

Why does this file have duplicated underscore in its name?

mkaput (approved):
I like this!

tests/e2e/test_install_sh.py
kasperski95: The name would suggest the script installs a testing harness. I didn't come up with a better name.

tests/e2e/test_install_sh/dist/protostar/info/pyproject.toml
kasperski95: Actually, this file can be removed.

tests/e2e/test_install_sh.py
mkaput: I did understood from the context that this is a testing harness for the install script hence I find this underscore superfluous. But this is nitpick which doesn't block this code from merging.

kasperski95: Renamed

tests/e2e/test_install_sh/dist/protostar/info/pyproject.toml
kasperski95: Removed

Issue #630, opened 2022-08-18
Set fuzz max examples in setup case instead of CLI argument


mkaput:
```cairo
def __setup__():
    %{
        # Can also be set for all test cases
        # (overriden in setup cases)
        max_examples(10)
    %}
end

def setup_foo_bar_baz():
    %{
        max_examples(20)        
        given(a = integers(10, 50), b = felts())
    %}
end

def test_foo_bar_baz(a : felt, b : felt):
    ...
end
```

related to #625

PR #652, opened 2022-08-24
Introduce stateful TestConfig class for tracking any test settings during runtime

mkaput:


kasperski95 (approved):
I'm not a fan of `Remove this` comments (and comments in general). 

These kinds of comments cause more harm than good, because they introduce confusion (Is it safe to remove? If so, why it wasn't removed?)

mkaput:
> I'm not a fan of `Remove this` comments (and comments in general).
> 
> These kinds of comments cause more harm than good, because they introduce confusion (Is it safe to remove? If so, why it wasn't removed?)

The PR for #630 will remove these. I put these markers here to note that this parts will be removed soon.

PR #658, opened 2022-08-24
Move TestExecutionEnvironment selection into TestCaseRunner class in preparation for invoking setup cases

mkaput:
This PR moves the logic of constructing `TestExecutionEnvironment` into the `TestCaseRunner` class, because future invokes of setup cases will influence this logic path. Because of that, the inheritance structure of `TestCaseRunner` was entirely dropped and now `TestCaseRunner` is self-contained.

This PR also moves `TestExecutionResult` classes into `test_results.py` module, because this solves the cyclic imports conflict that arose.


protostar/commands/test/test_case_runners/test_case_runner_factory.py
```diff
@@ -29,15 +26,16 @@ def make(self, test_case: TestCase) -> TestCaseRunner:
...
 
         if state.config.mode is TestMode.FUZZ:
-            return FuzzTestCaseRunner(
-                fuzz_test_execution_environment=FuzzTestExecutionEnvironment(state),
+            return TestCaseRunner(
+                execution_environment=FuzzTestExecutionEnvironment(state),
                 test_case=test_case,
                 output_recorder=state.output_recorder,
+                test_case_result_decorator=FuzzTestCaseResultDecorator(),
             )
```
kasperski95, line 34: `TestCaseRunner` isn't designed well. It allows providing incompatible environments and decorators. Can you constrain it somehow with generics? The previous solution didn't allow that.

IMO, Template method pattern was better choice than decorators.

protostar/commands/test/fuzzing/fuzz_test_case_result_decorator.py
```diff
@@ -0,0 +1,50 @@
...
+    PassedTestCaseResult,
+    TestCaseResultDecorator,
+)
+
+
+class FuzzTestCaseResultDecorator(TestCaseResultDecorator):
+    def decorate_passed(
+        self,
+        result: PassedTestCaseResult,
+        execution_result: FuzzTestExecutionResult,
```
kasperski95, line 22: pyright should've scream here...

When `reportIncompatibleMethodOverride` the following error is thrown:
```
[REDACTED]/protostar/commands/test/fuzzing/fuzz_test_case_result_decorator.py
  [REDACTED]/protostar/commands/test/fuzzing/fuzz_test_case_result_decorator.py:19:9 - error: Method "decorate_passed" overrides class "TestCaseResultDecorator" in an incompatible manner
    Parameter 3 type mismatch: base parameter is type "TestExecutionResult", override parameter is type "FuzzTestExecutionResult"
      "TestExecutionResult" is incompatible with "FuzzTestExecutionResult" (reportIncompatibleMethodOverride)
```

(Btw. we need to enable this check.)

kasperski95 (changes requested):
Requesting changes due to `incompatibleMethodOverride`.

protostar/commands/test/test_case_runners/test_case_runner_factory.py
mkaput: > IMO, Template method pattern was better choice than decorators.

Don't decorators inheritance realize exactly the same pattern, but decorators themselves are encapsulated via a strategy pattern now?

mkaput:
> Requesting changes due to `incompatibleMethodOverride`.

As far as I see this option is totally incompatible with Cairo's Visitor API...

Arcticae (commented):
It is not really flexible to include modular logic passed like this into the test case runner - i was thinking about kind of a pipeline oriented processing of results, perhaps something like that could be done? There could be a different pipeline for different types of cases. If further processing of results is not something we'd like to do in the future, it's fine as-is

kasperski95:
> As far as I see this option is totally incompatible with Cairo's Visitor API...

Probably that's why this check is disabled. However, because this check is disabled, it doesn't mean we should break this rule if there's such possibility.

mkaput:
I am thinking about how to remove the decorators abstraction here. It probably will never be extended, except maybe adding some extra fields related to fuzzing.

I wonder if we really need the fuzzing related inheritance in the `TestExecutionResult` class family. If this class was final and fuzz related data was optional fields there, there wouldn't be any problems.

mkaput:
We have discussed this privately with @kasperski95. I will submit in this PR an alternative approach that we have just worked out.

protostar/commands/test/test_case_runners/test_case_runner_factory.py
mkaput: this comment is no more relevant

protostar/commands/test/fuzzing/fuzz_test_case_result_decorator.py
mkaput: Done.

kasperski95 (commented):
IMO this code introduce bad design in the Protostar's "core" where the code should be pristine. A lot of code seems forced:
`TestMode`, `if isinstance`, `TEST_CASE_RESULT_TYPES`, unlogical function name, and operating on classes as first-class citizens.

I am leaving the decision whether to merge or not to other team members.

protostar/commands/test/test_results.py
```diff
@@ -45,38 +80,49 @@ class FuzzResult:
 @dataclass(frozen=True)
 class PassedFuzzTestCaseResult(PassedTestCaseResult, FuzzResult):
     @classmethod
-    def from_passed_test_case_result(
-        cls, passed_test_case_result: PassedTestCaseResult, fuzz_result: FuzzResult
+    def from_test_execution_result(
+        cls,
+        test_execution_result: TestExecutionResult,
+        **kwargs,
```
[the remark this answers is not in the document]
mkaput: Indeed, but Pyright forced me to do so with the `incompatibleMethodOverride` rule :( https://github.com/software-mansion/protostar/pull/658/commits/8ba19c102fe5b6de0c6c45cd921a114a30f8add2

mkaput:
> IMO this code introduce bad design in the Protostar's "core" where the code should be pristine. A lot of code seems forced:
> `TestMode`, `if isinstance`, `TEST_CASE_RESULT_TYPES`, unlogical function name, and operating on classes as first-class citizens.

`TestMode` is a requirement in order to get setup cases working -- setup case will be executed before test execution environment selection and it will possibly change the mode from STANDARD to FUZZ.

The rest is debatable and I also don't like these but couldn't find anything better.

protostar/commands/test/test_results.py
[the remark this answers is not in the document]
kasperski95: It's a design issue. Pyright merely gives feedback.

kasperski95 (approved):
Discussed privately, I'm leaving the decision whether to merge it or not to you.

I'd do this in the following way:
There should be `SetupCaseExecutionEnvironement` which would pass the execution time and other info to TestCaseExecutionEnvironment.`SetupCaseExecutionEnvironement` could throw `FailedTestCaseResult` to prevent broken test suite error. `TestCaseRunner` utilities functions could be extracted to a separate class and injected to both environments.

mkaput:
Discussed the topic with @kasperski95 more and we have decided to redo the approach once again. This PR doesn't make sense anymore hende I'm closing it.

PR #661, opened 2022-08-24
`create_protostar_project` fixture

kasperski95:
- added `create_protostar_project`
- refactored `protostar format` integration tests
---
- added deploy command to `ProtostarFixture` and used it in `tests/integration/deploying`


tests/integration/conftest.py
```diff
@@ -163,7 +165,35 @@ def protostar_fixture(
     session_mocker: MockerFixture,
     protostar_project_root_path: Path,
 ) -> ProtostarFixture:
+    """@deprecated: Use `create_protostar_project` fixture instead."""
```
Arcticae, line 168: why not replace all usages right now?

tests/integration/conftest.py
```diff
@@ -163,7 +165,35 @@ def protostar_fixture(
...
     return build_protostar_fixture(
         mocker=session_mocker,
         project_root_path=protostar_project_root_path,
     )
+
+
+class CreateProtostarProjectFixture(Protocol):
```
Arcticae, line 175: I don't understand how this is helping us, is that something that we really need?  Typing the fixtures seems a little excessive

tests/integration/conftest.py
```diff
@@ -163,7 +165,35 @@ def protostar_fixture(
...
+
+ProtostarProjectFixture = ProtostarFixture
+
+
+@pytest.fixture(name="create_protostar_project", scope="module")
+def create_protostar_project_fixture(
+    session_mocker: MockerFixture,
+):
+    @contextmanager
+    def create_protostar_project(project_root_path: Path):
```
Arcticae, line 188: This function returns protostar fixture instance, but you're naming the variables `protostar_project`, this seems misleading to what this really returns

Arcticae (changes requested):

tests/integration/conftest.py
mkaput: it helps IDE autocompletion

mkaput (changes requested):
To my eyes, usage of this fixture looks like still invocations of protostar and personally I would drop the `project` word from the name, i.e. replace protostar_fixture with this. Is something blocking such move?

tests/integration/conftest.py
kasperski95: And refactoring, and gives immediate feedback in case of misuse, and [insert type checking benefits]

kasperski95:
> To my eyes, usage of this fixture looks like still invocations of protostar and personally I would drop the project word from the name, i.e. replace protostar_fixture with this. Is something blocking such move?

- conflict with the `protostar` fixture (which I'll replace in this PR)
- `ProtostarFixture` depends on `project root path` which binds the instance of protostar to project. At the other hand, the `ProtostarCLI` is also bound to a project after instantiation, so I'll drop the `project` word to be consistent

tests/integration/conftest.py
kasperski95: Replaced

tests/integration/conftest.py
kasperski95: Removed the concept of `ProtostarProject` but left the name of this function, because `create_protostar` would be more misleading.

mkaput (approved):

tests/integration/conftest.py
```diff
@@ -152,20 +154,31 @@ async def run_cairo_test_runner(
...
+    tmp_path_factory: TempPathFactory,
     signing_credentials: Tuple[str, str],
-) -> ProtostarFixture:
-    return build_protostar_fixture(
-        mocker=session_mocker,
-        project_root_path=protostar_project_root_path,
-        signing_credentials=signing_credentials,
-    )
+):
+    @contextmanager
```
Arcticae, line 168: does this have to be context manager? From what i know, we're using i.e. devnet in similar manner and it doesn't have to be used with `with` keyword

tests/integration/deploying/deploy_command_test.py
```diff
@@ -1,34 +1,32 @@
...
+from tests.integration.conftest import CreateProtostarProjectFixture
+from tests.integration.protostar_fixture import ProtostarFixture
+
+
+@pytest.fixture(name="protostar", scope="module")
+def protostar_fixture(create_protostar_project: CreateProtostarProjectFixture):
+    with create_protostar_project() as protostar:
+        protostar.create_files({"./src/main.cairo": CONTRACT_WITH_CONSTRUCTOR})
+        protostar.build_sync()
+        yield protostar
```
Arcticae, line 15: this is duplicated code

Arcticae: can we make a few of those, just with different scopes ?

Arcticae: this would make things more clear and re-usable

Arcticae (changes requested):

tests/integration/deploying/deploy_command_test.py
kasperski95: > this would make things more clear and re-usable

IMO the opposite would happen.
- This approach allows modifying project files before building the project, and such modification usually needs to be done per test suite / test case.
- The test suite would be harder to understand because it would operate on an environment defined in a different file or files.

Arcticae: >  would operate on an environment defined in a different file or files.
that's usually how our fixtures work, and in terms of reusability it's totally fine. if anything, it makes code cleaner because we don't need to repeat ourselves. 
> such modification usually needs to be done per test suite / test case.
That's usually only adding some test files, or a tree structure right? We could parametrize the fixture with some predefined sources which would be copied to the tree.

tests/integration/conftest.py
kasperski95: Unlike the devnet fixture, the side effect that needs to be reverted is caused when the function is called. There's no point of cleaning up if the function is never called.

Arcticae: got it

tests/integration/deploying/deploy_command_test.py
Arcticae: i can see this kind of code littering our tests all over - that's what i'm afraid of

kasperski95: > that's usually how our fixtures work, and in terms of reusability it's totally fine. if anything, it makes code cleaner because we don't need to repeat ourselves.

It makes the code cleaner visually, but much more obscure logically and harder to maintain.

- colocation https://kentcdodds.com/blog/colocation#the-principle
- the principle of proximity https://kula.blog/posts/proximity_principle/

> That's usually only adding some test files, or a tree structure right? We could parametrize the fixture with some predefined sources which would be copied to the tree.

Not always. Sometimes there's a need to build the project after adding files (migration tests), sometimes there's no such need (`formatter_test.py`). You would have to parametrize the fixture per test case, which would cause duplication (consider `migrator_deploy_contract_cheatcode_test.py`).

Arcticae: This approach won't scale well that's all im saying

Arcticae:
btw i just had a thought about `create_protostar_project` - is there a case where we don't actually create the project? (don't call the function). Each time we want the cleanup logic, so having it as a context manager is kind of weird

kasperski95:
> btw i just had a thought about create_protostar_project - is there a case where we don't actually create the project? (don't call the function). Each time we want the cleanup logic, so having it as a context manager is kind of weird

There's no such case. However, we need to control project scope (function/module). I could add and use two fixtures `protostar_project: ProtostarFixture` and `protostar_project_module: ProtostarFixture`.

@mkaput @Arcticae WDYT?

mkaput:
> > btw i just had a thought about create_protostar_project - is there a case where we don't actually create the project? (don't call the function). Each time we want the cleanup logic, so having it as a context manager is kind of weird
> 
> There's no such case. However, we need to control project scope (function/module). I could add and use two fixtures `protostar_project: ProtostarFixture` and `protostar_project_module: ProtostarFixture`.
> 
> @mkaput @Arcticae WDYT?

I don't like the fact that some tests will use one name while others another.

kasperski95:
> I don't like the fact that some tests will use one name while others another.

So we have to use factory fixture (`create_protostar_project`) as it is now, similarly to `tmp_path_factory`.

The question is where to put the cleanup code. In the fixture or in the `contextmanager`. IMO `create_protostar_project` creates mess (changing cwd) so it should be responsible for cleaning this up.

mkaput:
Yeah, at this point I also lean towards using context managers here.

tests/integration/deploying/deploy_command_test.py
kasperski95: I'd say the opposite. It gives flexibility to adapt environment for test suite needs. Without that, test cases would have to adapt to the environment, which would result in lower quality tests. Fixture environments would become very rigid, because any change could break tests.

tests/integration/protostar_fixture.py
```diff
@@ -10,30 +13,20 @@
...
-    MigrateCommand,
-)
-from protostar.commands.init.project_creator.new_project_creator import (
-    NewProjectCreator,
-)
+from protostar.commands import (BuildCommand, DeclareCommand, FormatCommand,
+                                InitCommand, MigrateCommand)
+from protostar.commands.deploy.deploy_command import DeployCommand
+from protostar.commands.init.project_creator.new_project_creator import \
+    NewProjectCreator
```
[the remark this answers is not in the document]
kasperski95: done and configured `isort` to use black profile

pyproject.toml
```diff
@@ -115,6 +115,8 @@ target-version = [
 
 [tool.isort]
 lines_between_sections = 1
+profile = "black"
```
mkaput, line 118: could this change be cherry picked to separate PR? :)

Arcticae (approved):

mkaput (approved):

PR #664, opened 2022-08-25
Added wrapping starknet_py.TransactionFailedError in ProtostarException

Radinyn:


protostar/starknet_gateway/gateway_facade.py
```diff
@@ -94,15 +95,18 @@ async def deploy(
...
+            register_response(dataclasses.asdict(result))
+            if wait_for_acceptance:
+                if self._logger:
+                    self._logger.info("Waiting for acceptance...")
+                _, status = await self._gateway_client.wait_for_tx(
+                    result.transaction_hash, wait_for_accept=wait_for_acceptance
+                )
+                result.code = status
+        except TransactionFailedError as ex:
+            raise TransactionException(str(ex)) from None
```
mkaput, line 109:
```suggestion
            raise TransactionException(str(ex)) from ex
```

etc.

`from None` part has no sense to me and I would rather keep causality chain

mkaput (changes requested):

protostar/starknet_gateway/gateway_facade.py
Radinyn: done

mkaput (approved):

protostar/starknet_gateway/gateway_facade.py
```diff
@@ -168,15 +172,18 @@ async def declare(
...
+            register_response(dataclasses.asdict(result))
+            if wait_for_acceptance:
+                if self._logger:
+                    self._logger.info("Waiting for acceptance...")
+                _, status = await self._gateway_client.wait_for_tx(
+                    result.transaction_hash, wait_for_accept=wait_for_acceptance
+                )
+                result.code = status
+        except TransactionFailedError as ex:
+            raise TransactionException(str(ex)) from ex
```
Arcticae, line 186: Shouldn't that be ex.message?

mkaput: `str(ex)` decorates `ex.message` with information about exception kind

kasperski95 (approved):

PR #668, opened 2022-08-26
Add `ReportedException.get_metadata_by_type`

mkaput:
Small change which increases code robustness in case new metadata types are added.

### DIFF
--- a/protostar/commands/test/test_environment_exceptions.py
+++ b/protostar/commands/test/test_environment_exceptions.py
@@ -28,6 +28,10 @@ TMetadata = TypeVar("TMetadata", bound=ExceptionMetadata)
 class ReportedException(BaseException):
     """
     The exception used for catching unexpected errors thrown from test cases and as a base class.
+
+    Contract:
+        It is illegal to attach many same-type instances of ExceptionMetadata to the same exception
+        object.
     """
 
     def __init__(self, *args: object) -> None:
--- a/protostar/commands/test/test_environment_exceptions.py
+++ b/protostar/commands/test/test_environment_exceptions.py
@@ -50,7 +50,6 @@ class ReportedException(BaseException):
     def get_metadata_by_type(
         self, metadata_type: Type[TMetadata]
     ) -> Optional[TMetadata]:
-        # Note: Typing this function is hard, hence omitted.
         for metadata in self.metadata:
             if isinstance(metadata, metadata_type):
                 return metadata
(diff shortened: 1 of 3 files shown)


protostar/commands/test/test_environment_exceptions_test.py
```diff
@@ -102,6 +102,24 @@ def test_matching_by_partial_error_message():
...
+        def format(self) -> str:
+            return "unused"
+
+    ex = SimpleReportedException("foo")
+
+    metadata = MockMetadata()
+    ex.metadata.append(metadata)
+
+    assert ex.get_metadata_by_type(MockMetadata) is metadata
+    assert ex.get_metadata_by_type(UnusedMockMetadata) is None
```
kasperski95, line 120: What should happen, when somebody tries to add the same metadata twice?

kasperski95 (approved):

protostar/commands/test/test_environment_exceptions_test.py
mkaput: It's an undocumented programmer error. I will document that the same metadata type can only be added once.

mkaput: Done.

protostar/commands/test/test_environment_exceptions.py
```diff
@@ -22,9 +22,16 @@ def format(self) -> str:
...
+TMetadata = TypeVar("TMetadata", bound=ExceptionMetadata)
+
+
 class ReportedException(BaseException):
     """
     The exception used for catching unexpected errors thrown from test cases and as a base class.
+
+    Contract:
+        It is illegal to attach many same-type instances of ExceptionMetadata to the same exception
+        object.
```
kasperski95, line 34: Maybe making `metadata_type` private and creating a setter with an assertion checking this restriction would be a better approach.

---

#### Hybrid Objects (Objects vs Data Structures)

> This confusion sometimes leads to unfortunate hybrid data structures that are half object and half data structure. They have functions that do significant things, and they also have either public variables or public accessors and mutators that, for all intents and purposes, make the private variables public, tempting other external functions to use those variables the way a procedural program would use a data structure.

> Such hybrids make it hard to add new functions but also make it hard to add new data structures. They are the worst of both worlds. Avoid creating them. They are indicative of a muddled design whose authors are unsure of - or worse, ignorant of - whether they need protection from functions or types.

kasperski95 (approved):

protostar/commands/test/test_environment_exceptions.py
mkaput: It's not that simple. I would have to make the `metadata` field private, expose a read only view of this list as a `metadata` property, and add a new method `attach_metadata` which performs the check. A topic for separate PR nonetheless, question: is it worth it?

PR #669, opened 2022-08-26
Fork test execution state in TestRunner

mkaput:
This change allows calling future setup case running logic from TestRunner.

### DIFF
--- a/protostar/commands/test/test_runner.py
+++ b/protostar/commands/test/test_runner.py
@@ -17,14 +17,15 @@ from protostar.commands.test.starkware.test_execution_state import TestExecution
 from protostar.commands.test.test_case_runners.test_case_runner_factory import (
     TestCaseRunnerFactory,
 )
-from protostar.commands.test.test_config import TestConfig
+from protostar.commands.test.test_config import TestConfig, TestMode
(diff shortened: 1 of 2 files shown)


protostar/commands/test/test_runner.py
```diff
@@ -179,7 +180,20 @@ async def _invoke_test_cases(
...
             self.shared_tests_state.put_result(test_result)
+
+    @staticmethod
+    async def _invoke_test_case(
+        test_case: TestCase, initial_state: TestExecutionState
+    ) -> TestResult:
+        state: TestExecutionState = initial_state.fork()
+
+        # TODO(mkaput): Remove this in favor of setting mode explicitly by cheatcodes in setup hooks.
+        state.config.mode = TestMode.infer_from_contract_function(
```
kasperski95, line 193:
```suggestion
        state.test_case_config.mode = TestMode.infer_from_contract_function(
```
`state.config` could be interpreted as some kind of `StateConfiguration`

kasperski95 (approved):

protostar/commands/test/test_runner.py
mkaput: I will do this in separate PR

PR #672, opened 2022-08-26
ToPythonTransformer + DataTransformer unit tests

Radinyn:


protostar/utils/data_transformer.py
```diff
@@ -9,35 +9,76 @@
...
+            raise DataTransformerException(str(ex)) from ex
+
+    return transform
+
+
+def to_python_events_transformer(
+    contract_abi: AbiType, event_name: str
+) -> ToPythonTransformer:
+    structure_transformer = CairoSerializer(identifier_manager_from_abi(contract_abi))
+    event_abi_item = find_abi_item(contract_abi, event_name)
```
mkaput, line 74:
```suggestion
    event_abi_item = find_abi_item(contract_abi, event_name)
    structure_transformer = CairoSerializer(identifier_manager_from_abi(contract_abi))
```

to be aligned with `to_python_transformer` impl. and please do the same in `from_python_transformer`

protostar/utils/data_transformer_test.py
```diff
@@ -0,0 +1,229 @@
...
+        "type": "event",
+        "from_address": "0",
+    },
+    {
+        "data": [{"name": "input", "type": "(felt, felt)"}],
+        "name": "tuple_event",
+        "type": "event",
+        "from_address": "0",
+    },
+]
```
kasperski95, line 91: Can you create a test utility function that generates the ABI? It would make the test suite more readable and maintainable.

 e.g.
```python
abi = create_abi(
"""\
@constructor
func 

# ...
"""
)
```

Also, you split ABI/PYTHON_INPUT/CAIRO_INPUT into triplets grouped by the input type.
arg, felt_arg, uint256, etc.

mkaput (commented):
@Radinyn could you please outline to me how are you planning to use this for input validation? I know that you will depend on the exceptions being raised here, but I would love to see the incoming code :)

protostar/utils/data_transformer_test.py
```diff
@@ -0,0 +1,229 @@
...
+    [
+        ["constructor", "inputs", VALID_CONSTRUCTOR_PYTHON_INPUT],
+        ["increase_balance", "outputs", {}],
+        ["get_balance", "outputs", {"res": 0x42}],
+    ],
+)
+def test_from_python_pass(
+    name: str, mode: Literal["inputs", "outputs"], data: PythonData
+):
+    from_python_transformer(TEST_ABI, name, mode)(data)
```
kasperski95, line 156: You don't check the output of the transformer

Consider:
```python
cairo_input = from_python(python_input)
assert_cairo_input_is_list(cairo_input)
python_input_2 = to_python(cairo_input)
assert python_input == python_input_2
```

protostar/utils/data_transformer_test.py
```diff
@@ -0,0 +1,229 @@
+from typing_extensions import Literal
```
kasperski95, line 1: I'd say this is an integration test because it relies on Starknet.py

protostar/utils/data_transformer.py
Radinyn: DataTransformer is going to raise an exception when output is invalid which will be later caught by the validator. I know using a transformer for validation isn't particularly clean but I don't think we have much options if we want to delegate it to starknet.py. Nevertheless afaik we wanted to add this functionality to the DataTransformer either way.

mkaput: OK, would it make sense to do wrapping of ValueError/TypeError exceptions into DataTransformerException in validation code instead of here? In order to make such shenanigans more use-case specific.

Radinyn: I don't think we would ever want Protostar to crash on DataTransformer fail, this way we can finish cleanly.

mkaput: hmm, right. please reorder statements then and resolve this discussion :)

protostar/utils/data_transformer_test.py
mkaput: But we are not testing integration of our code swatches; instead we test single functions which uses Starknet.py as an internal dependency. 🤔

Radinyn: I've decided to go with the integration tests approach, because it allows us to leverage ProtostarFixture.

protostar/utils/data_transformer_test.py
Radinyn: done

protostar/utils/data_transformer_test.py
Radinyn: done

tests/data/contracts.py
```diff
@@ -59,3 +59,76 @@
...
+
+DATA_TRANSFORMER_OUTPUT_FELT_CONTRACT = """%lang starknet
+
+@external
+func output_felt() -> (res: felt):
+    return (42)
+end
+"""
+
+DATA_TRANSFORMER_EVENT_CONTRACT = """%lang starknet
```
kasperski95, line 129: Consider making names more generic, or moving contracts to the data transformer test suite.

tests/integration/conftest.py
```diff
@@ -182,3 +184,17 @@ def create_protostar_project():
...
 
     return create_protostar_project
+
+
+@pytest.fixture(name="get_abi_from_contract", scope="module")
+def get_abi_from_contract_fixture(create_protostar_project) -> Callable[[str], AbiType]:
+    def get_abi_from_contract(contract_source_code: str) -> AbiType:
+        with create_protostar_project() as protostar:
+            with open("src/main.cairo", "w") as f:
+                f.write(contract_source_code)
```
kasperski95, line 194:
```suggestion
            protostar.create_files({"src/main.cairo", contract_source_code})
```

tests/integration/data_transformer/data_transformer_test.py
```diff
@@ -0,0 +1,379 @@
...
+        #     "inputs",
+        #     ["very very very very long string"],
+        # ],
+        # AFTER STARKNET.py ADDS EXCEPTION ON TOO MANY VALUES
+        # [
+        #     DATA_TRANSFORMER_FELT_CONSTRUCTOR_CONTRACT,
+        #     "constructor",
+        #     "inputs",
+        #     [0x2342, 0x24],
+        # ],
```
kasperski95, line 259: - Can you add links to Starknet.py issues?
- Consider using `@pytest.mark.skip` (check `gateway_facade_test.py::test_call_to_with_incorrect_args`). 
- WHY DO YOU WRITE COMMENTS IN **UPPERCASE**?

tests/integration/data_transformer/data_transformer_contract.cairo
```diff
@@ -0,0 +1,124 @@
+
```
kasperski95: Is this file still needed?

tests/data/contracts.py
Radinyn: moved

tests/integration/conftest.py
Radinyn: done

tests/integration/data_transformer/data_transformer_test.py
Radinyn: done

kasperski95 (approved):

tests/integration/data_transformer/contracts.py
```diff
@@ -0,0 +1,72 @@
+DATA_TRANSFORMER_UINT256_CONTRACT = """%lang starknet
```
mkaput, line 1:
```suggestion
DATA_TRANSFORMER_UINT256_CONTRACT = """
%lang starknet
```

does this work? if not, can you put `.strip()` there? or `"""\`

mkaput (approved):

PR #673, opened 2022-08-26
Invoke setup cases

mkaput:


protostar/commands/test/environments/setup_case_execution_environment.py
```diff
@@ -0,0 +1,22 @@
...
+from protostar.commands.test.environments.setup_execution_environment import (
+    SetupExecutionEnvironment,
+    SetupCheatcodeFactory,
+)
+from protostar.commands.test.test_context import TestContextHintLocal
+
+
+class SetupCaseExecutionEnvironment(SetupExecutionEnvironment):
+    async def invoke(self, function_name: str):
+        self.set_cheatcodes(SetupCaseCheatcodeFactory(self.state))
```
kasperski95, line 11:
```suggestion
    async def invoke(self, function_name: str):
        self.set_cheatcode_factory(self._setup_case_cheatcode_factory)
```
- [ ] If you inject the factory, dependencies can be injected directly to that factory instead of going through the ExecutionEnvironment.

protostar/commands/test/test_case_runners/setup_case_runner.py
```diff
@@ -0,0 +1,39 @@
...
+    SetupCaseResult,
+    PassedSetupCaseResult,
+    FailedSetupCaseResult,
+)
+from protostar.commands.test.test_suite import TestCase
+
+
+async def run_setup_case(
+    test_case: TestCase, state: TestExecutionState
+) -> SetupCaseResult:
```
kasperski95, line 16: [INFO] I'll probably make this function a class to inject `SetupCaseExecutionEnvironment` at some point in the future. 

My vision on the testing module:
- `TestRunner.worker` instantiates `TestRunnerBuilder/Factory`
- `TestRunnerBuilder/Factory` acts as a `composition_root`
- The core testing logic is moved out of `TestCommand` to `protostar.test_runner` component

protostar/commands/test/test_results.py
```diff
@@ -78,6 +78,23 @@ def from_failed_test_case_result(
         )
 
 
+@dataclass(frozen=True)
+class SetupCaseResult(TestResult, TimedTestResult):
```
kasperski95, line 82: - [ ] `SetupCaseResult` sounds weird. Wouldn't explicit form (`SetupTestCaseResult`) be better?

tests/integration/testing_hooks/setup_case_test.cairo
```diff
@@ -0,0 +1,25 @@
...
+
+@external
+func test_setup_hooks{
+    syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr
+}():
+    %{ assert context.setup_hook_executed == "yes" %}
+    %{ assert context.setup_case_executed == "yes" %}
+
+    return ()
+end
```
kasperski95, line 25: - [ ] it'd be nice to add a test case that checks if `setup_case_executed` is unknown.

protostar/commands/test/environments/setup_case_execution_environment.py
mkaput: Cheatcode factories do not have uniform constructor interfaces. Additionally, in `FuzzTestExecutionStrategy` I am building the factory with a `strategy selector` object which is local to the particular `invoke` call. (Though this part is probably refactorable when we'll move `given` cheatcode to setup).

But in general, I would love to move factory construction to a separate method.

protostar/commands/test/test_case_runners/setup_case_runner.py
mkaput: Beware that `TestExecutionState` is sensitive to forks, and it will have to be passed carefully.

tests/integration/testing_hooks/setup_case_test.cairo
mkaput: Done.

protostar/commands/test/test_results.py
mkaput: I'm not really keen to rename this to `SetupTestCaseResult` because it sounds broken English to me. This can be named `TestCaseSetupResult`, but in general this contradicts the `setup case` terminology I use everywhere.

I'm open to renaming the concept of `setup case` to something else, like `test case setup hook` (though this one is loong). Just need ideas :)

protostar/commands/test/environments/setup_case_execution_environment.py
kasperski95: > Cheatcode factories do not have uniform constructor interfaces.

I don't see a problem with that.

> Additionally, in FuzzTestExecutionStrategy I am building the factory with a strategy selector object which is local to the particular invoke call. 

I'm confused about this part, might come back to it in the future.

protostar/commands/test/test_results.py
kasperski95: `SetupTestResult`?

Anyway, let's stick with the setup case as long as it used consistently.

Arcticae (approved):

protostar/commands/test/test_results.py
```diff
@@ -78,6 +78,23 @@ def from_failed_test_case_result(
...
+    test_case_name: str
+    setup_case_name: str
+
+
+@dataclass(frozen=True)
+class PassedSetupCaseResult(SetupCaseResult):
+    pass
+
+
+# Note: We rely on the fact that this class inherits from FailedTestCaseResult in formatting code.
```
MaksymilianDemitraszek, line 92: This seems a bit weird to me. We provide a distiction and then we tie it up in some confusing way.
Can't we just add?
```
    if isinstance(test_result, FailedSetupCaseResult):
        return _format_failed_test_case_result(test_result)
```
Is there a reason which prevents us from doing that?

MaksymilianDemitraszek (approved):

protostar/commands/test/test_results.py
mkaput: Done

PR #681, opened 2022-08-26
Set fuzz max examples in setup case instead of CLI argument

### LINKED ISSUE #630: Set fuzz max examples in setup case instead of CLI argument

mkaput:
fix #630

depends on #673 (tests will fail until rebased)


protostar/commands/test/cheatcodes/max_examples_cheatcode.py
```diff
@@ -0,0 +1,30 @@
...
+    def name(self) -> str:
+        return "max_examples"
+
+    def build(self) -> Callable[[int], None]:
+        return self.max_examples
+
+    def max_examples(self, max_examples: int) -> None:
+        if max_examples <= 0:
+            raise CheatcodeException(
+                "max_examples", "Max examples value must greater than 0."
```
kasperski95, line 27:
```suggestion
                self, "Max examples value must greater than 0."
```

kasperski95 (approved):

Arcticae (approved):

protostar/commands/test/cheatcodes/max_examples_cheatcode.py
```diff
@@ -0,0 +1,28 @@
...
+        self,
+        syscall_dependencies: Cheatcode.SyscallDependencies,
+        test_config: TestConfig,
+    ):
+        super().__init__(syscall_dependencies)
+        self.test_config = test_config
+
+    @property
+    def name(self) -> str:
+        return "max_examples"
```
MaksymilianDemitraszek, line 19:
```suggestion
        return "max_fuzz_examples"
```
Not sure, maybe would be easier to interpret the purpose

tests/integration/fuzzing_strategies/multiple_learning_steps_test.cairo
```diff
@@ -2,6 +2,12 @@
 
 from starkware.cairo.common.math import assert_le
 
+@external
+func __setup__():
+    %{ max_examples(5) %}
```
MaksymilianDemitraszek, line 7: Can we add tests:
```
    %{ max_examples(-1) %}
```
```
    %{ max_examples(0) %}
```

protostar/commands/test/cheatcodes/max_examples_cheatcode.py
mkaput: I have mixed feelings about this -- originally I wanted to reserve space here for extending usage of this cheatcode in some future, unknown nature, test modes. This cheatcode also does work in standard mode -- it's essentially a no-op there, but it works; so that you can slap `max_examples` in `__setup__` in a suite with mixed standard and fuzz tests.

tests/integration/fuzzing_strategies/multiple_learning_steps_test.cairo
mkaput: Done

MaksymilianDemitraszek (approved):

PR #694, opened 2022-08-30
Fix activity indicator when running in CI

kasperski95:


protostar/cli/activity_indicator.py
```diff
@@ -36,3 +40,7 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_value, exc_tb):
         self.stop()
+
+
+def is_terminal() -> bool:
+    return StreamWrapper(sys.stdout, sys.stdin).isatty()
```
mkaput, line 46: duplicates code from binary_entrypoint. is this desired?

mkaput (approved):

protostar/cli/activity_indicator.py
kasperski95: It is. I tried to avoid creating cyclic dependencies and preserve responsiveness of the initial activity indicator.

mkaput: 👍🏻 would be nice to document this but this is mergeable

PR #702, opened 2022-08-30
`invoke` interface with config

Arcticae:
Moreover, remove InvokeResult from the call (not needed)


protostar/migrator/cheatcodes/migrator_invoke_cheatcode.py
```diff
@@ -14,15 +20,39 @@
...
+class SignedCheatcodeConfig(CheatcodeNetworkConfig):
+    max_fee: NotRequired[int]  # In Wei
+    auto_estimate_fee: NotRequired[bool]
+
+
+@dataclass
+class ValidatedSignedCheatcodeConfig:
+    max_fee: Optional[int] = None
+    wait_for_acceptance: bool = False
+    auto_estimate_fee: bool = False
```
mkaput, line 32:
```suggestion
Wei = int

class SignedCheatcodeConfig(CheatcodeNetworkConfig):
    max_fee: NotRequired[Wei]
    auto_estimate_fee: NotRequired[bool]


@dataclass
class ValidatedSignedCheatcodeConfig:
    max_fee: Optional[Wei] = None
    wait_for_acceptance: bool = False
    auto_estimate_fee: bool = False
```

protostar/migrator/cheatcodes/migrator_invoke_cheatcode.py
```diff
@@ -14,15 +20,39 @@
...
+    def from_dict(
+        cls, config: Optional[SignedCheatcodeConfig]
+    ) -> "ValidatedSignedCheatcodeConfig":
+        if not config:
+            return cls()
+        return cls(
+            wait_for_acceptance=config.get("wait_for_acceptance", False),
+            max_fee=config.get("max_fee", None),
+            auto_estimate_fee=config.get("auto_estimate_fee", False),
+        )
```
mkaput, line 44: Maybe we should also typecheck the values here and raise errors eagerly? Wdyt?

Arcticae: That's not necessary, since have type safety from typeddict. End user error is possible, but type-check step in CI won't pass then.

mkaput: oh didn't notice that. great!

mkaput (approved):

protostar/migrator/cheatcodes/migrator_invoke_cheatcode.py
```diff
@@ -14,15 +20,41 @@
...
         self,
         contract_address: int,
         function_name: str,
-        inputs: Optional[CairoOrPythonData] = None,
-        max_fee: Optional[int] = None,
-        auto_estimate_fee: bool = False,
+        inputs: Optional[CairoOrPythonData],
+        *args,
+        config: Optional[SignedCheatcodeConfig],
     ) -> Any:
```
kasperski95, line 58:
```suggestion
    ) -> None:
```

kasperski95 (approved):

Issue #633, opened 2022-08-19
Restore and update strategies docs


kasperski95:
+ fluent API

kasperski95:
Blocked by https://github.com/software-mansion/protostar/issues/625

PR #704, opened 2022-08-30
Rework fuzzing strategies API

### LINKED ISSUE #574: Chop FuzzTestExecutionEnvironment.invoke into small building blocks

### LINKED ISSUE #625: Rework fuzzing strategies API
```cairo
def setup_foo_bar_baz():
    %{
        example(a=0, b=1)
        example(a=100, b=-2)

        given(a = integers(10, 50), b = felts())
    %}
end

def test_foo_bar_baz(a : felt, b : felt):
    ...
end

def setup_only_parametric_and_with_one_param():
    %{
        example(0)
        example(5)

        # Without `given` call, fuzzer will only run in `explicit` phase.
        # That is, only parameterized mode.
    %}
end

def test_only_parametric_and_with_one_param(a : felt):
    ...
end
```

mkaput:
fix #625, fix #574

Documenting this is part of #633


protostar/commands/test/cheatcodes/given_cheatcode.py
```diff
@@ -41,17 +41,14 @@ def given(self, **kwargs: Any) -> None:
...
-
-
-class StrategyLearnedException(BaseException):
-    """
-    An exception raised from the ``given`` cheatcode, indicating that the set of fuzzing strategies
-    has changed (fuzzer _learned_ a new strategy).
+            # Raise an error if user is trying to reassign a strategy descriptor.
+            # While this is not a big problem for us, it is a rare situation that somebody
+            # might want to do this explicitly, and more often this is just a typo bug.
+            if param in self.test_config.fuzz_declared_strategies:
```
MaksymilianDemitraszek, line 47: Possible scenario, not sure if it makes sense:
- I have 5 test cases with params `a`, `b`
- I want to have the same strategy for 4 of them, and different for one
- I use given in setup for 5 of them
- I override it for the one in case hook

mkaput: I have explicitly added the `given` cheatcode only to the `SetupCaseExecutionEnvironment` (used only for `setup_*` cases), not `SetupExecutionEnvironment` (used by both `__setup__` and `setup_*`), in order to prevent such usage patterns.

I find trying to deduplicate strategy definitions here an antipattern because the user is prone to silently break tests by trying to tinker with strategies of other seemingly unrelated tests, defined in `__setup__` out of laziness/overengineering.

kasperski95 (approved):

mkaput:
As discussed in the office, I will skip the failing test and fix it in a separate issue.

PR #706, opened 2022-08-30
Redesigning `protostar.toml` 1 — `ConfigurationFileV1`

kasperski95:
This PR adds `ConfigurationFile` interface and `ConfigurationFileV1` implementation. `ConfigurationFileV2` will be responsible for interacting with the redesigned `protostar.toml`.


protostar/configuration_file/configuration_file.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def get_command_argument(
+        self, command_name: str, argument_name: str, profile_name: Optional[str] = None
+    ) -> Optional[
+        Union[
+            PrimitiveTypesSupportedByConfigurationFile,
+            List[PrimitiveTypesSupportedByConfigurationFile],
+        ]
+    ]:
+        ...
```
mkaput, line 33: wouldn't it be easier to offload the role of filtering to call-site and let this method simply return everything it knows about all commands?

seems like you will end up duplicating filtering in `V2` struct

protostar/configuration_file/configuration_file.py
```diff
@@ -0,0 +1,40 @@
...
+from typing_extensions import Protocol
+
+from protostar.protostar_exception import ProtostarException
+from protostar.utils.protostar_directory import VersionType
+
+PrimitiveTypesSupportedByConfigurationFile = Union[str, int, bool]
+
+
+class ConfigurationFile(Protocol):
+    def get_min_protostar_version(self) -> Optional[VersionType]:
```
mkaput, line 13: `get` -> `fetch` wdyt? so that it won'd look like these calls are cheap

protostar/configuration_file/configuration_file_v1.py
```diff
@@ -0,0 +1,68 @@
...
+from .configuration_file import (
+    ConfigurationFile,
+    ContractNameNotFoundException,
+    PrimitiveTypesSupportedByConfigurationFile,
+)
+
+
+class ConfigurationFileV1(ConfigurationFile):
+    def __init__(
+        self, protostar_toml_reader: ProtostarTOMLReader, project_root_path: Path
```
mkaput, line 16: I think these classes should **own and construct** `ProtostarTOMLReader` class by themselves, so that these could be replaceable by future versions of `ConfigurationFile`.

protostar/configuration_file/configuration_file.py
```diff
@@ -0,0 +1,40 @@
...
+
+from typing_extensions import Protocol
+
+from protostar.protostar_exception import ProtostarException
+from protostar.utils.protostar_directory import VersionType
+
+PrimitiveTypesSupportedByConfigurationFile = Union[str, int, bool]
+
+
+class ConfigurationFile(Protocol):
```
mkaput, line 12: what about writing `protostar.toml` ?

mkaput (changes requested):

protostar/configuration_file/configuration_file.py
kasperski95: Configuration is going to be cached, so it will be cheap.

protostar/configuration_file/configuration_file_v1.py
kasperski95: Disagree. Dependencies (constructor args) aren't part of the ConfigurationFile interface.

protostar/configuration_file/configuration_file.py
kasperski95: I am working on a different branch. I added a method that returns `ConfigurationWriteFriendlyRepresentation` that's going to be used by `ProtostarTOMLWriter`.

protostar/configuration_file/configuration_file.py
kasperski95: > wouldn't it be easier to offload the role of filtering to call-site and let this method simply return everything it knows about all commands?

Nah, the caller (`ArgumentValueFromConfigProvider`) needs only argument value.

The code won't be duplicated in V2 because redesigned protostar.toml won't use `protostar` prefix.

protostar/configuration_file/configuration_file.py
mkaput: then why not using `@property` then? :)

protostar/configuration_file/configuration_file_v1.py
mkaput: My thing is, `ProtostarTOMLReader` is **not** an external dependency of `ConfigurationFile` but internal one.

If we will hypothetically introduce `ProtostarJSONReader` in the future, with an intent of replacing `protostar.toml` with `protostar.json`; it will be natural to introduce `ConfigurationFileV3`, but having the TOML reader as an external dependency passed from outside will make all of this tricky.

Conceptually, nobody outside `ConfigurationFile` should know that `ProtostarTOMLReader` is used internally. It should be a hidden implementation detail.

protostar/configuration_file/configuration_file.py
mkaput: I don't understand.

`ArgumentValueFromConfigProvider` does accept command name and argument name and returns argument value. This entity can easily filter the expected value from a dictionary `dict[tuple[CommandName, ArgumentName], ArgumentValue]`, and I would expect `ConfigurationFile` to simply return such dictionary.

protostar/configuration_file/configuration_file.py
mkaput: > I am working on it in a different branch/PR.

👍🏻 



> ConfigurationFileV1 will be responsible for writing data. ConfigurationFileV1 should communicate with the ProtostarTOMLReader/Writer in format independent way.

See my other comments why I don't like this.

protostar/configuration_file/configuration_file.py
kasperski95: `ConfigurationFile` is meant to be used as an object, not as a data structure. ConfigurationFile will handle converting relative to absolute paths and reduce the tension between configuration file structure and application needs.

protostar/configuration_file/configuration_file_v1.py
kasperski95: (Internal discussion ended without consensus)

I'm sure I can make it work in a nice loosely coupled. I'll probably create big PR though.

protostar/configuration_file/configuration_file.py
kasperski95: This change will complicate the `ConfigurationFile` and `ArgumentValueFromConfigProvider`. `ConfigurationFile` still needs to expose `get_attribute` method.

mkaput (approved):

PR #709, opened 2022-08-31
Added `incompatibleMethodOverride` to pyright

Radinyn:


kasperski95 (approved):

protostar/commands/test/test_environment_exceptions.py
```diff
@@ -158,7 +158,11 @@ def match(self, other: "RevertableException") -> bool:
     def can_pattern_be_found(pattern: str, strings: List[str]) -> bool:
         return any(pattern in string for string in strings)
 
-    def __reduce__(self):
+    def __reduce__(
+        self,
+    ) -> Tuple[
+        Type, Tuple[Union[List[str], Optional[str], Optional[int]], ...], Dict[str, Any]
+    ]:
```
mkaput, line 165:
```suggestion
    def __reduce__(self) -> Any:
```

I would avoid typing this monster

mkaput (approved):

protostar/commands/test/test_environment_exceptions.py
```diff
@@ -1,6 +1,6 @@
 import re
 from abc import ABC, abstractmethod
-from typing import Dict, List, Optional, Union, TypeVar, Type
+from typing import Dict, List, Optional, Union, TypeVar, Type, Tuple, Any
```
Radinyn, line 3:
```suggestion
from typing import Dict, List, Optional, Union, TypeVar, Type, Any
```

PR #710, opened 2022-08-31
Remove "was used" semantics from TestingSeed and simplify it down to a simple int passed via args

mkaput:
The usage of context variables and passing multiprocess shared memory objects in pickle-insensitive object trees occured to be a very risky code and it has exploded in Python 3.9 resulting in Protostar deadlocking.

This PR removes all these patterns from the code, effectively converting the `TestingSeed` class into a plain `int` value passed around via `TestConfig` object.

This PR changes the functionality of Protostar, the seed will be **always** printed out, even if no fuzz tests were executed. I find this a very neglible UX change.

cc #697

I recommend ignoring whitespace changes when browsing diff.


website/docs/tutorials/07-testing/03-fuzzing/reproducing-failures.md
```diff
@@ -6,14 +6,14 @@ As for now, Protostar provides one mean of achieving this.
...
+examples:
 
 ```
 17:15:36 [INFO] Test suites: 1 failed, 1 total            
 17:15:36 [INFO] Tests:       1 failed, 1 total
 17:15:36 [INFO] Seed:        3658430058
 ```
 
-You can then recreate that test suite run with
+You can then recreate fuzzing test suites runs with
```
kasperski95, line 18:
```suggestion
You can then recreate examples with
```

tests/e2e/test_testing.py
```diff
@@ -11,7 +11,7 @@
 def test_basic_contract(protostar):
     result = protostar(["test", "tests"])
     assert "1 passed" in result
-    assert "Seed:" not in result
+    assert "Seed:" in result
```
kasperski95, line 14: Should we display seed if there are no fuzz tests? IMO this assertion should be in a test case that checks fuzzing.

website/docs/tutorials/07-testing/03-fuzzing/reproducing-failures.md
```diff
@@ -6,14 +6,14 @@ As for now, Protostar provides one mean of achieving this.
 
 ## Reproducing a test run with seed
 
-When a test suite with fuzz test is run, Protostar will print out a numeric seed of a pseudo-random
-number generator used as a source of generated examples:
+Protostar prints a numeric seed of a pseudo-random number generator used as a source of generated
```
kasperski95, line 9: Do we have a test checking if the seed changes with every run?

kasperski95 (approved):

tests/e2e/test_testing.py
mkaput: The whole point of this PR is to display the seed if there are no tests 😅 

Previously we did so, but such flag required use of shared memory objects which broke when we started passing these in a bazilion of dataclasses that we have (state, config, worker args etc.). The change in this test reflects the change in behaviour -- the seed will always be displayed now, because I have simply removed that shared memory usage entirely.

There is also an assertion for this in fuzz related e2e tests. It's not changed so it's not visible in PR diff.

website/docs/tutorials/07-testing/03-fuzzing/reproducing-failures.md
mkaput: #712

tests/e2e/test_testing.py
kasperski95: The point of this PR is fixing deadlock. Displaying seed is a side-effect, that's why I don't understand the purpose of this assertion.

mkaput: This assertion test the behaviour that has changed "in the past" and it guards against reverting this behaviour change accidentally.

PR #715, opened 2022-08-31
Redesigning `protostar.toml` 2 — `ConfigurationFileV1Model`

kasperski95:
Saving v1 won't be supported. `ConfigurationFileV1Model` will be used by `ConfigurationFileV2Model.from`. Models are needed to specify what needs to be saved (check project creators).


protostar/configuration_file/configuration_file.py
```diff
@@ -17,19 +15,10 @@
...
-@dataclass(frozen=True)
-class ConfigurationFileData:
-    min_protostar_version: Optional[str]
-    contract_name_to_path_str: Dict[ContractName, str]
-    lib_path_str: Optional[str]
-    command_name_to_config: CommandNameToConfig
-    shared_command_config: CommandConfig
-    profile_name_to_commands_config: Dict[ProfileName, CommandNameToConfig]
-    profile_name_to_shared_command_config: Dict[ProfileName, CommandConfig]
+TConfigurationFileModel = TypeVar("TConfigurationFileModel")
```
mkaput, line 18: Newer versions of pylint will yell here (see https://github.com/software-mansion/protostar/pull/713/commits/a07d020b88af037ef9993603e254473ff86c3b0a). Valid name would be `ConfigurationFileModelT`

protostar/configuration_file/configuration_file.py
```diff
@@ -52,9 +41,12 @@ def get_command_argument(
     ]:
         ...
 
-    def create_configuration_file_data(
+    def create_model(
```
mkaput, line 50:
```suggestion
    def read(
```

given that there is `save` function ?

also from fmt perspective, this could be a one liner, but the comma forces Black to put new line style here

protostar/configuration_file/configuration_file.py
```diff
@@ -32,6 +46,16 @@ def get_command_argument(
...
         ...
 
+    @abstractmethod
+    def create_model(
+        self,
+    ) -> TConfigurationFileModel:
+        ...
+
+    @abstractmethod
+    def save(self, configuration_file_model: TConfigurationFileModel) -> Path:
```
mkaput, line 56: why is this returning file path?

protostar/configuration_file/configuration_file_v1.py
```diff
@@ -66,3 +92,82 @@ def get_command_argument(
...
+    ) -> CommandConfig:
+        return (
+            self._protostar_toml_reader.get_section(
+                "shared_command_configs", profile_name=profile_name
+            )
+            or {}
+        )
+
+    def save(self, configuration_file_model: ConfigurationFileV1Model) -> Path:
+        assert False, "Operation not supported"
```
mkaput, line 173:
```suggestion
        raise NotImplementedError("Operation not supported.")
```

mkaput (changes requested):

protostar/configuration_file/configuration_file.py
kasperski95: It was convenient from the test perspective.

mkaput: Aren't tests able to get this value from elsewhere? It doesn't really make any sense to me to have such contract here.

kasperski95: I think it makes sense because`ConfigurationFile`'s callers don't know the format of the file, hence they can't know the exact path upfront.

mkaput: hmm but now they cannot learn about the file without saving it...

would it make sense to expose the path as a `property` here?

mkaput (approved):

protostar/configuration_file/configuration_file.py
kasperski95: I removed the `save` method from the `ConfigurationFile` because readers don't care about the file format, but writers do.

PR #729, opened 2022-09-01
Redesigning `protostar.toml` 4 — Saving `ConfigurationFileV2`

kasperski95:


protostar/configuration_file/confiuration_toml_writer_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+    toml_writer.save(
+        configuration_model=model,
+        filepath=output_file_path,
+    )
+
+    result = output_file_path.read_text()
+    assert result is not None
+    assert not '["' in result
+    assert not '"]' in result
```
mkaput, line 54: wouldn't it be more fault-proof to verify whole file contents exactly?

protostar/configuration_file/confiuration_toml_writer_test.py
```diff
@@ -0,0 +1,54 @@
+from dataclasses import dataclass
```
mkaput, line 1: I'm missing a test that verifies that `ConfigurationTOMLWriter` writes stuff in fixed order. So that our writes will not cause nondeterministic diffs.

protostar/configuration_file/configuration_file_v2.py
```diff
@@ -89,5 +96,48 @@ def read(
     ) -> ConfigurationFileV2Model:
         raise NotImplementedError("Operation not supported.")
 
-    def save(self, configuration_file_model: ConfigurationFileV2Model) -> Path:
-        raise NotImplementedError("Operation not supported.")
+    def create_file_content(
```
mkaput, line 99: I wonder if it wouldn't be simpler to simply build a Python dict containing all data and pass it to the writer.

In general that would drop the need for the `ConfigurationFileContentBuilder` class.

mkaput (changes requested):
It's not clear to me how to write something like this, with exactly this formatting:

```toml
[dependencies]
foo = "^1.0.0"

[profile.foo.dependencies]
foo = { git = "...", branch = "..." }
```

I know that this is out of scope of this PR because we do not have such feature, but I don't know how to fit it into this design.

kasperski95:
> It's not clear to me how to write something like this, with exactly this formatting:

Added a test case that represents how to do it. I noticed inline tables weren't supported, so I replaced the `toml` package which is going to be useful anyway to add comments.

protostar/configuration_file/confiuration_toml_writer_test.py
kasperski95: `test_saving_configuration` is a snapshot test, and it doesn't fail randomly

protostar/configuration_file/configuration_file_v2.py
mkaput: Shouldn't `profile` en/decoding be shifted to a separate class that would be layered over all of this? You're making too powerful objects here.

protostar/configuration_file/configuration_file_v2_test.py
```diff
@@ -144,3 +138,31 @@ def test_saving_configuration(
 
     result = filepath.read_text()
     assert result == protostar_toml_content
+
+
+def test_saving_in_particular_order(
```
mkaput, line 143: I think 1. building the configuration file ad hoc specifically for this test and 2. testing exact contents would be more robust here and 3. testing only ConfigurationTOMLWriter independent of the model.

mkaput (changes requested):

pyproject.toml
```diff
@@ -26,6 +26,7 @@ tomli = "<2.0.0"
 tomli-w = "^1.0.0"
 tqdm = "^4.64.0"
 typing-extensions = "^4.0.1"
+tomlkit = "^0.11.4"
```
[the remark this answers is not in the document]
kasperski95: Maybe in the future. I'd leave the old library as long as we support V1.

protostar/configuration_file/configuration_file_v2_test.py
kasperski95: > 3. testing only ConfigurationTOMLWriter independent of the model.

I'll remove `ConfigurationTOMLWriter`. `ConfigurationTOMLContentBuilder` is what should be tested.

protostar/configuration_file/configuration_file_v2.py
kasperski95: - there might be different conventions for encoding `profiles` per format 
- `ConfigurationTOMLContentBuilder` has 54 lines of cohesive code
- there are more important things to do

mkaput (changes requested):

poetry.lock
```diff
@@ -54,7 +54,7 @@ optional = false
 python-versions = ">=3.7"
 
 [package.extras]
-tests = ["mypy (>=0.800)", "pytest", "pytest-asyncio"]
+tests = ["pytest", "pytest-asyncio", "mypy (>=0.800)"]
```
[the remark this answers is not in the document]
kasperski95: I believe the most recent one (I need to check it on my second laptop).

protostar/configuration_file/configuration_toml_content_builder.py
```diff
@@ -0,0 +1,64 @@
...
+import tomlkit
+from tomlkit.items import InlineTable, Table
+
+from .configuration_file import ConfigurationFileContentBuilder
+
+ConfigurationTOMLContent = str
+
+
+class ConfigurationTOMLContentBuilder(
+    ConfigurationFileContentBuilder[ConfigurationTOMLContent]
```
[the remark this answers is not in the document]
kasperski95: It's not necessary. It's a leftover after a different flow.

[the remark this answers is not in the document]
kasperski95: Removed generic

protostar/configuration_file/configuration_toml_content_builder_test.py
```diff
@@ -0,0 +1,97 @@
...
+        profile_name="devnet", section_name="declare", data={"network": "devnet"}
+    )
+    content_builder = ConfigurationTOMLContentBuilder()
+
+    result = content_configurator.create_file_content(
+        content_builder=content_builder,
+        model=model,
+    )
+
+    assert "[profile.devnet.declare]" in result
```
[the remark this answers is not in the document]
kasperski95: Done

poetry.lock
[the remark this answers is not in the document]
kasperski95: Reverted changes

mkaput (approved):
LGTM

PR #731, opened 2022-09-02
Added BrokenTestCaseResult

Radinyn:
- added BreakingReportedException, makes the tests finish with the BROKEN status
- renamed TestingSummary.broken -> TestingSummary.broken_suites
- added TestingSummary.broken for BrokenTestCaseResults
- fixed tests

This is a potential breaking change.

### DIFF
--- a/protostar/commands/test/cheatcodes/reflect/cairo_struct.py
+++ b/protostar/commands/test/cheatcodes/reflect/cairo_struct.py
@@ -1,13 +1,15 @@
 from typing import OrderedDict, Union, Any, List, Tuple, Type
 from starkware.cairo.lang.vm.relocatable import RelocatableValue
-from protostar.commands.test.test_environment_exceptions import SimpleReportedException
+from protostar.commands.test.test_environment_exceptions import (
+    SimpleBreakingReportedException,
+)
 from protostar.starknet.hint_local import HintLocal
 
 
 class CairoStruct:
     def __init__(self, *args, **kwargs) -> None:
         if len(args) > 0:  # / syntax not avaliable in py3.7
-            raise SimpleReportedException(
+            raise SimpleBreakingReportedException(
                 "CairoStruct constructor takes only keyword arguments."
             )
 
@@ -15,7 +17,7 @@ class CairoStruct:
 
         for key, value in kwargs.items():
             if not isinstance(value, VALID_CAIRO_TYPES):
-                raise SimpleReportedException(
+                raise SimpleBreakingReportedException(
                     f'"{type(value).__name__}" is not a valid CairoType.'
                 )
             self._ordered_dict[key] = value
@@ -23,16 +25,18 @@ class CairoStruct:
     def __getattr__(self, name: str) -> "CairoValueType":
         if name in self._ordered_dict:
             return self._ordered_dict[name]
-        raise SimpleReportedException(f'"{name}" is not a member of this CairoStruct.')
+        raise SimpleBreakingReportedException(
+            f'"{name}" is not a member of this CairoStruct.'
+        )
 
     def __setattr__(self, name: str, value: Any) -> None:
         if name != "_ordered_dict":
-            raise SimpleReportedException("CairoStruct is immutable.")
+            raise SimpleBreakingReportedException("CairoStruct is immutable.")
         super().__setattr__(name, value)
 
     def _set(self, name: str, value: "CairoValueType") -> None:
         if not isinstance(value, VALID_CAIRO_TYPES):  # type: ignore
-            raise SimpleReportedException(
+            raise SimpleBreakingReportedException(
                 f'"{type(value).__name__}" is not a valid CairoType.'
             )
         self._ordered_dict[name] = value
--- a/protostar/commands/test/cheatcodes/reflect/cairo_struct.py
+++ b/protostar/commands/test/cheatcodes/reflect/cairo_struct.py
@@ -6,10 +6,14 @@ from protostar.commands.test.test_environment_exceptions import (
 from protostar.starknet.hint_local import HintLocal
 
 
+class CairoStructException(SimpleBreakingReportedException):
+    pass
+
+
 class CairoStruct:
     def __init__(self, *args, **kwargs) -> None:
         if len(args) > 0:  # / syntax not avaliable in py3.7
-            raise SimpleBreakingReportedException(
+            raise CairoStructException(
                 "CairoStruct constructor takes only keyword arguments."
             )
 
@@ -17,7 +21,7 @@ class CairoStruct:
 
         for key, value in kwargs.items():
             if not isinstance(value, VALID_CAIRO_TYPES):
-                raise SimpleBreakingReportedException(
+                raise CairoStructException(
                     f'"{type(value).__name__}" is not a valid CairoType.'
                 )
             self._ordered_dict[key] = value
@@ -25,18 +29,16 @@ class CairoStruct:
     def __getattr__(self, name: str) -> "CairoValueType":
         if name in self._ordered_dict:
             return self._ordered_dict[name]
-        raise SimpleBreakingReportedException(
-            f'"{name}" is not a member of this CairoStruct.'
-        )
+        raise CairoStructException(f'"{name}" is not a member of this CairoStruct.')
 
     def __setattr__(self, name: str, value: Any) -> None:
         if name != "_ordered_dict":
-            raise SimpleBreakingReportedException("CairoStruct is immutable.")
+            raise CairoStructException("CairoStruct is immutable.")
         super().__setattr__(name, value)
 
     def _set(self, name: str, value: "CairoValueType") -> None:
         if not isinstance(value, VALID_CAIRO_TYPES):  # type: ignore
-            raise SimpleBreakingReportedException(
+            raise CairoStructException(
                 f'"{type(value).__name__}" is not a valid CairoType.'
             )
         self._ordered_dict[name] = value
--- a/protostar/commands/test/test_case_runners/fuzz_test_case_runner.py
+++ b/protostar/commands/test/test_case_runners/fuzz_test_case_runner.py
@@ -56,13 +56,8 @@ class FuzzTestCaseRunner(TestCaseRunner[FuzzTestExecutionResult]):
             reported_exception, execution_metadata
         )
         fuzz_result = self._map_reported_exception_to_fuzz_result(reported_exception)
-        if fuzz_result:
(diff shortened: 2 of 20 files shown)


protostar/commands/test/test_results.py
```diff
@@ -28,15 +31,25 @@ class TimedTestResult:
 
 
 @dataclass(frozen=True)
-class PassedTestCaseResult(TestCaseResult, TimedTestResult):
+class TimedTestCaseResult(TestCaseResult, TimedTestResult):
+    pass
```
mkaput, line 35: I don't think we need this extra level of abstraction.

Radinyn: It's used to provide typing for the TestingSummary, however now I think that adding TimedTestResult as a parent of TestCaseResult might be a better idea.

mkaput: @kasperski95 wdyt?

kasperski95: It might conflict with the FuzzTestCaseResult. Fuzz test case results could expect `execution_time` to be passed as a statistic.

Radinyn: I've checked, there are no conflicts with fuzzing related results.

kasperski95: Currently, there are no conflicts. At the other hand, time statistic can be unrelated to the total execution time. Inheritance should be ok.

protostar/commands/test/test_case_runners/test_case_runner.py
```diff
@@ -42,7 +46,13 @@ async def run(self) -> TestCaseResult:
                 execution_result,
                 TestCaseRunner.ExecutionMetadata(self._stopwatch.total_elapsed),
             )
+        except BreakingReportedException as ex:
+            return self._map_reported_exception_to_broken_test_result(
+                ex,
+                TestCaseRunner.ExecutionMetadata(self._stopwatch.total_elapsed),
+            )
         except ReportedException as ex:
+            print(f"CAUGHT EXCEPTION: {type(ex)}")
```
kasperski95:
```suggestion
```

protostar/commands/test/test_case_runners/test_case_runner.py
```diff
@@ -42,7 +46,13 @@ async def run(self) -> TestCaseResult:
                 execution_result,
                 TestCaseRunner.ExecutionMetadata(self._stopwatch.total_elapsed),
             )
+        except BreakingReportedException as ex:
+            return self._map_reported_exception_to_broken_test_result(
```
kasperski95, line 50:
```suggestion
            return self._map_breaking_reported_exception_to_broken_test_result(
```

protostar/commands/test/cheatcodes/reflect/cairo_struct.py
```diff
@@ -1,38 +1,42 @@
...
+    SimpleBreakingReportedException,
+)
 from protostar.starknet.hint_local import HintLocal
 
 
 class CairoStruct:
     def __init__(self, *args, **kwargs) -> None:
         if len(args) > 0:  # / syntax not avaliable in py3.7
-            raise SimpleReportedException(
+            raise SimpleBreakingReportedException(
```
mkaput, line 12: seems like nice opportunity to define a `CairoStructError` class

protostar/commands/test/test_case_runners/fuzz_test_case_runner.py
```diff
@@ -61,6 +65,25 @@ def _map_reported_exception_to_failed_test_result(
...
+        fuzz_result = self._map_reported_exception_to_fuzz_result(reported_exception)
+        if fuzz_result:
+            return BrokenFuzzTestCaseResult.from_broken_test_case_result(
+                broken_test_case_result,
+                fuzz_result,
+            )
+        return BrokenFuzzTestCaseResult.from_broken_test_case_result(
+            broken_test_case_result,
+            fuzz_result=None,
+        )
```
mkaput, line 85:
```suggestion
        if fuzz_result:
            fuzz_result = self._map_reported_exception_to_fuzz_result(reported_exception)

        return BrokenFuzzTestCaseResult.from_broken_test_case_result(
            broken_test_case_result,
            fuzz_result,
        )
```

protostar/commands/test/test_case_runners/test_case_runner.py
mkaput, line 55:
```suggestion
```

protostar/commands/test/testing_summary.py
```diff
@@ -183,7 +188,9 @@ def _format_slow_test_cases_list(
         local_log_color_provider: LogColorProvider = log_color_provider,
     ) -> str:
 
-        slowest_test_cases = self._get_slowest_test_cases_list(self.failed + self.passed, count)  # type: ignore
+        slowest_test_cases = self._get_slowest_test_cases_list(
+            self.failed + self.passed + self.broken, count  # type: ignore
```
mkaput, line 192:
```suggestion
            self.case_results, count
```

doesn't this work?

protostar/commands/test/test_results.py
Radinyn: Reverted back to inheritance

mkaput (changes requested):

protostar/commands/test/test_runner.py
```diff
@@ -183,8 +183,8 @@ async def _invoke_test_case(
 
         if test_case.setup_fn_name:
             setup_case_result = await run_setup_case(test_case, state)
-            if isinstance(setup_case_result, FailedSetupCaseResult):
-                return setup_case_result.into_failed_test_case_result()
+            if isinstance(setup_case_result, BrokenSetupCaseResult):
+                return setup_case_result.into_broken_test_case_result()
```
kasperski95, line 187: Off topic:
`into` is a code smell and can always be replaced with `from`. Change my mind.

@mkaput

protostar/commands/test/testing_summary.py
```diff
@@ -73,15 +78,17 @@ def log_slowest_test_cases(
             )
 
     def assert_all_passed(self):
-        if self.failed or self.broken:
+        if self.failed or self.broken_suites:
```
kasperski95, line 81:
```suggestion
        if self.failed or self.broken_suites or self.broken:
```

protostar/commands/test/testing_summary.py
Radinyn: No, because case_results also stores BrokenTestSuites. We might consider changing it though.

protostar/commands/test/testing_summary.py
```diff
@@ -96,7 +103,7 @@ def _get_test_suites_summary(self, collected_test_suites_count: int) -> str:
         for suit_case_results in self.test_suites_mapping.values():
             partial_summary = TestingSummary(suit_case_results, self.testing_seed)
 
-            if len(partial_summary.broken) > 0:
+            if len(partial_summary.broken_suites) + len(partial_summary.broken) > 0:
```
kasperski95, line 106:
```suggestion
            if len(partial_summary.broken_suites) > 0:
```
Consider use case: 10 broken test cases in a one test suite

protostar/commands/test/testing_summary.py
mkaput: I actually wonder why this method accepts a list of test cases as argument, given that it knows all results due to being a method of the summary class. Maybe we should simply remove this argument and filter case_results by isinstance internally?

protostar/commands/test/testing_summary.py
```diff
@@ -169,12 +176,10 @@ def _get_preprocessed_core_testing_summary(
 
     def _get_slowest_test_cases_list(
         self,
-        failed_and_passed_list: List[Union[PassedTestCaseResult, FailedTestCaseResult]],
+        test_case_list: List[TestCaseResult],
```
kasperski95:
```suggestion
        test_case_results: List[TestCaseResult],
```

protostar/commands/test/testing_summary.py
Radinyn: Oh, I think it's a leftover from a time when this method was in a different file.

kasperski95, line 192: I think `type: ignore` can be now removed.

tests/integration/conftest.py
```diff
@@ -56,10 +59,13 @@ def assert_cairo_test_cases(
     )
 
     if expected_broken_test_cases_names is None:
-        actual_broken = len(testing_summary.broken)
+        actual_broken = len(testing_summary.broken_suites) + len(testing_summary.broken)
```
kasperski95, line 62: up

kasperski95 (changes requested):

protostar/commands/test/test_case_runners/fuzz_test_case_runner.py
Radinyn, line 85:
```suggestion
        fuzz_result = self._map_reported_exception_to_fuzz_result(reported_exception)
        return BrokenFuzzTestCaseResult.from_broken_test_case_result(
            broken_test_case_result,
            fuzz_result=fuzz_result or None,
        )
```

tests/integration/conftest.py
kasperski95: Why is `actual_broken` a number here, but 2 lines below is a set?

protostar/commands/test/test_runner.py
mkaput: `from` should be preferred over `into` wherever possible, provided you know the target type

here we use `into` because we don't know the concrete type we convert into

protostar/commands/test/cheatcodes/reflect/cairo_struct.py
```diff
@@ -6,37 +6,39 @@
...
 
+class CairoStructException(SimpleBreakingReportedException):
+    pass
+
+
 class CairoStruct:
     def __init__(self, *args, **kwargs) -> None:
         if len(args) > 0:  # / syntax not avaliable in py3.7
-            raise SimpleBreakingReportedException(
+            raise CairoStructException(
```
mkaput, line 16: assertions in tests are not changed

mkaput (changes requested):

tests/integration/conftest.py
Radinyn: Refactored this function

protostar/commands/test/testing_summary.py
Radinyn: I couldn't find a way to make it work without including some ugly code

kasperski95 (approved):

mkaput (approved):

PR #732, opened 2022-09-02
Added article to README.md

Radinyn:


README.md
```diff
@@ -24,6 +24,8 @@
 ## About
 Protostar manages your dependencies, compiles your project, and runs tests.
 
+Check out [this article](https://blog.swmansion.com/testing-starknet-contracts-made-easy-with-protostar-2ecdad3c9133) to get started.
```
MaksymilianDemitraszek, line 27: Maybe would be better do add `### Additional resources` subsection of documentation? We may add more of them in the future

Radinyn: We could add it when we will have at least two resources

mkaput: I think we would forget to add it in the future 😃 And for example there was a comparison with Nile blog post -- also worth adding here, so @MaksymilianDemitraszek suggestion is totally cool here IMHO

mkaput (approved):

PR #758, opened 2022-09-05
Add cairo-migrator command

Arcticae:


website/docs/cli-reference.md
```diff
@@ -31,6 +31,10 @@ Additional directories to look for sources.
...
...ples/basic_with_constructor.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/protostar/commands/test/examples/fuzzing/test_fuzz.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/protostar/commands/test/examples/broken/test_basic_broken.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/protostar/commands/test/examples/broken/basic_broken.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/protostar/commands/test/examples/empty/test_no_test_functions.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/protostar/commands/test/examples/basic/test_basic.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/protostar/commands/test/examples/invalid/test_invalid_syntax.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/protostar/commands/test/examples/failure/test_basic_failure.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/cairo/protostar/_utils.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/cairo/protostar/asserts.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/templates/default/tests/test_main.cairo'), PosixPath('/Users/timelock/swm/starkware/protostar/templates/default/src/main.cairo')]`
```
mkaput, line 36: 😆

protostar/cli/misc.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
+from typing import List
+
+
+def resolve_cairo_targets(targets: List[str]) -> List[Path]:
```
mkaput, line 5: could we reuse globbing logic used by the test runner here?

protostar/cli/misc.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
```
mkaput, line 1: Personally I don't like this module name and in general I am big opponent of making _utility_ modules.

As last resort, this module could be named the same as the main function it exports, like you would do with React components.

website/docs/cli-reference.md
Arcticae: oh f.. me

protostar/cli/misc.py
Arcticae: i reused the logic from the format command

protostar/cli/misc.py
Arcticae: yeah i thought somebody might provide a better solution than this, so i left it up to reviewers :D

protostar/cli/misc.py
mkaput: OK, that's a great starting point 😄 

do as you wish, you can resolve this comment

mkaput (approved):

protostar/cairo_migrator.py
```diff
@@ -0,0 +1,45 @@
...
+            if self._single_return_functions
+            else None,
+        )
+        new_content = ast.format()
+        return new_content
+
+    def save(self):
+        for filepath, new_content in self._formatted_files:
+            with open(filepath, "w", encoding="utf-8") as file:
+                self._logger.info(f"Writing {filepath}")
```
kasperski95, line 35: Consider using logger only in the `Command` classes. Check `protostar.formatter`.

protostar/cairo_migrator.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import List
+
+from starkware.cairo.lang.compiler.ast.formatting_utils import FormattingError
+from starkware.cairo.lang.migrators.migrator import parse_and_migrate, MIGRATE_FUNCTIONS
+
+from protostar.protostar_exception import ProtostarException
+
+
+class CairoMigrator:
+    def __init__(self, logger: Logger, single_return_functions: bool = True):
```
kasperski95, line 12: Defaults for booleans are `False` in the entire project.

protostar/commands/format/format_command.py
```diff
@@ -88,15 +89,15 @@ async def run(self, args):
...
+        self, targets: List[str], check=False, verbose=False, ignore_broken=False
     ) -> FormattingSummary:
 
         callback: Callable[[FormattingResult], None] = lambda result: print(
             format_formatting_result(result, check)
         )
 
         summary = self._formatter.format(
-            targets=targets,
+            cairo_targets=resolve_cairo_targets(targets),
```
kasperski95, line 100:
```suggestion
            filepaths=resolve_cairo_targets(targets),
```

kasperski95: What's the difference between `target` and `cairo_target`?

protostar/cli/resolve_cairo_targets.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
+from typing import List
+
+
+def resolve_cairo_targets(targets: List[str]) -> List[Path]:
```
kasperski95, line 5:
```suggestion
def map_targets_to_filepaths(targets: List[str]) -> List[Path]:
```

protostar/commands/format/format_command.py
```diff
@@ -36,7 +37,7 @@ def arguments(self) -> List[Command.Argument]:
             Command.Argument(
                 name="target",
                 description=("Target to format, can be a file or a directory."),
-                type="path",
+                type="str",
```
kasperski95, line 40: Maybe new type is needed — `glob`.

protostar/cairo_migrator.py
kasperski95: `single_return_functions` is never used and could be removed from `CairoMigrator`.

protostar/commands/format/format_command.py
Arcticae: meh

Arcticae: maybe when we're going to use it more, now it's overkill

protostar/commands/format/format_command.py
Arcticae: renamed

protostar/cli/resolve_cairo_targets.py
Arcticae: renamed

protostar/cairo_migrator.py
Arcticae: deleted

protostar/cairo_migrator.py
Arcticae: done

protostar/commands/format/format_command.py
kasperski95: You could move `map_targets_to_filepaths` logic there.

Arcticae: yeah i know but it's too much effort (introducing the new type, testing, refactoring) for the payoff right now. i don't think there's going to be much maintenance with what we have now, and we're probably never going to use that type anywhere

kasperski95: A user doesn't know why target uses `str` type instead of `Path` type. It would make the help description and the CLI docs more explicit.

@mkaput wdyt?

kasperski95 (approved):

PR #763, opened 2022-09-06
Hack: Move cheatable state to CachedState

### LINKED ISSUE #700: Port Protostar state cheats to Cairo 0.10 new classes
We don't have to maintain 2 forkable states. Protostar state should be possibly independent of cairo-lang.

### LINKED ISSUE #722: Fix field `loop` removed from BusinessLogicSysCallHandler

mkaput:
fix #700
fix #722

### DIFF
--- a/protostar/starknet/cheatable_cached_state.py
+++ b/protostar/starknet/cheatable_cached_state.py
@@ -1,0 +1,124 @@
+from typing import Dict, List
+
+from starkware.starknet.business_logic.state.state import CachedState
+from starkware.starknet.public.abi import AbiType
+from typing_extensions import Self
+
+from protostar.commands.test.test_environment_exceptions import SimpleReportedException
+from protostar.starknet.types import AddressType, SelectorType, ClassHashType
+
+
+# pylint: disable=too-many-instance-attributes
+class CheatableCachedState(CachedState):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.pranked_contracts_map: Dict[int, int] = {}
+        self.mocked_calls_map: Dict[AddressType, Dict[SelectorType, List[int]]] = {}
+        self.event_selector_to_name_map: Dict[int, str] = {}
+
+        self.event_name_to_contract_abi_map: Dict[str, AbiType] = {}
+        self.class_hash_to_contract_abi_map: Dict[ClassHashType, AbiType] = {}
+        self.contract_address_to_class_hash_map: Dict[AddressType, ClassHashType] = {}
+
+        self.contract_address_to_block_timestamp: Dict[AddressType, int] = {}
+        self.contract_address_to_block_number: Dict[AddressType, int] = {}
+
+    def _copy(self):
+        copied = self.__class__(block_info=self.block_info, state_reader=self)
+
+        copied.pranked_contracts_map = self.pranked_contracts_map.copy()
+        copied.mocked_calls_map = self.mocked_calls_map.copy()
+        copied.event_selector_to_name_map = self.event_selector_to_name_map.copy()
+
+        copied.event_name_to_contract_abi_map = (
+            self.event_name_to_contract_abi_map.copy()
+        )
+        copied.class_hash_to_contract_abi_map = (
+            self.class_hash_to_contract_abi_map.copy()
+        )
+        copied.contract_address_to_class_hash_map = (
+            self.contract_address_to_class_hash_map.copy()
+        )
+
+        copied.contract_address_to_block_timestamp = (
+            self.contract_address_to_block_timestamp.copy()
+        )
+        copied.contract_address_to_block_number = (
+            self.contract_address_to_block_number.copy()
+        )
+        return copied
+
+    def _apply(self, parent: Self):
+        assert isinstance(parent, self.__class__)
+        super()._apply(parent)
+
+        parent.pranked_contracts_map = {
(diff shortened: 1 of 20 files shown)


protostar/starknet/cheatable_execute_entry_point.py
```diff
@@ -127,30 +111,40 @@ def _run(  # type: ignore
...
+                segments=runner.segments, data=self.calldata
+            ),
         ]
 
         try:
             runner.run_from_entrypoint(
                 entry_point.offset,
                 *entry_points_args,
-                hint_locals=hint_locals,
+                # --- MODIFICATIONS START --- # TODO
```
kasperski95, line 132:
```suggestion
                # --- MODIFICATIONS START ---
```

protostar/starknet/cheatable_syscall_handler.py
```diff
@@ -202,36 +179,28 @@ def _deploy(
             constructor_calldata=constructor_calldata,
             deployer_address=deployer_address,
         )
-        # BEGIN: PROTOSTAR_MODIFICATION — update mappings
+
+        # --- MODIFICATIONS START --- # TODO
```
kasperski95, line 183: What's the ponit of `TODO` without a description of what needs to be done?

protostar/starknet/cheatable_syscall_handler.py
```diff
@@ -202,36 +179,28 @@ def _deploy(
...
-        self.deployed_contracts.append(contract_address)
-
         return contract_address
 
 
 class CheatableHintsWhitelist(HintsWhitelist):
-    def verify_hint_secure(self, hint, reference_manager):  # type: ignore
-        return True
+    def verify_hint_secure(self, hint, reference_manager):
+        pass
```
kasperski95, line 206: Did compiler stop complaining about Protostar's cheatcodes?

protostar/starknet/cheatable_syscall_handler.py
```diff
@@ -29,44 +24,48 @@ def __init__(self, message: str):
...
+    #
+    # def _get_block_timestamp(self):
+    #     if (
+    #         self.contract_address
+    #         in self.cheatable_state.contract_address_to_block_timestamp
+    #     ):
+    #         return self.cheatable_state.contract_address_to_block_timestamp[
+    #             self.contract_address
+    #         ]
+    #     return super()._get_block_timestamp()
```
kasperski95, line 54: Old code could be removed. We can always bring it back with the git.

protostar/starknet/forkable_starknet.py
```diff
@@ -62,13 +60,13 @@ async def deploy(
             disable_hint_validation=disable_hint_validation,
         )
 
-        self.cheatable_state.cheatable_carried_state.update_event_selector_to_name_map(
+        self.cheatable_state.cheatable_state.update_event_selector_to_name_map(
```
kasperski95, line 63:
```suggestion
        self.cheatable_starknet_state.cheatable_state.update_event_selector_to_name_map(
```

Using the same name for 2 concepts is a bit confusing.

kasperski95 (approved):

protostar/starknet/cheatable_cached_state.py
```diff
@@ -0,0 +1,124 @@
...
+from starkware.starknet.business_logic.state.state import CachedState
+from starkware.starknet.public.abi import AbiType
+from typing_extensions import Self
+
+from protostar.commands.test.test_environment_exceptions import SimpleReportedException
+from protostar.starknet.types import AddressType, SelectorType, ClassHashType
+
+
+# pylint: disable=too-many-instance-attributes
+class CheatableCachedState(CachedState):
```
kasperski95, line 12: Was moving `CheatableState` to the `TestExecutionState` problematic, or did you decide not to mix refactoring and adding support for Cairo 0.10?

protostar/starknet/cheatable_syscall_handler.py
mkaput: That's the pattern that was used everywhere else in codebase, I have simply unified things here.

protostar/starknet/cheatable_syscall_handler.py
mkaput: Nope, that is caused by the fact that cheatcodes are not injected into VM locals and this is separate issue.

protostar/starknet/cheatable_cached_state.py
mkaput: Both 😅

protostar/starknet/cheatable_syscall_handler.py
mkaput: Replaced this with [`#region syntax`](https://stackoverflow.com/a/20225477)

protostar/starknet/forkable_starknet.py
mkaput: The tricky part is that without `cheatable` this would be `self.state.state` still.

protostar/starknet/cheatable_syscall_handler.py
mkaput: But after 4d1eb4f1fb5036de4bdacb3e5936d0ad59874b2f this is fixed. Done.

protostar/starknet/forkable_starknet.py
kasperski95: `Cheatable` prefix can be preserved.

PR #764, opened 2022-09-06
Use execute instead of invoke for calling Cairo functions

mkaput:


kasperski95 (approved):

protostar/starknet/execution_environment.py
```diff
@@ -21,15 +21,15 @@ def __init__(self, state: ExecutionState):
...
         ...
 
-    async def perform_invoke(
+    async def perform_execute(
         self, function_name: str, *args, **kwargs
     ) -> StarknetCallInfo:
         try:
             func = getattr(self.state.contract, function_name)
-            return await func(*args, **kwargs).invoke()
+            return await func(*args, **kwargs).execute()
```
kasperski95, line 32: Is `execute` a new name for a method of StarkNet functions or did you rename this method accidentally?

mkaput: Its the new name:

https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/testing/contract.py#L281-L293

Arcticae (approved):

PR #776, opened 2022-09-08
Introduce Cheaters abstraction and use it to fix warp and roll cheatcodes

### LINKED ISSUE #767: Fix warp and roll cheatcodes

mkaput:
Not only this PR fixes two cheatcodes, it also introduces the first step in decoupling our `CheatebleState` inheritors from the `Cheatcodes` and other cheating inheritors like `CheatableSysCallHandler`.

fix #767


protostar/commands/test/cheatcodes/warp_cheatcode.py
```diff
@@ -17,14 +17,6 @@ def warp(
...
-        ] = blk_timestamp
-
-        def stop_warp():
-            del self.cheatable_state.contract_address_to_block_timestamp[
-                target_contract_address
-            ]
-
-        return stop_warp
+        # TODO
+        # target_contract_address = target_contract_address or self.contract_address
```
kasperski95, line 21:
```suggestion
       assert target_contract_address is not None, "target_contract_address functionality hasn't been migrated to Cairo 0.10"
```

kasperski95 (approved):
`cheatBlockInfo` code could have better types

protostar/starknet/cheatable_cached_state.py
```diff
@@ -122,3 +113,42 @@ def get_abi_from_contract_address(self, contract_address: int) -> AbiType:
...
         return self.class_hash_to_contract_abi_map[class_hash]
+
+
+def cheaters_of(state: StateProxy) -> Cheaters:
+    """
+    Extracts the ``Cheaters`` object from any State structures.
+
+    This function workarounds limitations of the inheritance design of ``State`` classes family,
+    preventing us from exposing the `cheaters` field via state interface classes like ``SyncState``.
+    """
```
Arcticae, line 124: does the syscall handler always interact with one of those, and not just one specific implementation that we use? i'm kind of missing that part

protostar/starknet/cheatable_syscall_handler.py
```diff
@@ -23,6 +25,14 @@ def __init__(self, message: str):
 
 
 class CheatableSysCallHandler(BusinessLogicSysCallHandler):
+    def __init__(self, state: SyncState, **kwargs):
+        # This field must be set before entering super constructor,
+        # because it calls the setter for the `block_info` property.
```
Arcticae, line 30: Why does it matter when this propery is set?

protostar/starknet/cheaters/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+from .block_info import BlockInfoCheater
+
+
+@dataclass(frozen=True)
+class Cheaters:
+    block_info: BlockInfoCheater
+
+    def copy(self) -> Self:
+        return Cheaters(block_info=self.block_info.copy())
```
Arcticae, line 13: We might want to make the generalization easier here, so we don't have to add new bits here every time we add a cheater

Arcticae (approved):
LGTM - although i do have a couple of questions about implementation

protostar/starknet/cheaters/__init__.py
mkaput: I had a working code for `copy` and `apply` which used a combination of `dir()`, `getattr` and `setattr` but reverted it back because I sounded to me overly complicated at this step.

I have also thought about avoiding having a class that knows about all cheaters, and having a system which did `give me a cheater of this type`; but dynamic auto initialization of cheaters is not straightforward in this case because they take arbitrary dependencies from Starknet internals, like the `BlockInfo` thingy done here.

protostar/starknet/cheatable_syscall_handler.py
mkaput: Because of this https://github.com/starkware-libs/cairo-lang/blob/v0.10.0/src/starkware/starknet/core/os/syscall_utils.py#L77-L80, which is a super super constructor of this class.

protostar/starknet/cheatable_cached_state.py
mkaput: Syscall handler always interacts with a thing typed as `SyncState` which is an abstract interface implemented by arbitrary classes. Ideally I would like to add a method to `SyncState` but I cannot do this because I cannot modify a defined type nor define an extension to it in Python 😅 

I also avoid making assumptions that syscalls always get an instance of `StateSyncifier`, because this fact may change in the future.

mkaput: Extension methods [are theoretically possible](https://stackoverflow.com/a/7139161), but I think both @kasperski95 and Pyright will not like it 😆

MaksymilianDemitraszek (approved):

PR #779, opened 2022-09-08
Align deploy and declare cheatcodes with new state implementation

### LINKED ISSUE #766: Fix declare and deploy cheatcodes

Arcticae:
fixes #766

### DIFF
--- a/protostar/commands/test/cheatcodes/declare_cheatcode.py
+++ b/protostar/commands/test/cheatcodes/declare_cheatcode.py
@@ -5,6 +5,7 @@ from typing import Optional, Dict
 from starkware.python.utils import from_bytes
 from starkware.starknet.business_logic.transaction.objects import InternalDeclare
 from starkware.starknet.public.abi import AbiType
+from starkware.starknet.services.api.gateway.transaction import DEFAULT_DECLARE_SENDER_ADDRESS
 from starkware.starknet.testing.contract import DeclaredClass
 from starkware.starknet.testing.contract_utils import EventManager, get_abi
 
@@ -59,21 +60,26 @@ class DeclareCheatcode(Cheatcode):
     async def _declare_contract(self, contract_path: Path):
         contract_class = self._starknet_compiler.compile_contract(contract_path)
 
-        tx = await InternalDeclare.create_for_testing(
-            ffc=self.state.ffc,
+        tx = InternalDeclare.create(
             contract_class=contract_class,
             chain_id=self.general_config.chain_id.value,
+            sender_address=DEFAULT_DECLARE_SENDER_ADDRESS,
+            max_fee=0,
+            version=0,
+            signature=[],
+            nonce=0,
         )
 
-        with self.state.copy_and_apply() as state_copy:
-            tx_execution_info = await tx.apply_state_updates(
-                state=state_copy, general_config=self.general_config
-            )
+        await tx.apply_state_updates(
+            state=self.cheatable_state, general_config=self.general_config
+        )
 
         abi = get_abi(contract_class=contract_class)
         self._add_event_abi_to_state(abi)
-        class_hash = tx_execution_info.call_info.class_hash
+        class_hash = tx.class_hash
         assert class_hash is not None
+        await self.cheatable_state.set_contract_class(class_hash, contract_class)
+
         return DeclaredClass(
             class_hash=from_bytes(class_hash),
             abi=get_abi(contract_class=contract_class),
--- a/protostar/commands/test/cheatcodes/declare_cheatcode.py
+++ b/protostar/commands/test/cheatcodes/declare_cheatcode.py
@@ -70,9 +70,10 @@ class DeclareCheatcode(Cheatcode):
             nonce=0,
         )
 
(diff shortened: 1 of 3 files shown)


mkaput:
does this fix #766 ?

protostar/commands/test/cheatcodes/deploy_cheatcode.py
```diff
@@ -27,5 +32,26 @@ def deploy_prepared(
...
+            constructor_calldata=prepared.constructor_calldata,
+            hash_value=deploy_tx_hash,
+            version=0,
+        )
+
+        asyncio.run(self._run_deploy_tx(tx))
+        return DeployedContract(contract_address=prepared.contract_address)
+
+    async def _run_deploy_tx(self, tx: InternalDeploy):
+        await tx.apply_state_updates(
```
mkaput, line 55: again, isn't copy_and_apply needed here?

protostar/commands/test/cheatcodes/deploy_cheatcode.py
```diff
@@ -27,5 +32,26 @@ def deploy_prepared(
...
+        tx = InternalDeploy(
+            contract_address=prepared.contract_address,
+            contract_address_salt=prepared.salt,
+            contract_hash=to_bytes(prepared.class_hash),
+            constructor_calldata=prepared.constructor_calldata,
+            hash_value=deploy_tx_hash,
+            version=0,
+        )
+
+        asyncio.run(self._run_deploy_tx(tx))
```
mkaput, line 51: There's a method called `sync_apply_state_updates` in `InternalStateTransaction` which I think we can use here instead of calling into asyncio

mkaput (changes requested):
Also, it would be awesome to hide the logic of this behind the `Cheater` interface introduced in #776, but I can also do this in separate PR.

protostar/commands/test/cheatcodes/declare_cheatcode.py
```diff
@@ -59,21 +60,26 @@ def declare(
...
+            version=0,
+            signature=[],
+            nonce=0,
         )
 
-        with self.state.copy_and_apply() as state_copy:
-            tx_execution_info = await tx.apply_state_updates(
-                state=state_copy, general_config=self.general_config
-            )
+        await tx.apply_state_updates(
```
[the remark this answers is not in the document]
Arcticae: if anything fails we want the app to basically crash, right? i don't see the scenario where we would want to roll back from that kind of failure

protostar/commands/test/cheatcodes/deploy_cheatcode.py
Arcticae: OK i will check that one out

protostar/commands/test/cheatcodes/declare_cheatcode.py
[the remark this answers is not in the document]
mkaput: We want it to crash but that doesn't mean that we should leave the state in dirty state IMHO. Who knows if the lack of rolling will not break things in the future for example... 🤔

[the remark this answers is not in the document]
Arcticae: Yeah this sounds reasonable.

Arcticae:
> Also, it would be awesome to hide the logic of this behind the `Cheater` interface introduced in #776, but I can also do this in separate PR.

I am wondering how should i abstract away this logic into a Cheater - i operate directly on the state here and don't need to store anything else really

mkaput:
> > Also, it would be awesome to hide the logic of this behind the `Cheater` interface introduced in #776, but I can also do this in separate PR.
> 
> I am wondering how should i abstract away this logic into a Cheater - i operate directly on the state here and don't need to store anything else really

The trick is that I would like to revoke a direct access to the state in Cheatcodes. Okay let's do this in separate PR because this is not fully clear indeed.

protostar/commands/test/cheatcodes/deploy_cheatcode.py
Arcticae: this one is a bummer because the `cheatable_state` does not pass instance check of a sync state, which crashes in case i use `sync_apply_state_updates`.

protostar/commands/test/cheatcodes/declare_cheatcode.py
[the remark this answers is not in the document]
Arcticae: done

protostar/commands/test/cheatcodes/deploy_cheatcode.py
Arcticae: done

MaksymilianDemitraszek (approved):
Please address @mkaput comments before merging though

protostar/commands/test/cheatcodes/deploy_cheatcode.py
mkaput: I think you can call copy and apply manually in the same way copy_and_apply does internally. I have asked StarkWare people about this internally.

mkaput (approved):

PR #786, opened 2022-09-09
Fixed load, store

MaksymilianDemitraszek:


kasperski95 (approved):
LGTM, if load/store tests pass.

MaksymilianDemitraszek:
@kasperski95 @mkaput 
I set me as a codeowner of those files

PR #800, opened 2022-09-12
Fixed sharing state between fuzz test runs

Radinyn:


tests/integration/fuzzing/fuzzing_test.py
```diff
@@ -109,3 +109,16 @@ async def test_max_examples_invalid_arguments(
         expected_passed_test_cases_names=[],
         expected_broken_test_cases_names=["test_zero", "test_negative"],
     )
+
+
+async def test_should_not_shared_state(
```
Radinyn, line 114:
```suggestion
async def test_should_not_share_state(
```

mkaput (approved):

PR #803, opened 2022-09-12
Improve crash reporting in E2E tests and fix failing ones

### LINKED ISSUE #798: Upgrade Cairo code in E2E tests

mkaput:
The only failing tests left are the devnet-dependent ones.

fixes #798

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,4 +1,5 @@
 # pylint: disable=redefined-outer-name
+import shlex
 import shutil
 from os import chdir, mkdir, path, getcwd
 from pathlib import Path
@@ -103,7 +104,12 @@ def init_project(
 
 
 class ProtostarFixture(Protocol):
-    def __call__(self, args: List[str], ignore_exit_code=False) -> str:
+    def __call__(
+        self,
+        args: List[str],
+        expect_exit_code: int = 0,
+        ignore_exit_code: bool = False,
+    ) -> str:
         ...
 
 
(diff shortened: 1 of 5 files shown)


tests/e2e/test_building.py
```diff
@@ -138,20 +137,26 @@ def test_disable_hint_validation(protostar):
...
-                    "",
-                    "@external",
-                    "func __execute__():",
-                    "    return ()",
-                    "end",
-                ]
-            )
-        )
+    Path("./src/main.cairo").write_text(
+        """\
```
Arcticae, line 141: can we do `dedent` here as well?

tests/e2e/conftest.py
```diff
@@ -155,18 +161,32 @@ def protostar(
...
+    ) -> str:
+        completed = run(
+            [path.join(tmp_path, "dist", "protostar", "protostar")] + args,
+            stdout=PIPE,
+            stderr=STDOUT,
+            encoding="utf-8",
         )
 
+        if not ignore_exit_code:
+            if completed.returncode != expect_exit_code:
```
Arcticae, line 177: i'd write a custom assertion here, while returning some typed object. assertion could even be added as a method on that object, that way test would be more readable and "conventional". not needed ASAP though, can we add a task to the backlog for this?

Arcticae (approved):

kasperski95 (approved):

tests/e2e/test_building.py
mkaput: Done.

tests/e2e/conftest.py
mkaput: Discussed internally.

PR #810, opened 2022-09-12
Added test skipping

Radinyn:
image

- [x] Add functionality
- [x] Add tests

Docs in seperate PR


protostar/commands/test/test_case_runners/test_case_runner.py
```diff
@@ -97,3 +101,13 @@ def _map_breaking_reported_exception_to_broken_test_result(
             execution_time=execution_metadata.execution_time,
             captured_stdout=self._output_recorder.get_captures(),
         )
+
+    def _map_skipping_reported_exception_to_skipped_test_result(
```
Arcticae, line 105: this seems obsolete, since it's only used in 1 place and returns immediately

protostar/commands/test/test_environment_exceptions.py
```diff
@@ -81,6 +85,12 @@ class SimpleBreakingReportedException(
     pass
 
 
+class SimpleSkippingReportedException(
```
Arcticae, line 88: Why 2 subclasses instead of 1?

protostar/commands/test/cheatcodes/skip_cheatcode.py
```diff
@@ -0,0 +1,20 @@
...
+class SkipCheatcode(Cheatcode):
+    @property
+    def name(self) -> str:
+        return "skip"
+
+    def build(self) -> Callable[..., Any]:
+        return self.skip
+
+    def skip(self, condition: bool = True, reason: str = "") -> None:
+        if condition:
```
Arcticae, line 19: i don't know if i like that `condition` thing... i'll let others decide on that one, but i have mixed feelings about this, since one can just use `if` inside of the hint.

Arcticae (changes requested):
LGTM overall, small nitpicks

protostar/commands/test/cheatcodes/skip_cheatcode.py
mkaput: Hmm, using the `if` is actually neat!

Pytest doesn't have such opportunity because `skip` is a decorator there. I agree with @Arcticae

protostar/commands/test/cheatcodes/skip_cheatcode.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SkipCheatcode(Cheatcode):
+    @property
+    def name(self) -> str:
+        return "skip"
+
+    def build(self) -> Callable[..., Any]:
+        return self.skip
+
+    def skip(self, condition: bool = True, reason: str = "") -> None:
```
mkaput, line 18:
```suggestion
    def skip(self, condition: bool = True, reason: Optional[str] = None) -> None:
```

wouldn't this be more semantically correct?

protostar/commands/test/test_results.py
```diff
@@ -47,6 +47,11 @@ class BrokenTestCaseResult(TimedTestCaseResult):
     exception: ReportedException
 
 
+@dataclass(frozen=True)
+class SkippedTestCaseResult(TestCaseResult):
+    reason: str
```
mkaput, line 52:
```suggestion
    reason: Optional[str]
```

tests/integration/skipping/skipping_test.cairo
```diff
@@ -0,0 +1,39 @@
+%lang starknet
+
+@external
+func test_skip() {
+    %{ skip(True, "AAA") %}
```
mkaput, line 5: `skip` should not be available in test functions directly, only in all setup hooks (both `setup_*` and `__setup__`). I don't want to pollute `TestExecutionEnvironment` logic with handling this cheatcode.

I know that slapping `skip` within test case would often be the least amount of code to write, but I suspect this would be a problematic thing in long term.

I think it might be high time to extract a `CommonTestCheatcodesFactory` class and not treat `SetupCheacodesFactory` as one.

tests/integration/skipping/skipping_test.py
```diff
@@ -0,0 +1,60 @@
...
+            "test_skip_failed",
+            "test_skip_no_reason",
+            "test_skip_no_input",
+        ],
+    )
+
+    assert_any_has_key_as_reason(testing_summary.skipped, "AAA")
+    assert_any_has_key_as_reason(testing_summary.skipped, "CCC")
+    assert_none_have_key_as_reason(testing_summary.skipped, "BBB")
+    assert_none_have_key_as_reason(testing_summary.skipped, "DDD")
```
mkaput, line 40: You can dump reasons from all test cases as a dict and compare each exactly. Would not only be simpler code, but also future-proof.

protostar/commands/test/test_result_formatter.py
```diff
@@ -210,6 +213,26 @@ def _format_passed_fuzz_test_case_result(
...
+    result: List[str] = []
+    first_line: List[str] = []
+    first_line.append(f"[{log_color_provider.colorize('YELLOW', 'SKIP')}]")
+    formatted_file_path = _get_formatted_file_path(skipped_test_case_result.file_path)
+    first_line.append(
+        f"{formatted_file_path} {skipped_test_case_result.test_case_name}"
+    )
+    result.append(" ".join(first_line))
+
+    if skipped_test_case_result.reason:
```
mkaput, line 226:
```suggestion
    reason = skipped_test_case_result.reason
    if reason is not None:
```

protostar/commands/test/test_environment_exceptions.py
mkaput: On top of this, why not simply one class `TestSkipped` which has reason out of the box?

Also, I don't think this has to be a reported exception? `HypothesisFailureSmugglingError` is not for example.

mkaput (changes requested):

protostar/commands/test/test_case_runners/test_case_runner.py
Radinyn: I added it for consistency with other results

protostar/commands/test/environments/common_test_cheatcode_factory.py
```diff
@@ -0,0 +1,67 @@
...
+            ),
+            MockCallCheatcode(syscall_dependencies),
+            WarpCheatcode(syscall_dependencies),
+            RollCheatcode(syscall_dependencies),
+            StartPrankCheatcode(syscall_dependencies),
+            StoreCheatcode(syscall_dependencies),
+            LoadCheatcode(syscall_dependencies),
+            ReflectCheatcode(syscall_dependencies),
+            SkipCheatcode(syscall_dependencies),
+            MaxExamplesCheatcode(syscall_dependencies, self._state.config),
```
mkaput, line 60: these two should go to setup factory

mkaput (changes requested):

tests/integration/skipping/skipping_test.cairo
Radinyn: Done

protostar/commands/test/test_environment_exceptions.py
Radinyn: Done

tests/integration/skipping/skipping_test.py
Radinyn: Not needed with the new logic

protostar/commands/test/test_case_runners/test_case_runner.py
```diff
@@ -15,6 +15,7 @@
     FailedTestCaseResult,
     PassedTestCaseResult,
     BrokenTestCaseResult,
+    SkippedTestCaseResult,
```
mkaput, line 18: unused import

protostar/commands/test/cheatcodes/skip_cheatcode.py
```diff
@@ -0,0 +1,24 @@
...
+        raise TestSkipped(reason)
+
+
+class TestSkipped(BaseException):
+    def __init__(self, reason: Optional[str] = None) -> None:
+        self.reason = reason
+        super().__init__(reason)
+
+    def __str__(self) -> str:
+        return str(self.reason or "No reason specified.")
```
mkaput, line 24: is this used anywhere?

protostar/commands/test/test_results.py
```diff
@@ -47,6 +47,11 @@ class BrokenTestCaseResult(TimedTestCaseResult):
     exception: ReportedException
 
 
+@dataclass(frozen=True)
+class SkippedTestCaseResult(TimedTestCaseResult):
+    reason: Optional[str]
```
mkaput, line 52: not applicable now?

protostar/commands/test/test_results.py
```diff
@@ -134,12 +139,33 @@ def into_broken_test_case_result(self) -> BrokenTestCaseResult:
...
 @dataclass(frozen=True)
 class BrokenTestSuiteResult(TestResult):
     test_case_names: List[str]
     exception: BaseException
 
 
+@dataclass(frozen=True)
+class SkippedTestSuiteResult(TestResult):
+    test_case_names: List[str]
+    reason: Optional[str]
```
mkaput, line 166: not applicable now?

tests/integration/skipping/skipping_test.py
```diff
@@ -0,0 +1,40 @@
...
+from protostar.commands.test.test_result_formatter import format_test_result
+from protostar.commands.test.test_results import SkippedTestCaseResult
+
+from tests.integration.conftest import (
+    RunCairoTestRunnerFixture,
+    assert_cairo_test_cases,
+)
+
+
+@pytest.mark.asyncio
```
mkaput, line 15:
```suggestion
```

tests/integration/skipping/skipping_test.py
```diff
@@ -0,0 +1,40 @@
+from pathlib import Path
```
mkaput, line 1: move this to `test/integration/cheatcodes/skip/skip_test.py`

tests/integration/skipping/skipping_test.py
```diff
@@ -0,0 +1,40 @@
...
+    assert_cairo_test_cases(
+        testing_summary,
+        expected_failed_test_cases_names=["test_skip_outside_of_setup"],
+        expected_skipped_test_cases_names=[
+            "test_skip",
+            "test_skip_no_reason",
+        ],
+    )
+
+    output = get_formatted_output(testing_summary.skipped)
```
mkaput, line 32: We do not test formatting in integration tests, but rather by unit testing TestingSummary, or by doing E2E tests.

This test should rather assert that expected data structures are returned from test runner.

mkaput (changes requested):

protostar/commands/test/test_results.py
Radinyn: still used for mapping SkippedSetupCaseResult to SkippedTestCaseResult

tests/integration/skipping/skipping_test.py
Radinyn: Added E2E test

mkaput (approved):
nice!

PR #813, opened 2022-09-12
Fix up PYTHONPATH to work as described in the docs

### LINKED ISSUE #422: Allow for setting sys.path in protostar.toml
As mentioned [in discord](https://canary.discord.com/channels/793094838509764618/964116575387021413/994337657842839653), I'd like to import my own python functions in hints. This is currently not straightforward because I have to manipulate `sys.path`.

What I've currently settled on is to do the path manipulation in `__setup__` and imports in the hints themselves.

```
@external
func __setup__():
    %{
        import os, sys
        sys.path.append(os.path.abspath(os.path.dirname(".")))
    %}

    return ()
end

@external
func test_foo():
    # ...
    
    local expected
    %{
        from tests.my_module.utils import my_func
        ids.expected = my_func(42)
    }%
    # ...
end
```

It works, but could be made better.

One suggestion is to automatically have the directory in which the `test_foo.cairo` file is located placed in `sys.path` so that I can have `utils.py` sitting in the dir structure right next to my test files. Another thing that would be beneficial is to have a config setting in `protostar.toml` where I could list paths that I want to have added to `sys.path` for each test file so that I don't have to manipulate the path inside `__setup__`.

Arcticae:
TODO: 
- [x] Add more explicit docs about it, since it concerns not only signing (custom signer classes), but also availability of modules through hints.


protostar/protostar_cli.py
```diff
@@ -15,6 +17,13 @@
 from protostar.utils.log_color_provider import LogColorProvider
 
 
+def _consume_pythonpath():
+    pythonpath_env_var = os.environ.get("PYTHONPATH") or ""
+    split_paths = pythonpath_env_var.split(":")
+    if split_paths != [""]:
+        sys.path.extend(split_paths)
```
mkaput, line 24: Edge cases I don't know if are dealt with properly:
1. I'm not sure how this will work with `~` in the path (maybe https://docs.python.org/3/library/os.path.html#os.path.expanduser must also be called?)
2. How to add a path which contains `:` within?

protostar/protostar_cli.py
```diff
@@ -15,6 +17,13 @@
 from protostar.utils.log_color_provider import LogColorProvider
 
 
+def _consume_pythonpath():
```
mkaput, line 20: hmm, this function is not clearing nor removing the env var, so the term `consume` is misleading here. maybe:

```suggestion
def _apply_pythonpath():
```

website/docs/tutorials/07-testing/README.md
```diff
@@ -159,3 +159,53 @@ func test_something(a: felt) {
     return ();
 }
 ```
+
+### Custom python modules in hints
```
mkaput, line 163:
```suggestion
### Importing Python modules in hints
```

website/docs/tutorials/07-testing/README.md
```diff
@@ -159,3 +159,53 @@ func test_something(a: felt) {
...
+
+You might want to use some non-standard python code in your hint segments to i.e. verify a signature using an external library.
+
+In order to do that, you should use `PYTHONPATH` env variable (https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH), and protostar will consume it to extend `sys.path` accordingly.
+This approach will make it possible to include some packages from your virtual environment (by adding `site_packages` to your `PYTHONPATH`).
```
mkaput, line 168:
```suggestion
The `PYTHONPATH` environment variable (https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH) is fully supported, and Protostar will extend `sys.path` with this variable's value in executed Cairo code.
This approach can be used to include some packages from Python virtual environment (by adding `site_packages` to the `PYTHONPATH`).
```

website/docs/tutorials/07-testing/README.md
```diff
@@ -159,3 +159,53 @@ func test_something(a: felt) {
...
+
+Another approach is leveraging that project's `cairo-path` is added to `sys.path`, so you can put your python code alongside your cairo files.
```
mkaput, line 170:
```suggestion
The `cairo-path` is automatically to `sys.path` in executed hints. This includes project root, `src` and `lib` directories. Any Python module files stored there can be imported without any extra configuration.
```

mkaput, line 170: I think I would prefer using cairo path over python path, so I would move this paragraph up.

website/docs/tutorials/07-testing/README.md
```diff
@@ -159,3 +159,53 @@ func test_something(a: felt) {
...
+```
+
+In `pymodule.py`:
+
+```python
+def get_three():
+    return 3
+```
+
+In `test_main.cairo` you should be able to do this:
```
mkaput, line 192:
```suggestion
The `get_three` function can be used in `test_main.cairo` like this:
```

website/docs/tutorials/07-testing/README.md
```diff
@@ -159,3 +159,53 @@ func test_something(a: felt) {
...
+
+```cairo
+%lang starknet
+from src.main import balance, increase_balance
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+
+@view
+func test_getting_tree(){
```
mkaput, line 200:
```suggestion
func test_getting_tree() {
```

protostar/protostar_cli.py
Arcticae: 1. Tested - doesn't need to be called
2. I think this would be rare enough to just ignore it for the time being. If somebody reports this as an issue we should circle back on this

protostar/protostar_cli.py
Arcticae: done

website/docs/tutorials/07-testing/README.md
Arcticae: Done

protostar/protostar_cli.py
```diff
@@ -15,6 +17,13 @@
...
+    pythonpath_env_var = os.environ.get("PYTHONPATH") or ""
+    split_paths = pythonpath_env_var.split(":")
```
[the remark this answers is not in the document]
Arcticae: done

protostar/protostar_cli.py
mkaput: 👍🏻

mkaput:
didn't this fix #422 ?

PR #815, opened 2022-09-13
Fuzzing strategies documentation

### LINKED ISSUE #633: Restore and update strategies docs

mkaput:
fix #633, expands them with new features info

### DIFF
--- a/website/docs/tutorials/07-testing/02-cheatcodes/given.md
+++ b/website/docs/tutorials/07-testing/02-cheatcodes/given.md
@@ -1,0 +1,42 @@
+# `given`
+
+```python
+def given(**kwargs: StrategyDescriptor) -> None:
+```
+
+Instructs the fuzzer to adopt a new fuzzing strategy to input parameters.
+The built-in strategies are provided by the [`strategy`](./strategy.md) cheatcode, and a list of
+available strategies, see the [fuzzing strategies](../03-fuzzing/strategies.md) guide page.
+
+:::warning
+This cheatcode is only available in [fuzz tests](../fuzzing).
+:::
+
+```cairo title="Example"
+@external
+func test_integers{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt):
+    %{
+        given(
+            a = strategy.integers(10, 20),
+            b = strategy.integers(30, 40),
+        )
+    %}
+    assert_le(a, b)
+    return ()
+end
+```
+
+## Performance caveats
+
+While it is possible to call `given` multiple times in a test case, try to always do it once.
+Multiple input parameters should be configured by passing many arguments to the cheatcode.
+Due to Cairo language limitations and resulting therefrom implementation details,
+each call to the `given` cheatcode causes the fuzzer to restart immediately with a fresh state.
+In consequence, each call effectively reduces a space of explored input parameters values,
+because fuzzer will test previously tested values, thus making the test less effective.
+
+Avoid binding fuzzing strategies to input parameters values or any other external state,
+such as random data of system time.
+This would cause non-deterministic behaviour of strategy learning and may yield bad results.
+Protostar detects if a test exhausts the maximum number of fuzzing runs and reports an appropriate
+error.
--- a/website/docs/tutorials/07-testing/02-cheatcodes/given.md
+++ b/website/docs/tutorials/07-testing/02-cheatcodes/given.md
@@ -4,39 +4,30 @@
 def given(**kwargs: StrategyDescriptor) -> None:
 ```
 
-Instructs the fuzzer to adopt a new fuzzing strategy to input parameters.
-The built-in strategies are provided by the [`strategy`](./strategy.md) cheatcode, and a list of
-available strategies, see the [fuzzing strategies](../03-fuzzing/strategies.md) guide page.
+Instructs the fuzzer to adopt a different fuzzing strategy for input parameters.
+The built-in strategies are provided by the [`strategy`](./strategy.md) cheatcode,
+and for a list of available strategies, see the [fuzzing strategies](../03-fuzzing/strategies.md)
+guide page.
 
 :::warning
-This cheatcode is only available in [fuzz tests](../fuzzing).
+This cheatcode is only available in [setup cases](../README.md#setup-case).
 :::
 
 ```cairo title="Example"
 @external
-func test_integers{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt):
+func setup_integers() {
     %{
         given(
             a = strategy.integers(10, 20),
             b = strategy.integers(30, 40),
         )
     %}
-    assert_le(a, b)
-    return ()
-end
+    return ();
+}
+
+@external
+func test_integers{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt) {
+    assert_le(a, b);
+    return ();
+}
 ```
-
-## Performance caveats
-
-While it is possible to call `given` multiple times in a test case, try to always do it once.
-Multiple input parameters should be configured by passing many arguments to the cheatcode.
-Due to Cairo language limitations and resulting therefrom implementation details,
-each call to the `given` cheatcode causes the fuzzer to restart immediately with a fresh state.
-In consequence, each call effectively reduces a space of explored input parameters values,
-because fuzzer will test previously tested values, thus making the test less effective.
-
-Avoid binding fuzzing strategies to input parameters values or any other external state,
-such as random data of system time.
-This would cause non-deterministic behaviour of strategy learning and may yield bad results.
-Protostar detects if a test exhausts the maximum number of fuzzing runs and reports an appropriate
-error.
--- a/website/docs/tutorials/07-testing/02-cheatcodes/given.md
+++ b/website/docs/tutorials/07-testing/02-cheatcodes/given.md
@@ -15,7 +15,7 @@ This cheatcode is only available in [setup cases](../README.md#setup-case).
 
 ```cairo title="Example"
 @external
-func setup_integers() {
+func setup_less_equal_compare() {
     %{
         given(
             a = strategy.integers(10, 20),
@@ -26,7 +26,7 @@ func setup_integers() {
 }
 
 @external
-func test_integers{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt) {
+func test_less_equal_compare{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt) {
     assert_le(a, b);
     return ();
 }
--- a/website/docs/tutorials/07-testing/03-fuzzing/strategies.md
+++ b/website/docs/tutorials/07-testing/03-fuzzing/strategies.md
@@ -68,7 +68,6 @@ strategy.integers(11)
 Fuzzer picks integers from provided range and then converts them to felts.
 If `min_value` is not `None` then all values will be greater than or equal to `min_value`,
 and if `max_value` is not `None` then all values will be less than or equal to `max_value`.
-When applied to field elements, the unbounded values may rarely overflow.
 
 ## Adapting strategies
 
--- a/website/docs/tutorials/07-testing/03-fuzzing/strategies.md
+++ b/website/docs/tutorials/07-testing/03-fuzzing/strategies.md
@@ -1,13 +1,12 @@
 # Strategies
 
-While a felt is a basis of all types in Cairo, its whole value space rarely matches what values are
-acceptable in tested code (for example, when using range checks builtin).
+Often in code we operate on some assumptions about the possible values of variables. 
+For example we can perform greater than zero validation early in the code, and we can assume that variables are greater than zero in the subsequent code. 
+Fuzzer allows to provide such assumptions to avoid testing against values which are not covered by the code. 
 
-Using the [`given`](../02-cheatcodes/given.md) cheatcode,
-one can instruct Protostar how constrained or adjusted the value space of fuzzed inputs should be.
-Such constraints are provided declaratively, by assigning _strategies_ to the input parameters.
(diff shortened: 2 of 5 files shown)


website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
+# Strategies
+
+While a felt is a basis of all types in Cairo, its whole value space rarely matches what values are
+acceptable in tested code (for example, when using range checks builtin).
```
MaksymilianDemitraszek, line 4:
```suggestion
Often in code we operate on some assumptions about the possible values of variables. 
For example we can perform greater than zero validation early in the code, and we can assume that variables are greater than zero in the subsequent code. 
Fuzzer allows to provide such assumptions to avoid testing against values which are not covered by the code. 
```

MaksymilianDemitraszek: I think it better explains strategies use case

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+
+Using the [`given`](../02-cheatcodes/given.md) cheatcode,
+one can instruct Protostar how constrained or adjusted the value space of fuzzed inputs should be.
+Such constraints are provided declaratively, by assigning _strategies_ to the input parameters.
+A strategy is an object that is capable of generating next input examples in specific way,
+and also is capable of other internal features, such as simplifying failing examples.
```
MaksymilianDemitraszek, line 10:
```suggestion
Cheatcode [`given`](../02-cheatcodes/given.md) ,
instructs fuzzer how to constraint set of values code is tested against.
Such constraints are provided declaratively, by assigning _strategies_ to the input parameters as on the example below.
```

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+
+By default, Protostar applies the [`strategy.felts()`](#strategyfelts) strategy to all felt
```
MaksymilianDemitraszek, line 12: Would move this under the example and add info callout

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+### `strategy.integers`
+
+```python
+def integers(
+    min_value: Optional[int] = None,
+    max_value: Optional[int] = None,
+) -> Strategy: ...
+```
+
+Generates integer values, possibly bounded to provided range.
```
MaksymilianDemitraszek, line 59:
```suggestion
Generates integer values, possibly bounded by provided range.
```

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+
+Assuming real numbers comparison semantics,
```
MaksymilianDemitraszek, line 61:
```suggestion
Fuzzer picks integers from provided range and then converts them to felts.
```

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+if `min_value` is not `None` then all values will be greater than or equal to `min_value`,
```
MaksymilianDemitraszek, line 62:
```suggestion
If `min_value` is not `None` then all values will be greater than or equal to `min_value`,
```

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+strategy.integers(11)
+```
+
+## Adapting strategies
+
+Often it is the case that a strategy does not produce exactly what is desired and a need to further
+adapt the generated values arise.
+Although this could be done in tests directly, this hurts because adaptation may waste fuzzing
+cycles (by repeatedly testing same values after adaptation) and the code has to be copied in every
+test.
```
MaksymilianDemitraszek, line 78:
```suggestion
```
It is useless information from user perspective

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+There are also the [`assume`] and [`reject`] cheatcodes, but this should only be used to discard
+only few unwanted inputs.
```
MaksymilianDemitraszek, line 80:
```suggestion
[`assume`] and [`reject`] cheatcodes provide simple interfaces to adapt a advanced strategy. Those are not very good considering the performance. Fuzzer can execute test on rejected data anyway and will just ignore failure if it happens.
```

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Filtering
+
+```python
+class Strategy:
+    def filter(self, filter_function: Callable[[int], bool]) -> Strategy: ...
+```
+
+Allows rejecting examples matching a condition.
+
```
MaksymilianDemitraszek, line 105: Please move example here, generally try to elevate examples higher, it's easier to understand for a reader

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+The outcome is similar to using the [`assume`] or [`reject`] cheatcodes, but `filter` does not
+require executing tested Cairo function and thus is more performant.
+Try to use `filter` only to avoid unwanted corner cases rather than attempting to cut out a large
+portion of the searched input values.
+
+The `filter` method is not magic and hard to satisfy conditions may cause the fuzzer to fail.
```
MaksymilianDemitraszek, line 111: I have no idea what this sentence means being honest

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,134 @@
...
+
+```python title="Example"
```
MaksymilianDemitraszek, line 113: Right now it looks like it presents
>  hard to satisfy conditions may cause the fuzzer to fail.

MaksymilianDemitraszek (changes requested):

website/docs/tutorials/07-testing/02-cheatcodes/given.md
```diff
@@ -0,0 +1,33 @@
...
+        given(
+            a = strategy.integers(10, 20),
+            b = strategy.integers(30, 40),
+        )
+    %}
+    return ();
+}
+
+@external
+func test_integers{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt) {
```
mkaput, line 29:
```suggestion
func test_less_equal_compare{syscall_ptr : felt*, range_check_ptr}(a : felt, b : felt) {
```

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
mkaput: `filter(lambda: 10 < x < 100)` will cause Hypothesis to raise exceptions that strategies are bad

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
mkaput: wouldn't such example be within `warning` block?

MaksymilianDemitraszek: I think even if we use such convention, reader does not have to be aware of it.
Also the loop of
- information
- example
- information
- example

is much beter from the cognition perspective than

-information
-information
-example
-example

It's better to keep examples close to related information

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
mkaput:
```suggestion
Fuzzer draws random data from the original strategy and only afterwards checks if it passes filter conditions.
If too many variables are restricted, fuzzer will reject test execution.
```

MaksymilianDemitraszek: Discussed on slack

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
MaksymilianDemitraszek: Discussed on slack

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
mkaput: ok, makes sense. thanks for tip!

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,135 @@
...
+Fuzzer picks integers from provided range and then converts them to felts.
+If `min_value` is not `None` then all values will be greater than or equal to `min_value`,
+and if `max_value` is not `None` then all values will be less than or equal to `max_value`.
+When applied to field elements, the unbounded values may rarely overflow.
+
+## Adapting strategies
+
+Often it is the case that a strategy does not produce exactly what is desired and a need to further
+adapt the generated values arise.
+Although this could be done in tests directly, this hurts because fuzzer does not know about the adaptation and may repeatedly test same values.
```
mkaput, line 77:
```suggestion
Although this could be done in tests directly, this hurts because fuzzer does not know about the adaptation and may repeatedly test the same values.
```

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,135 @@
+# Strategies
+
+Often in code we operate on some assumptions about the possible values of variables. 
+For example we can perform greater than zero validation early in the code, and we can assume that variables are greater than zero in the subsequent code. 
+Fuzzer allows to provide such assumptions to avoid testing against values which are not covered by the code. 
+
+Cheatcode [`given`](../02-cheatcodes/given.md) ,
```
MaksymilianDemitraszek, line 7:
```suggestion
Cheatcode [`given`](../02-cheatcodes/given.md),
```
Not sure if it will render correctly

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,135 @@
...
+```python title="Examples"
+strategy.integers(0, 100)
+strategy.integers(max_value=3000)
+strategy.integers(11)
+```
+
+Fuzzer picks integers from provided range and then converts them to felts.
+If `min_value` is not `None` then all values will be greater than or equal to `min_value`,
+and if `max_value` is not `None` then all values will be less than or equal to `max_value`.
+When applied to field elements, the unbounded values may rarely overflow.
```
MaksymilianDemitraszek, line 71: I am not sure what this sentence means

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Adapting strategies
+
+Often it is the case that a strategy does not produce exactly what is desired and a need to further
+adapt the generated values arise.
```
MaksymilianDemitraszek, line 76:
```suggestion
Often it is the case that a strategy does not produce exactly what is desired and it is necessary to further
adapt the generated values.
```
Simpler to read

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+### Filtering
+
+```python
+class Strategy:
+    def filter(self, filter_function: Callable[[int], bool]) -> Strategy: ...
+```
+
+Allows rejecting examples matching a condition.
```
MaksymilianDemitraszek, line 104:
```suggestion
Rejects examples not matching a condition.
```
?

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,135 @@
...
+Fuzzer draws random data from the original strategy and only afterwards checks if it passes filter conditions.
+If too many variables are restricted, fuzzer will reject test execution.
+
+### Combining
+
+```python
+def one_of(*strategies: Strategy) -> Strategy: ...
+```
+
+Return a strategy which generates values from any of the argument strategies.
```
MaksymilianDemitraszek, line 124:
```suggestion
Returns a strategy which generates values from any of the argument strategies.
```

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -0,0 +1,135 @@
...
+
+```python title="Example"
+strategy.one_of(
+    strategy.integers(0, 100),
+    strategy.integers(1000, 1200),
+)
+```
```
MaksymilianDemitraszek, line 131: What happens if strategies are overlaping?
Probability for some values increases?

MaksymilianDemitraszek (changes requested):
Some minor issues

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
mkaput:
```suggestion
```

MaksymilianDemitraszek (approved):

PR #820, opened 2022-09-13
ARM support

kasperski95:
- [x] check if the installer build on M1 works on M2 -> `“protostar” can’t be opened because Apple cannot check it for malicious software.`

> With Apple Silicon M1 architecture, macOS introduced mandatory code signing, even if ad-hoc (i.e., without actual code-signing identity). This means that arm64 arch slices (but possibly also x86_64 ones, especially in universal2 binaries) in collected binaries always come with signature.
- https://pyinstaller.org/en/stable/feature-notes.html#macos-multi-arch-support

ARM  |  Intel
--- | ---
33.92 s  |  37.81 s
33.85 s  |  38.27 s

~12% performance drop by using Rosetta


install.sh
```diff
@@ -3,45 +3,72 @@ set -e
 
 echo Installing protostar
 
+RESULT=""
```
mkaput, line 6: unused variable

kasperski95:
The overhead caused by signing application, adapting the pipeline, increased maintenance cost is not worth the effort in the near future.

Issue #605, opened 2022-08-12
Protostar seems to be incompatible with new M2 chip


Th0rgal:
Hello there. I just installed protostar on a new M2 macbook air. When trying to run ``protostar`` I got this message:
```
zsh: bad CPU type in executable: protostar
```

## Protostar version

| Name       | Version              |
| ---------- | -------------------- |
| Protostar  | `3.0.0`  |
| Cairo-lang | `not sure` |
| Git        | `2.36.0`        |

## Current Behavior

```
zsh: bad CPU type in executable: protostar
```

## Expected Behavior

It should start protostar

## Steps to Reproduce

1. Install protostar on a M2 mac computer
2. profits

MaksymilianDemitraszek:
We may look on it in the future(we have no m2 machine to test it though). For now i suggest building it by yourself based on https://github.com/software-mansion/protostar#development

Th0rgal:
> We may look on it in the future(we have no m2 machine to test it though). For now i suggest building it by yourself based on https://github.com/software-mansion/protostar#development
> 
> 
> 
> 

Thank you, I'm available if you need me to try things. I tried to build it but couldn't get the poe command. Where does it come from?

Th0rgal:
When building it by myself and calling it:
- init works
- build works
- test doesn't work

When performing the tests directly from source code in a newly created protostar project via ``python ~/workspace/protostar/binary_entrypoint.py test``, I got this error:
https://gist.github.com/Th0rgal/03c6902c65fe5afbbe2010792145d2ec

Th0rgal:
> When building it by myself and calling it:
> 
> * init works
> * build works
> * test doesn't work
> 
> When performing the tests directly from source code in a newly created protostar project via `python ~/workspace/protostar/binary_entrypoint.py test`, I got this error: https://gist.github.com/Th0rgal/03c6902c65fe5afbbe2010792145d2ec

I've been able to fix the tests issue by replacing ``multiprocessing`` from standardlib by ``multiprocess``, a fork which uses the dill serializer instead of pickle (it's the only difference). You can find more info about it here: https://pypi.org/project/multiprocess/

Testing by calling binary_entrypoint was fixed, but when I tried to build protostar via ``poe build``, the generated dist folder didn't contain multiprocess and I kept an issue with "multiprocessing" so I probably didn't do it correctly, I would really appreciate if you could show me how to correctly build it, or even replace multiprocessing by multiprocess in the final version.

[...]

PR #824, opened 2022-09-14
Refactor `install.sh`

kasperski95:
I need to add more code to this file without making a mess.

Related: https://github.com/software-mansion/protostar/issues/605

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -19,7 +19,7 @@ function get_platform_name() {
     esac
 }
 
-echo Installing protostar
+echo "Installing protostar"
 
 PROTOSTAR_DIR=${PROTOSTAR_DIR-"$HOME/.protostar"}
 mkdir -p "$PROTOSTAR_DIR"
@@ -52,7 +52,7 @@ fi
 
 PROTOSTAR_REPO="https://github.com/software-mansion/protostar"
 
-echo Retrieving $VERSION version from $PROTOSTAR_REPO...
+echo "Retrieving $VERSION version from $PROTOSTAR_REPO..."
 
 REQUESTED_RELEASE=$(curl -L -s -H 'Accept: application/json' "${PROTOSTAR_REPO}/releases/${REQUESTED_REF}")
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,96 +1,143 @@
...
-  Darwin)
-    PLATFORM="macOS"
-    ;;
-  *)
-    echo "unsupported platform: $PLATFORM"
-    exit 1
-    ;;
-esac
+PROTOSTAR_REPO="https://github.com/software-mansion/protostar"
+RESULT=""
```
mkaput, line 5: 1. I have seen name `RETVAL` in the wild (for example here: https://github.com/rust-lang/rustup/blob/master/rustup-init.sh)
2. This statement is NOOP as far as I know bash. You can safely remove it.

install.sh
```diff
@@ -1,96 +1,143 @@
...
+
+function create_protostar_directory() {
```
mkaput, line 7:
```suggestion
create_protostar_directory() {
```

FYI The `function` keyword is unnecessary (really!) :)

install.sh
```diff
@@ -1,96 +1,143 @@
...
+
+    if [[ ":$PATH:" != *":${_protostar_binary_dir}:"* ]]; then
+        echo >>$_profile && echo "export PATH=\"\$PATH:$_protostar_binary_dir\"" >>$_profile
+    fi
+    echo && echo "Detected your preferred shell is ${_pref_shell} and added Protostar to PATH. Run 'source ${_profile}' or start a new terminal session to use Protostar."
+    echo "Then, run 'protostar --help'."
+}
+
+function main() {
+    _provided_version_arg=$1
```
mkaput, line 100: why are you prefixing names with `_`?

you can use `local` here

mkaput (changes requested):

install.sh
kasperski95: To make locals stand out compared to function calls.

(I forgot about the `local` keyword in bash.)

install.sh
kasperski95: I know, but VS Code highlights `function` keyword differently. I'll remove it.

install.sh
kasperski95: Removed prefixes.

install.sh
kasperski95: Removed function keyword

install.sh
kasperski95: Removed global declaration. Renamed to retval

mkaput (approved):

PR #826, opened 2022-09-14
Profiling graph

MaksymilianDemitraszek:
This PR provides mvp of merging algorithm for the profile graph.
I will procede with refactors, some code comments, more features in following PRs

Screenshot 2022-09-14 at 22 57 35


tests/integration/profile_tests/example_profile_test.py
```diff
@@ -8,7 +8,7 @@
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip
+# @pytest.mark.skip
```
mkaput, line 11:
```suggestion
```

tests/integration/profile_tests/example_profile_test.py
```diff
@@ -8,7 +8,7 @@
 
 
 @pytest.mark.asyncio
```
mkaput, line 10:
```suggestion
```

not needed anymore

protostar/starknet/cheatable_execute_entry_point.py
```diff
@@ -233,7 +276,17 @@ def save_profile(program, memory, trace, debug_info, runner):
...
+    profile = build_profile(
+        tracer_data, runner.segments, runner.segment_offsets, runner.accessed_addresses
+    )
+    return profile
+
+def merge_and_save(contract_samples: List[ContractProfile]):
+    merged = merge_profiles(contract_samples)
+    proto = to_protobuf(merged)
+    serialized = serialize(proto)
     with open("profile.pb.gz", "wb") as file:
```
mkaput, line 290: https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_bytes

protostar/profiler/profile.py
```diff
@@ -15,20 +18,25 @@
...
+if TYPE_CHECKING:
+    from protostar.starknet.cheatable_execute_entry_point import ContractProfile
 
 Address = int
 StringID = int
 FunctionID = int
 
+import random
+def unique_id():
+    return random.randint(0, 10000000000)
```
mkaput, line 30:
```suggestion
_unique_id_counter = 0
def unique_id():
    global _unique_id_counter
    _unique_id_counter += 1
    return _unique_id_counter
```

This is more collision proof. Or use UUIDv4 if you need global uniqueness.

protostar/profiler/profile.py
```diff
@@ -2,8 +2,11 @@
 from collections import defaultdict
 from dataclasses import dataclass
 import functools
+import itertools
 import math
-from typing import Callable, List, Dict, Set, cast
+from re import A, U
+from typing import TYPE_CHECKING, Callable, List, Dict, Set, Tuple, cast
+from xxlimited import new
```
mkaput, line 9: unused import?
```suggestion
```

protostar/profiler/profile.py
MaksymilianDemitraszek: I will have to change the alghoritm (forgot about it). This solution won't work because it is used as an offset so we need some free spots after each id.

protostar/profiler/profile.py
MaksymilianDemitraszek: i didn't lint the code

MaksymilianDemitraszek: opening PRs at midnight 😅

protostar/profiler/profile.py
MaksymilianDemitraszek: #832

MaksymilianDemitraszek:
Merging, as probably nobody will be reviewing the logic

PR #839, opened 2022-09-16
Extracting `test_runner` module

kasperski95:
Conflict-prone PR. Quick review plz.


mkaput (commented):
nitpick: What about `protostar.testing` module? It does not have underscore in names and looks cooler 😅

protostar/test_runner/test_environment_exceptions.py
```diff
@@ -1,14 +1,15 @@
...
+from typing import Any, Dict, List, Optional, Type, TypeVar, Union
 
 from starkware.starknet.business_logic.execution.objects import Event
 from typing_extensions import Literal
 
-from protostar.commands.test.expected_event import ExpectedEvent
 from protostar.starknet.hint_local import HintLocal
 from protostar.utils.log_color_provider import SupportedColorName, log_color_provider
 
+from .expected_event import ExpectedEvent
```
mkaput, line 11: ReportedException, CheatcodeException and some more should go to `protostar.starknet` module. They are also used in the migrator!

protostar/test_runner/testing_seed.py
```diff
@@ -5,6 +5,7 @@
 
 
 def determine_testing_seed(seed: Optional[Seed]) -> Seed:
+
```
mkaput, line 8:
```suggestion
```

mkaput (changes requested):

mkaput (approved):

protostar/test_runner/test_environment_exceptions.py
kasperski95: Is this resolved?

kasperski95: Marking as resolved because PR is approved.

PR #850, opened 2022-09-19
Delay Git exception

Radinyn:


protostar/git.py
```diff
@@ -0,0 +1,215 @@
...
+
+    @staticmethod
+    def clone(path_to_repo: Path, repo_to_clone: "GitRepository"):
+        path_to_repo.parent.mkdir(parents=True, exist_ok=True)
+        cloned_repo = GitRepository(path_to_repo)
+        cloned_repo.clone(repo_to_clone.path_to_repo)
+        return cloned_repo
+
+    @staticmethod
+    def from_existing(path_to_repo: Path):
```
kasperski95, line 69: - `from_*` naming convention is used by static factories. It suggests creating `Git` not `GitRepository`. Maybe `create_from_initialized`
- `path_to_repo` could be provided through the constructor

protostar/git.py
```diff
@@ -0,0 +1,215 @@
...
+        "stderr": subprocess.DEVNULL,
+    }
+)
+
+CREDENTIALS = [
+    "-c",
+    'user.name="Protostar"',
+    "-c",
+    'user.email="protostar@protostar.protostar"',
+]
```
kasperski95, line 30: I think we should use the user credentials.
(Next.js uses user credentials for the first commit.)

protostar/git.py
```diff
@@ -0,0 +1,215 @@
...
+                "add",
+                str(path_to_item),
+            ],
+            **SHARED_KWARGS,
+            cwd=self.path_to_repo,
+        )
+
+    # pylint: disable=invalid-name
+    @wrap_git_exception
+    def rm(self, path_to_item: Path, force: bool = False):
```
kasperski95, line 170:
```suggestion
    def remove_submodule(self, submodule_name: str):
```
Just to be consistent with, `get`, `add`, and `update` submodule and to narrow down the use case.

protostar/git.py
```diff
@@ -0,0 +1,215 @@
...
+    'user.email="protostar@protostar.protostar"',
+]
+
+
+class ProtostarGitException(ProtostarException):
+    def __init__(self, message: str, details: Optional[str] = None):
+        super().__init__("Error while executing Git command:\n" + message, details)
+
+
+class NotARepositoryException(ProtostarGitException):
```
kasperski95, line 38: We can skip articles. I'd rename this exception to:

```suggestion
class InvalidGitRepositoryException(ProtostarGitException):
```

protostar/git.py
```diff
@@ -0,0 +1,215 @@
...
+    def wrapper(*args, **kwargs):
+        try:
+            func(*args, **kwargs)
+        except subprocess.CalledProcessError as ex:
+            raise ProtostarGitException(str(ex)) from ex
+
+    return wrapper
+
+
+class Git:
```
kasperski95, line 53: It could be moved to a separate file.

protostar/git.py
kasperski95, line 38: Exceptions could be at the bottom of a file.


Vertical formatting: https://levelup.gitconnected.com/javascript-clean-code-vertical-formatting-d15097f5f30e

protostar/git.py
```diff
@@ -0,0 +1,215 @@
+import subprocess
```
kasperski95, line 1: This class could be tested.

protostar/git.py
```diff
@@ -0,0 +1,215 @@
...
+    def from_existing(path_to_repo: Path):
+        path = GitRepository.get_repo_root(path_to_repo)
+        return GitRepository(path)
+
+
+class GitRepository:
+    def __init__(self, path_to_repo: Path):
+        """
+        This class should not be instantiated directly, use `Git.init/clone/from_existing` instead.
+        """
```
kasperski95, line 78: Is this comment necessary?

kasperski95 (approved):

kasperski95:
You can resolve threads in another PR.

protostar/git.py
Radinyn: I'm not sure whether this will work because we pipe stdout to dev/null

PR #853, opened 2022-09-19
Add profiling cli API

MaksymilianDemitraszek:


tests/e2e/test_testing.py
```diff
@@ -19,6 +20,18 @@ def test_safe_collecting(protostar):
...
+@pytest.mark.usefixtures("init")
+def test_basic_contract_profile(protostar):
+    result = protostar(["test", "--profiling", "tests/test_main.cairo::test_increase_balance"])
+    assert "1 passed" in result
+    assert 'profile.tar.gz' in listdir(".")
+
+@pytest.mark.usefixtures("init")
+def test_basic_contract_profile_multiple(protostar):
+    result = protostar(["test", "--profiling","tests"], ignore_exit_code=True)
+    assert "Please specifiy path to a single test case whe running profiling" in result
```
mkaput, line 32:
```suggestion
    assert "Only one test case can be profiled at the time. Please specify path to a single test case." in result
```

protostar/commands/test/test_case_runners/test_case_runner_factory.py
```diff
@@ -14,16 +14,19 @@
...
 class TestCaseRunnerFactory:
     def __init__(self, state: TestExecutionState) -> None:
         self._state = state
 
-    def make(self, test_case: TestCase) -> TestCaseRunner:
+    def make(self, test_case: TestCase, profiling=False) -> TestCaseRunner:
         mode = self._state.config.mode
 
         assert mode, "Test mode should be determined at this point."
+        if mode is TestMode.FUZZ and profiling:
```
mkaput, line 28: missing test for this condition

tests/e2e/test_testing.py
```diff
@@ -1,3 +1,4 @@
+from os import listdir
```
mkaput, line 1: Why not integrate test this?

protostar/commands/test/environments/standard_test_execution_environment.py
```diff
@@ -0,0 +1,14 @@
+from protostar.commands.test.environments.test_execution_environment import (
+    TestExecutionEnvironment,
+    TestExecutionResult,
+)
+
+
+class StandardTestExecutionEnvironment(TestExecutionEnvironment):
```
mkaput, line 7: I think people might want to profile fuzz tests and gather averages/boundaries of runs..

In such scenario I think it would be more scalable implementation where `TestExecutionEnvironment` stays with the same role it had and accepts the `profiling` flag, while `FuzzTestExecutionEnvironment` raises an error if this flag is set to `True`.

wdyt?

protostar/commands/test/test_case_runners/test_case_runner_factory.py
```diff
@@ -14,16 +14,19 @@
...
+            raise ProtostarException("You cannot profile fuzz tests")
```
mkaput, line 29:
```suggestion
            raise ProtostarException("Fuzz tests cannot be profiled yet.")
```

mkaput (changes requested):

protostar/commands/test/environments/standard_test_execution_environment.py
```diff
@@ -0,0 +1,14 @@
...
+    def __init__(self, *args, profiling=False, **kwargs):
+        self.profiling = profiling
+        super().__init__(*args, **kwargs)
+
+    async def execute(self, function_name: str) -> TestExecutionResult:
+        self.set_profile_flag(self.profiling)
```
kasperski95, line 13: `set_profile_flag` is defined in the `ExecutionEnvironment`. IMO that method should be moved to this class or this class should be removed.

The class name is also vague. I'd name it `ProfilableTestExecutionEnvironment` but we should composition over inheritance.

Maybe `TestExecutionProfiler` passed as a dependency to `TestExecutionEnvironment` would be better approach.

protostar/commands/test/test_case_runners/test_case_runner_factory.py
MaksymilianDemitraszek: I am not sure if they will be at any point being honest, I don't see the need for changing the message

protostar/commands/test/environments/standard_test_execution_environment.py
MaksymilianDemitraszek: Done

kasperski95 (changes requested):

kasperski95 (approved):
https://github.com/software-mansion/protostar/issues/875

protostar/commands/test/test_case_runners/test_case_runner_factory.py
mkaput: I am mostly objecting the use of word `you` 😅 `Fuzz tests cannot be profiled.` is also fine.

mkaput (approved):
Looks good except that one message 😅

protostar/commands/test/test_case_runners/test_case_runner_factory.py
MaksymilianDemitraszek: Fair point

MaksymilianDemitraszek: Done

PR #857, opened 2022-09-20
Update Starknet.py to v0.6.0 and update minor dependencies

### LINKED ISSUE #835: Update Starknet.py to v0.6.0a0

mkaput:
fix #835

This PR follows Starknet.py and introduces a breaking change to Signer interface, by adding a new abstract method. I think it is so minor that we can describe migration steps in the changelog without bumping Protostar major version.


kasperski95 (approved):

protostar/cli/network_command_util.py
```diff
@@ -75,10 +75,7 @@ def get_network_config(self) -> NetworkConfig:
 
     def get_gateway_client(self) -> GatewayClient:
         network_config = self.get_network_config()
```
Radinyn, line 77: Is chain id needed in the network config anymore?
It definitely doesn't break anything so it could stay, but we should probably research it in the future.

mkaput: Yes, this value is still passed to the Signer class constructor.

PR #859, opened 2022-09-20
Add rc_bound kwarg to felts strategy

### LINKED ISSUE #579: Add strategy.comparable()
Add and document a new fuzzing strategy:

```python
def strategy.comparable(
    min_value: Optional[int] = None,
    max_value: Optional[int] = None,
): ...
```

That will behave exactly the same as the existing `integers` strategy, but will limit explored range (no matter if unbound or specified by `min_value`/`max_value` arguments), to `range_check_builtin.bound` values, as Cairo stdlib comparison functions expect.

The goal is to let users not have to worry about values exceeding the range check in this builtin functions.

Consider making this the default strategy for `felt` parameters.

Consider extending the fuzzing tutorial in docs to use this strategy as a new step to simplify the test from having to expect_revert range overflow. If that makes sense (it doesn't have to).

mkaput:
fix #579

### DIFF
--- a/website/docs/tutorials/07-testing/03-fuzzing/strategies.md
+++ b/website/docs/tutorials/07-testing/03-fuzzing/strategies.md
@@ -43,10 +43,12 @@ cheatcode-namespace.
 ### `strategy.felts`
 
 ```python
-def felts() -> Strategy: ...
+def felts(*, comparable: bool = False) -> Strategy: ...
 ```
 
-Explores all possible felt values.
+By default, explores all possible felt values.
+
+If `comparable` is true, explores felts which can be range checked, i.e. `assert_le` will not crash.
 
 ### `strategy.integers`
 
--- a/website/docs/tutorials/07-testing/03-fuzzing/strategies.md
+++ b/website/docs/tutorials/07-testing/03-fuzzing/strategies.md
@@ -43,12 +43,12 @@ cheatcode-namespace.
 ### `strategy.felts`
 
 ```python
-def felts(*, comparable: bool = False) -> Strategy: ...
+def felts(*, rc_bound: bool = False) -> Strategy: ...
 ```
 
 By default, explores all possible felt values.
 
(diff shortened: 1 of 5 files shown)


website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -43,10 +43,12 @@ cheatcode-namespace.
...
 
 ```python
-def felts() -> Strategy: ...
+def felts(*, comparable: bool = False) -> Strategy: ...
 ```
 
-Explores all possible felt values.
+By default, explores all possible felt values.
+
+If `comparable` is true, explores felts which can be range checked, i.e. `assert_le` will not crash.
```
kasperski95, line 51:
```suggestion
If `comparable` is `True`, explores felts which can be range checked, e.g. `assert_le` will not crash.
```
1) Maybe it's worth to mention this is a keyword only argument.
2) Why does `assert_le` crashes in the first place? Shouldn't this option be enabled by default? I think this parameter needs more detailed description.

mkaput: 1. Done
2. Not all felt values are allowed to be range checked, the limit is determined by the layout of the Cairo program (which as far as I understand, is the same for all "civil" Starknet contracts, hence the constant is hardcoded). Range checks are allowed only with felts from range `[0, RANGE_CHECK_BOUND)`. I think this comes from the fact that they are doing some bitwise ops in the builtin and these operations must not overflow. Whatever is default is a very philosophical question. I opted for maintaining compatibility with existing behaviour (`strategy.felts()` behaves the same as fuzzer behaved when it was released).

kasperski95: IMO `comparable` is a misleading name. It could be understood felts provided by fuzzer can't be compared by default.

kasperski95 (approved):

PR #865, opened 2022-09-21
Refresh fuzzing tutorial

### LINKED ISSUE #860: Revisit fuzzing tutorial to include latest features

mkaput:
fix #860

### DIFF
--- a/website/docs/tutorials/07-testing/03-fuzzing/README.md
+++ b/website/docs/tutorials/07-testing/03-fuzzing/README.md
@@ -48,16 +48,6 @@ func withdraw{
     balance.write(new_res);
     return ();
 }
-
-@constructor
-func constructor{
-    syscall_ptr: felt*,
-    pedersen_ptr: HashBuiltin*,
-    range_check_ptr
-}() {
-    balance.write(10000);
-    return ();
-}
 ```
 
 ### Unit testing
@@ -69,6 +59,16 @@ testing for:
 %lang starknet
 from src.main import balance, withdraw
 from starkware.cairo.common.cairo_builtins import HashBuiltin
+
+@external
+func setup_withdraw{
+    syscall_ptr: felt*,
+    pedersen_ptr: HashBuiltin*,
+    range_check_ptr
+}() {
+    balance.write(10000);
+    return ();
+}
 
 @external
 func test_withdraw{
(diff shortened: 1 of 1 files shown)


website/docs/tutorials/07-testing/03-fuzzing/README.md
```diff
@@ -139,96 +135,138 @@ func test_withdraw{
...
+```
+
+### Fixing the bug
+
+The test fails because `amount` has `felt` type, so it can overflow when subtracting.
+In particular, it is certain, that the overflow will happen if you try to
+withdraw `FIELD_PRIME - 1` (which is the number fuzzer found!).
+Although this bug should be fixed within the contract, for the purpose of this tutorial we will do
+it differently:
+we will instruct the fuzzer to avoid numbers outside `range_check` builtin boundary.
```
MaksymilianDemitraszek, line 232:
```suggestion
we will instruct the fuzzer to avoid numbers outside of `range_check` builtin boundary.
```

website/docs/tutorials/07-testing/03-fuzzing/README.md
```diff
@@ -139,96 +135,138 @@ func test_withdraw{
...
+
+Fuzzer input parameters are selected according to a _fuzzing strategy_ associated with each
+parameter.
+Protostar offers various strategies tailored for specific use cases, check out
+the [Strategies](./strategies.md) page for more information.
+Associating a fuzzing strategy to a parameter is done via the [`given`](../02-cheatcodes/given.md)
```
MaksymilianDemitraszek, line 238:
```suggestion
Associating a fuzzing strategy to a parameter is done using the [`given`](../02-cheatcodes/given.md)
```

website/docs/tutorials/07-testing/03-fuzzing/README.md
```diff
@@ -139,96 +135,138 @@ func test_withdraw{
...
+cheatcode, which is only available within setup cases.
```
MaksymilianDemitraszek, line 239: can we link setup cases?

MaksymilianDemitraszek (changes requested):

website/docs/tutorials/07-testing/03-fuzzing/README.md
mkaput: Done.

MaksymilianDemitraszek (approved):

PR #866, opened 2022-09-21
Fixed `expect_revert` error messages

kasperski95:


protostar/testing/cheatcodes/expect_revert_cheatcode.py
```diff
@@ -18,6 +18,13 @@ def __init__(self):
 
     def expect_revert(self, expected_error: RevertableException) -> Callable[[], None]:
         if self._expected_error is not None:
+            if (
+                self._expected_error.error_type is None
+                and not self._expected_error.error_messages
+            ):
+                raise SimpleReportedException(
```
mkaput, line 25: CheatcodeException is more suitable here IMHO

tests/integration/protostar_fixture.py
```diff
@@ -213,6 +217,46 @@ def create_migration_file(
         )
         return file_path
 
+    async def run_test_case(self, test_case_content: str):
```
mkaput, line 220: I don't like this method in comparison with existing flow because of following reasons (in such degree that I am hesistant to approve it):
1. It templates too much of the source file, not allowing passing setup hooks, controlling builtins, parameters, names.
2. It's less performant because it prevents bundling multiple related test cases together. This technique amortizes times like compilation.

mkaput (changes requested):

protostar/testing/cheatcodes/expect_revert_cheatcode.py
kasperski95: Done

tests/integration/protostar_fixture.py
kasperski95: Removed unnecessary integration test.

MaksymilianDemitraszek (approved):

protostar/testing/cheatcodes/expect_revert_cheatcode.py
```diff
@@ -37,9 +38,7 @@ def stop_expecting_revert():
                 "The callback returned by the `expect_revert` is deprecated."
             )
             if self._expected_error is not None:
-                raise SimpleReportedException(
-                    "Expected a transaction to be reverted before cancelling expect_revert"
-                )
+                raise SimpleReportedException("Expected a transaction to be reverted")
```
mkaput, line 41:
```suggestion
                raise SimpleReportedException("Expected a transaction to be reverted.")
```

protostar/testing/cheatcodes/expect_revert_cheatcode.py
```diff
@@ -22,12 +22,13 @@ def expect_revert(self, expected_error: RevertableException) -> Callable[[], Non
                 self._expected_error.error_type is None
                 and not self._expected_error.error_messages
             ):
-                raise SimpleReportedException(
-                    "Protostar is already expecting an exception."
+                raise CheatcodeException(
+                    "expect_revert", "Protostar is already expecting an exception"
```
mkaput, line 26:
```suggestion
                    "expect_revert", "Protostar is already expecting an exception."
```

mkaput (changes requested):

tests/integration/protostar_fixture.py
mkaput: Currently we were inspecting errors in integration tests by looking up for a `TestResult` in the `TestingSummary` returned from the fixture.

> This approach allows running specific test cases, which makes debugging easier.

This is a fair point about a feature that we are missing, but retrofitting it into existing method is easy.

> Such functionality can be added later.
> 
> `protostar.run_test_case("...", setup_fn="...", setup="...", params={"a": "felt"}, use_syscall_ptr=True)`

Doesn't this look like too much magic for you? I would definitely prefer to write full test in Cairo instead of relying on wild kwargs here.

kasperski95: Maybe `summary` should be a module-scoped fixture?

mkaput: `summary` is a result of running the test itself. I don't see how this can be a fixture.

kasperski95: I've just pushed the code with this approach.

protostar/commands/test/test_command.py
```diff
@@ -186,7 +186,7 @@ async def test(
         self._log_test_collector_result(test_collector_result)
 
         testing_summary = TestingSummary(
-            case_results=test_collector_result.broken_test_suites,  # type: ignore | pyright bug?
+            test_results=test_collector_result.broken_test_suites,  # type: ignore | pyright bug?
```
mkaput, line 189: hmm does the type error still exist?

tests/integration/cheatcodes/expect_revert/expect_revert_test.py
```diff
@@ -32,35 +37,21 @@ async def test_expect_revert(run_cairo_test_runner: RunCairoTestRunnerFixture):
...
-        %}
-        """
+    (_, formatted_protostar_test_result) = get_protostar_test_case_result(
+        testing_summary,
+        "test_already_expecting_error_message_when_no_arguments_were_provided",
     )
 
-    assert "matching the following error" not in formatted_test_result
-    assert "Protostar is already expecting an exception." in formatted_test_result
+    assert "matching the following error" not in formatted_protostar_test_result
```
mkaput, line 54: what do you think about asserting fields of `BrokenTestCaseResult` (or so) here instead of looking for substrings in formatted representation? 

This way this test could be more precise, because such assertion is much stronger.

And the infra can be simpler, because without formatting phase, the `get_protostar_test_case_result` function can be easily replaced by `[]` operator implementation in TestingSummary (which has logical sense and thus I would love to see there).

kasperski95: > what do you think about asserting fields of BrokenTestCaseResult (or so) here instead of looking for substrings in formatted representation?

This way this test could be more precise, because such assertion is much stronger.

I'd avoid testing internals. Checking test result type is done in test case above that uses `assert_cairo_test_cases`. 

> And the infra can be simpler, because without formatting phase, the get_protostar_test_case_result function can be easily replaced by [] operator implementation in TestingSummary (which has logical sense and thus I would love to see there).

- [ ] use `[]` operator

mkaput: > I'd avoid testing internals.

That's a good point, but I think `TestResult` structures are not so much internal aspects, given that we treat TestingSummary as tested entity and both are coupled.

kasperski95: Testing `test_result.exception.message` doesn't guarantee printing the error to the user. Testing formatted output makes the test more useful. Such tests can be treated as [Subcutaneous Tests](https://martinfowler.com/bliki/SubcutaneousTest.html).

Checking the formatted output also checks the integration between `protostar.testing` and `protostar.commands`.

mkaput: I think that if you want to properly ensure that the feedback is done correctly, you would have to migrate this test to E2E. Right now, you assume that stringified result is printed.

On the other hand the following suite of tests is more lightweight:
1. Test that proper TestResult is returned for each case imaginable (that's the kind I push for here).
2. Test that abstractly constructed TestResults are stringified properly (that's part you try to mix in here).
3. Test on one example, in E2E suite, that indeed stringified representation is printed.

Unless I am very stupid, this set of tests should **definitely mathematically cover** all possible combinations of behaviours assuming sane implementation (no random, wild branches etc. crafted deliberately to break tests in the code).

kasperski95: > Test that proper TestResult is returned for each case imaginable (that's the kind I push for here).

This is already handled by the test case above in `assert_cairo_test_cases`. I can add an extra assert in the new test case, but it's pointless.

> Test that abstractly constructed TestResults are stringified properly (that's part you try to mix in here).

This is done by "my" test case.

> Test on one example, in E2E suite, that indeed stringified representation is printed.

We have already plenty of E2E tests checking that formatted test results are indeed printed.

In summary, existing tests do all what you want.

---

> I think that if you want to properly ensure that the feedback is done correctly, you would have to migrate this test to E2E. Right now, you assume that stringified result is printed.

This remark conflicts with your suggestion about the suite of tests.

kasperski95: Overloaded operator. Added an extra assert.

kasperski95: Marek OOO. Resolving conversation.

protostar/commands/test/test_command.py
kasperski95: It still exists.

PR #881, opened 2022-09-22
Deprecate unsigned declare

kasperski95:


tests/conftest.py
```diff
@@ -74,6 +80,45 @@ def signing_credentials_fixture() -> Credentials:  # The same account is generat
...
+        for devnet_account_dict in devnet_account_dicts
+    ]
+
+
+@pytest.fixture(name="alice_devnet_account")
+def alice_devnet_account(
+    devnet_accounts: list[DevnetAccount], monkeypatch: pytest.MonkeyPatch
+) -> DevnetAccount:
+    alice_devnet_account = devnet_accounts[0]
+    monkeypatch.setenv(PRIVATE_KEY_ENV_VAR_NAME, alice_devnet_account.private_key)
```
Arcticae, line 118: this is an unwanted side effect of the fixture

tests/integration/deploying/declare_command_test.py
```diff
@@ -26,14 +27,12 @@ def compiled_contract_path_fixture() -> Path:
 async def test_declaring_contract(
     protostar: ProtostarFixture,
     devnet_gateway_url: str,
-    monkeypatch: pytest.MonkeyPatch,
+    alice_devnet_account: DevnetAccount,
```
Arcticae, line 30: why not just `devnet_account` ? or `predefined_account`

Arcticae (changes requested):

protostar/starknet_gateway/gateway_facade.py
```diff
@@ -139,65 +138,99 @@ def _prepare_constructor_inputs(
         validate_cairo_inputs(abi, cairo_inputs)
         return cairo_inputs
 
-    # pylint: disable=unused-argument
     async def declare(
```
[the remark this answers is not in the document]
kasperski95: Invoke doesn't follow such convention, and `declare_v0` will be removed in the near future.

tests/conftest.py
kasperski95: The same can be said about running the devnet when using `devnet_gateway_url` fixture. I'll make it explicit.

tests/integration/deploying/declare_command_test.py
kasperski95: Renamed in [7086b29](https://github.com/software-mansion/protostar/pull/881/commits/7086b29d3a818d84585fa0c1dd81e506c5ee43cf)

tests/conftest.py
kasperski95: Done in [8bebc31](https://github.com/software-mansion/protostar/pull/881/commits/8bebc31924e5c8c24462a97146fc7ca29e401bef)

protostar/starknet_gateway/gateway_facade.py
```diff
@@ -139,65 +138,99 @@ def _prepare_constructor_inputs(
...
+            self._project_root_path / compiled_contract_path
+        )
+        account_client = await self._create_account_client(
+            account_address=account_address, signer=signer
+        )
+        declare_tx = await self._create_declare_tx_v1(
+            compiled_contract=compiled_contract,
+            account_client=account_client,
+            auto_estimate_fee=True,
+            max_fee=None,
```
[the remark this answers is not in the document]
kasperski95: Done

protostar/migrator/cheatcodes/migrator_declare_cheatcode.py
```diff
@@ -29,11 +29,33 @@ def __call__(
         ...
 
 
+class DeclareCheatcodeNetworkConfig(CheatcodeNetworkConfig):
```
Arcticae, line 32: Note: We're going to pay for invoke in the future as well, so that's not going to be specific for this cheatcode only. Keep that in mind

website/docs/tutorials/06-deploying/02-migrations/declare.md
```diff
@@ -14,7 +14,16 @@ class DeclaredContract:
...
 - `contract_path_str` — A path to the **compiled** contract or a [contract name](../../compiling#contract-name). If you provide a contract name, Protostar will compile the contract to the migration output directory.
-- `config` — A keyword only argument that allows passing [network configuration](../03-network-config.md).
+- `config` — A keyword only argument that allows configuring transaction and network parameters.
+
+## `DeclareCheatcodeNetworkConfig`
+```python
+class DeclareCheatcodeNetworkConfig(CheatcodeNetworkConfig):
+    max_fee: NotRequired[Wei]
+```
+- `max_fee` — The maximum fee that the sender is willing to pay for the transaction.
```
Arcticae, line 24: I wonder if we should allow for silently interpreting lack of max_fee as `auto_estimate`. I know that requiring the user to pass `auto_estimate` explicitly can make the UX slightly worse, but user should be aware of the mechanism he is using in this case, since it can incur large costs depending on the gas price. WDYT?

Arcticae: It could also be made interactive, like: 
`This tx will cost you X wei, do you want to proceed?` (if the terminal is interactive) 
Else we should probably require one of `max_fee` or `auto_estimate`

Arcticae (changes requested):

website/docs/tutorials/06-deploying/02-migrations/declare.md
kasperski95: I think we should avoid interactive terminals if possible.

protostar/migrator/cheatcodes/migrator_declare_cheatcode.py
kasperski95: Currently, it is specific to this cheatcode hence I added it here. When it's shared, then we can easily refactor this code.
(Btw. I think NetworkConfig is a bad name. Maybe TxConfig would be better.)

website/docs/tutorials/06-deploying/02-migrations/declare.md
kasperski95: <Internal discussion> -> `max_fee="auto"`

kasperski95: Added `max_fee="auto"` support.

Arcticae (approved):
LGTM

Issue #580, opened 2022-08-04
Add strategy.short_strings()


mkaput:
Add a new fuzzing strategy based on https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.text which yields Cairo short strings.

PR #886, opened 2022-09-26
Add short string strategy

karol-bisztyga:
Adding the short-string strategy plus tests.

#580

### DIFF
--- a/protostar/testing/fuzzing/strategies/short_strings.py
+++ b/protostar/testing/fuzzing/strategies/short_strings.py
@@ -1,0 +1,24 @@
+import string
+
+from hypothesis.strategies import SearchStrategy, text
+from starkware.cairo.lang.compiler.ast.cairo_types import CairoType, TypeFelt
+
+from protostar.testing.fuzzing.exceptions import SearchStrategyBuildError
+from protostar.testing.fuzzing.strategy_descriptor import StrategyDescriptor
+
+
+class ShortStringsStrategyDescriptor(StrategyDescriptor):
+
+    def build_strategy(self, cairo_type: CairoType) -> SearchStrategy[int]:
+        if not isinstance(cairo_type, TypeFelt):
+            raise SearchStrategyBuildError(
+                "Strategy 'short strings' can only be applied to felt parameters."
+            )
+
+        def string_to_bytes_int(val: str) -> int:
+            if not val:
+                return 0
+            # pylint: disable=consider-using-f-string
+            return int('0x' + ''.join('{:02x}'.format(ord(c)) for c in val), 16)
+
+        return text(alphabet=list(string.ascii_lowercase), min_size=0, max_size=31).map(string_to_bytes_int)
--- a/protostar/testing/fuzzing/strategies/short_strings.py
+++ b/protostar/testing/fuzzing/strategies/short_strings.py
@@ -8,7 +8,6 @@ from protostar.testing.fuzzing.strategy_descriptor import StrategyDescriptor
 
 
 class ShortStringsStrategyDescriptor(StrategyDescriptor):
-
     def build_strategy(self, cairo_type: CairoType) -> SearchStrategy[int]:
         if not isinstance(cairo_type, TypeFelt):
             raise SearchStrategyBuildError(
@@ -19,6 +18,8 @@ class ShortStringsStrategyDescriptor(StrategyDescriptor):
             if not val:
                 return 0
             # pylint: disable=consider-using-f-string
-            return int('0x' + ''.join('{:02x}'.format(ord(c)) for c in val), 16)
+            return int("0x" + "".join("{:02x}".format(ord(c)) for c in val), 16)
 
-        return text(alphabet=list(string.ascii_lowercase), min_size=0, max_size=31).map(string_to_bytes_int)
+        return text(alphabet=list(string.ascii_lowercase), min_size=0, max_size=31).map(
+            string_to_bytes_int
+        )
--- a/protostar/testing/fuzzing/strategies/short_strings.py
+++ b/protostar/testing/fuzzing/strategies/short_strings.py
@@ -7,18 +7,19 @@ from protostar.testing.fuzzing.exceptions import SearchStrategyBuildError
 from protostar.testing.fuzzing.strategy_descriptor import StrategyDescriptor
 
 
+def string_to_bytes_int(val: str) -> int:
+    if not val:
+        return 0
+    # pylint: disable=consider-using-f-string
+    return int("0x" + "".join("{:02x}".format(ord(c)) for c in val), 16)
+
+
 class ShortStringsStrategyDescriptor(StrategyDescriptor):
     def build_strategy(self, cairo_type: CairoType) -> SearchStrategy[int]:
         if not isinstance(cairo_type, TypeFelt):
             raise SearchStrategyBuildError(
                 "Strategy 'short strings' can only be applied to felt parameters."
             )
-
-        def string_to_bytes_int(val: str) -> int:
-            if not val:
-                return 0
-            # pylint: disable=consider-using-f-string
-            return int("0x" + "".join("{:02x}".format(ord(c)) for c in val), 16)
 
         return text(alphabet=list(string.ascii_lowercase), min_size=0, max_size=31).map(
             string_to_bytes_int
--- a/protostar/testing/fuzzing/strategies/short_strings.py
+++ b/protostar/testing/fuzzing/strategies/short_strings.py
@@ -2,16 +2,10 @@ import string
 
 from hypothesis.strategies import SearchStrategy, text
 from starkware.cairo.lang.compiler.ast.cairo_types import CairoType, TypeFelt
+from starkware.python.utils import from_bytes
 
 from protostar.testing.fuzzing.exceptions import SearchStrategyBuildError
 from protostar.testing.fuzzing.strategy_descriptor import StrategyDescriptor
-
-
-def string_to_bytes_int(val: str) -> int:
-    if not val:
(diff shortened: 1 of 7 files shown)


tests/integration/fuzzing_strategies/short_strings_test.cairo
```diff
@@ -0,0 +1,13 @@
...
+func setup_correct_short_string() {
+    %{ given(value=strategy.short_strings()) %}
+    return ();
+}
+
+@external
+func test_correct_short_string{syscall_ptr: felt*, range_check_ptr}(value: felt) {
+    %{ assert 0 <= ids.value %}
+    return ();
+}
```
kasperski95, line 13: This test doesn't check if the value is less than the short string max possible value.

- [ ] add assert checking the short string value is in the expected range

kasperski95 (commented):
- [x] add docs (website/docs/tutorials/...)

tests/integration/fuzzing_strategies/short_strings_test.cairo
karol-bisztyga: Done in [276916c](https://github.com/software-mansion/protostar/pull/886/commits/276916c3f08b4704d6f5cc30a21b1e632d221abe)

tests/integration/fuzzing_strategies/short_strings_test.cairo
```diff
@@ -0,0 +1,13 @@
+%lang starknet
+
+@external
+func setup_correct_short_string() {
+    %{ given(value=strategy.short_strings()) %}
+    return ();
+}
```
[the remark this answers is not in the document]
karol-bisztyga: Honestly, I think this is very hard to do (but I may be missing something ofc). The string strategy seems to not work very well when we set the number of tries to some small amount and at the same time, the min size is also small. Mostly, we'll get one-char strings. I think we could do two tests, one with a small min size and one with a really big size, in order to enforce the big values (let's say, one test 0-31 (that's going to contain small values mostly) and the other 29-31 (containing big values)). Then, we'll not need a lot of tries in order to check values that are close to borders, what do you think?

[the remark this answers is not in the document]
kasperski95: 20-30 examples should be a good trade-off. Alternatively, you can try running this strategy as a Python test.

karol-bisztyga:
I added a docs entry in [423f29e](https://github.com/software-mansion/protostar/pull/886/commits/423f29e33fa90c38229c3c587a91408e397ce33c), if there should be anything else added somewhere, please, let me know.

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -75,6 +75,20 @@ Fuzzer picks integers from provided range and then converts them to felts.
...
 
+### `strategy.short_strings`
+```python
+def text(
+    alphabet: Union[Sequence[str], SearchStrategy[str]] = characters(
+        blacklist_categories=("Cs",)
+    ),
+    min_size: int = 0,
+    max_size: Optional[int] = None,
+) -> Strategy:
```
kasperski95, line 86:
```suggestion
def short_strings() -> Strategy:
```

The current implementation doesn't expose any configuration parameters. They can be added in future PRs.

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -75,6 +75,20 @@ Fuzzer picks integers from provided range and then converts them to felts.
...
+```
+
+Generates strings with ascii lowercase characters of length that passes the condition `0 <= length <= 31`.
```
kasperski95, line 89: The default strategy configuration shouldn't be limited to lowercase characters. The alphabet should consist of characters supported by short strings.

Exposing alphabet parameter can be done in a different PR.

kasperski95 (changes requested):

protostar/testing/fuzzing/strategies/short_strings.py
```diff
@@ -0,0 +1,26 @@
...
+
+from protostar.testing.fuzzing.exceptions import SearchStrategyBuildError
+from protostar.testing.fuzzing.strategy_descriptor import StrategyDescriptor
+
+
+def string_to_bytes_int(val: str) -> int:
+    if not val:
+        return 0
+    # pylint: disable=consider-using-f-string
+    return int("0x" + "".join("{:02x}".format(ord(c)) for c in val), 16)
```
MaksymilianDemitraszek, line 14: Cannot we use `from starkware.python.utils import from_bytes` instead?

protostar/testing/fuzzing/strategies/short_strings.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ShortStringsStrategyDescriptor(StrategyDescriptor):
+    def build_strategy(self, cairo_type: CairoType) -> SearchStrategy[int]:
+        if not isinstance(cairo_type, TypeFelt):
+            raise SearchStrategyBuildError(
+                "Strategy 'short strings' can only be applied to felt parameters."
+            )
+
+        return text(alphabet=list(string.ascii_lowercase), min_size=0, max_size=31).map(
```
MaksymilianDemitraszek, line 24: I would use all of the ascii here not only lowercase

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -75,6 +75,20 @@ Fuzzer picks integers from provided range and then converts them to felts.
...
+Max size `31` comes from [the docs](https://www.cairo-lang.org/docs/how_cairo_works/consts.html#short-string-literals).
```
MaksymilianDemitraszek, line 90: Maybe callout ::info ?

MaksymilianDemitraszek (changes requested):
Very good job 😄 I left couple of questions in the comments

protostar/testing/fuzzing/strategies/short_strings.py
karol-bisztyga: Yes, sure, how about https://docs.python.org/3/library/string.html#string.printable?

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
karol-bisztyga: I don't think it is specified what signs are allowed exactly. Maybe we could go with https://docs.python.org/3/library/string.html#string.printable

kasperski95: `string.printable` should be ok

protostar/testing/fuzzing/strategies/short_strings.py
karol-bisztyga: Nice, I didn't know about that, thanks.

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
karol-bisztyga: fixed in [ab28a1a](https://github.com/software-mansion/protostar/pull/886/commits/ab28a1ad8001eadf597047acc3e9553478ca918b)

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
karol-bisztyga: fixed in [ab28a1a](https://github.com/software-mansion/protostar/pull/886/commits/ab28a1ad8001eadf597047acc3e9553478ca918b), I removed the word `lowercase`

protostar/testing/fuzzing/strategies/short_strings.py
karol-bisztyga: Done in [1a1dc66](https://github.com/software-mansion/protostar/pull/886/commits/1a1dc66b90aed39d3e2737a1131b7e819748495e)

protostar/testing/fuzzing/strategies/short_strings.py
karol-bisztyga: Done in [1a1dc66](https://github.com/software-mansion/protostar/pull/886/commits/1a1dc66b90aed39d3e2737a1131b7e819748495e)

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
karol-bisztyga: Added in [ab28a1a](https://github.com/software-mansion/protostar/pull/886/commits/ab28a1ad8001eadf597047acc3e9553478ca918b)

MaksymilianDemitraszek (approved):

website/docs/tutorials/07-testing/03-fuzzing/strategies.md
```diff
@@ -75,6 +75,17 @@ Fuzzer picks integers from provided range and then converts them to felts.
 If `min_value` is not `None` then all values will be greater than or equal to `min_value`,
 and if `max_value` is not `None` then all values will be less than or equal to `max_value`.
 
+### `strategy.short_strings`
+```python
+def short_strings() -> Strategy:
+```
+
+Generates strings with ascii characters of length that passes the condition `0 <= length <= 31`.
```
kasperski95, line 83:
```suggestion
Generates strings with ASCII characters of length that passes the condition `0 <= length <= 31`.
```

karol-bisztyga: [ASCII instead of ascii in docs](https://github.com/software-mansion/protostar/pull/886/commits/5f354b9bf0a1801235d9f031e84ce461f3dc3bfc)

PR #887, opened 2022-09-26
Redesigning `protostar.toml` 7 — `ConfigurationFileFactory`

kasperski95:
`ConfigurationFileFactory` builds the proper version of the `ConfigurationFile`.

This PR doesn't change the Protostar logic yet. I'll merge this PR without review on 27 September at 3 pm.

### DIFF
--- a/protostar/configuration_file/configuration_file_factory.py
+++ b/protostar/configuration_file/configuration_file_factory.py
@@ -1,6 +1,8 @@
 from pathlib import Path
 from typing import Optional
 
+from protostar.protostar_exception import ProtostarException
+
 from .configuration_file import ConfigurationFile
 from .configuration_file_v1 import ConfigurationFileV1
 from .configuration_file_v2 import ConfigurationFileV2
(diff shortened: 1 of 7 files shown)


protostar/configuration_file/configuration_file_factory.py
```diff
@@ -0,0 +1,85 @@
...
+
+        raise ProtostarException(
+            f"{protostar_toml_path} must specify `protostar-version` for example:\n"
+            """
+            [project]
+            protostar-version = ?.?.?
+            """,
+        )
+
+    def _create_configuration_toml_v2(
```
MaksymilianDemitraszek, line 44: Shouldn't we have `AbstractConfigFactory`, `V1ConfigFactory` and `V2ConfigFactory`, imo factory shouldn't be an entity responsible for making decisions.

MaksymilianDemitraszek: My point is I wouldn't name it factory if you do not have a interchangeable family of those

MaksymilianDemitraszek (approved):
LGTM

protostar/configuration_file/configuration_file_factory.py
kasperski95: https://www.oreilly.com/library/view/head-first-design/0596007124/ch04.html
Chapter "Building a simple pizza factory"

In that chapter, the factory decides what Pizza needs to be build based on the provided type. In this case, the factory decides which configuration file should be created.

> My point is I wouldn't name it factory if you do not have a interchangeable family of those

Factory and Abstract Factory are two different things. In the link above, they use the Factory suffix even if the Factory is not abstract.

Issue #712, opened 2022-08-31
Add an E2E test that confirms that testing seed is different on each Protostar run


mkaput:
Do we have a test checking if the seed changes with every run?

_Originally posted by @kasperski95 in https://github.com/software-mansion/protostar/pull/710#discussion_r959529881_

PR #888, opened 2022-09-26
Add test for checking if the seed changes every time

karol-bisztyga:
Adding a test that checks whether the seed is different every time we run the fuzzer.

#712

### DIFF
--- a/tests/e2e/test_fuzzing.py
+++ b/tests/e2e/test_fuzzing.py
@@ -41,3 +41,24 @@ c = 0
 """
         in result
     )
+
+
+@pytest.mark.usefixtures("init")
+def test_fuzzing_changing_seed(protostar, copy_fixture):
+    seeds = []
+    copy_fixture("test_fuzz_single.cairo", "./tests")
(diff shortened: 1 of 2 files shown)


tests/e2e/test_fuzzing.py
```diff
@@ -41,3 +41,24 @@ def test_fuzzing(protostar, copy_fixture):
...
         in result
     )
+
+
+@pytest.mark.usefixtures("init")
+def test_fuzzing_changing_seed(protostar, copy_fixture):
+    seeds = []
+    copy_fixture("test_fuzz_single.cairo", "./tests")
+
+    for i in range(0, 5):
```
karol-bisztyga, line 51: I'm not sure if this number is good, the tests are pretty slow (in my local env, one test takes ~6s, so here we'll have 30s)

kasperski95: Running tests twice should be enough.

kasperski95 (approved):

Issue #693, opened 2022-08-29
CLI option to run last failed


milancermak:
pytest has a ton of useful CLI options. One of my most used ones is `--lf, --last-failed` to "rerun only the tests that failed at the last run (or all if none failed)". It would be lovely if protostar could add this as well.

karol-bisztyga:
#895 merged, closing

PR #889, opened 2022-09-26
Add possibility to invoke pytest with args

karol-bisztyga:
In the issue, there was a request to add a possibility to run the tests based on `pytest` with the option `--lf, --last-failed`, but I think a more generic approach could be also okay.

To apply any pytest arguments, you can just do something like this:
```
PYTEST_ARGS="--lf -q -s" poe test_unit
```
So it looks like
```
PYTEST_ARGS="[args]" poe [command name]
```
If you don't provide `PYTEST_ARGS`, it will run the test without pytest arguments.

#693

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,19 +87,19 @@ type_check = "pyright"
 update_cli_docs = "python ./scripts/generate_reference_docs.py"
 
 [tool.poe.tasks.benchmark]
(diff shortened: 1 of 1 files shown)


kasperski95 (changes requested):
The issue was about adding this option to the Protostar e.g.
```
protostar test --lf
```

karol-bisztyga:
Sorry, I misunderstood, closing in favor of https://github.com/software-mansion/protostar/pull/895

PR #892, opened 2022-09-26
Redesigning `protostar.toml` 8 — `CommandNamesProvider`

kasperski95:
To transform `ConfigurationFile` V1 into V2, command names must be known, and they must be the same as command names provided to `ProtostarCLI`.

Note: `ConfigurationFile` isn't integrated with the Protostar yet. Integration will be done in the last PR. If the PR is not reviewed within 24 hours, I am going to merge it 27 September at 6 pm.

### DIFF
--- a/protostar/cli/command_names_delayed_provider.py
+++ b/protostar/cli/command_names_delayed_provider.py
@@ -1,0 +1,18 @@
+from typing import Optional
+
+from protostar.configuration_file import CommandNamesProvider
+
+
+class CommandNamesDelayedProvider(CommandNamesProvider):
+    def __init__(self) -> None:
+        super().__init__()
+        self._command_names_provider: Optional[CommandNamesProvider] = None
+
+    def set_command_names_provider(
+        self, command_names_provider: CommandNamesProvider
+    ) -> None:
+        self._command_names_provider = command_names_provider
+
+    def get_command_names(self) -> list[str]:
+        assert self._command_names_provider is not None
+        return self._command_names_provider.get_command_names()
--- a/protostar/cli/command_names_delayed_provider.py
+++ b/protostar/cli/command_names_delayed_provider.py
@@ -1,15 +1,15 @@
 from typing import Optional
 
-from protostar.configuration_file import CommandNamesProvider
+from protostar.configuration_file import CommandNamesProviderProtocol
 
 
-class CommandNamesDelayedProvider(CommandNamesProvider):
+class CommandNamesDelayedProvider(CommandNamesProviderProtocol):
     def __init__(self) -> None:
         super().__init__()
-        self._command_names_provider: Optional[CommandNamesProvider] = None
+        self._command_names_provider: Optional[CommandNamesProviderProtocol] = None
 
     def set_command_names_provider(
-        self, command_names_provider: CommandNamesProvider
+        self, command_names_provider: CommandNamesProviderProtocol
     ) -> None:
         self._command_names_provider = command_names_provider
 
--- a/protostar/cli/command_names_delayed_provider_test.py
+++ b/protostar/cli/command_names_delayed_provider_test.py
@@ -1,0 +1,26 @@
+import pytest
+
+from protostar.configuration_file import CommandNamesProvider
+
+from .command_names_delayed_provider import CommandNamesDelayedProvider
+
+
+class CommandNamesProviderDouble(CommandNamesProvider):
+    def get_command_names(self) -> list[str]:
+        return ["foo", "bar"]
+
+
+def test_retrieving_command_names():
+    command_names_aggregator = CommandNamesDelayedProvider()
+    command_names_aggregator.set_command_names_provider(CommandNamesProviderDouble())
+
+    result = command_names_aggregator.get_command_names()
+
+    assert result == ["foo", "bar"]
+
+
+def test_retrieving_command_names_failure_when_actual_provider_is_not_set():
+    command_names_aggregator = CommandNamesDelayedProvider()
+
+    with pytest.raises(AssertionError):
+        command_names_aggregator.get_command_names()
--- a/protostar/cli/command_names_delayed_provider_test.py
+++ b/protostar/cli/command_names_delayed_provider_test.py
@@ -1,18 +1,18 @@
 import pytest
 
-from protostar.configuration_file import CommandNamesProvider
+from protostar.configuration_file import CommandNamesProviderProtocol
 
 from .command_names_delayed_provider import CommandNamesDelayedProvider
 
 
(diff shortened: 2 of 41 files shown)


protostar/configuration_file/configuration_file_v2_test.py
```diff
@@ -16,11 +16,20 @@
...
+from .configuration_file_v1 import (
+    CommandNamesProvider,
+    ConfigurationFileV1,
+    ConfigurationFileV1Model,
+)
 from .configuration_file_v2 import ConfigurationFileV2, ConfigurationFileV2Model
 from .configuration_legacy_toml_interpreter import ConfigurationLegacyTOMLInterpreter
 
 
+class CommandNamesProviderDouble(CommandNamesProvider):
```
MaksymilianDemitraszek, line 28: What does the name mean?

protostar/cli/command_names_delayed_provider_test.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from protostar.configuration_file import CommandNamesProvider
+
+from .command_names_delayed_provider import CommandNamesDelayedProvider
+
+
+class CommandNamesProviderDouble(CommandNamesProvider):
```
MaksymilianDemitraszek, line 8: Don't we have three classes with the same name? We cannot do that, it is *very* misleading (for grep users is somewhat hardcore)

protostar/protostar_cli.py
```diff
@@ -24,7 +25,7 @@ def _apply_pythonpath():
         sys.path.extend(split_paths)
 
 
-class ProtostarCLI(CLIApp):
+class ProtostarCLI(CLIApp, CommandNamesProvider):
```
MaksymilianDemitraszek, line 28: I think it would be better to keep provider as a property not as a protocol. It is very hard to reason about.

protostar/configuration_file/configuration_file_v1.py
```diff
@@ -27,25 +27,24 @@ class ConfigurationFileV1Model:
     profile_name_to_shared_command_config: dict[ProfileName, CommandConfig]
 
 
+class CommandNamesProvider(Protocol):
```
MaksymilianDemitraszek, line 30: I would make it a class, not protocol (Double inheritance is hard to reason about)

protostar/cli/command_names_delayed_provider.py
```diff
@@ -0,0 +1,18 @@
+from typing import Optional
+
+from protostar.configuration_file import CommandNamesProvider
+
+
+class CommandNamesDelayedProvider(CommandNamesProvider):
```
MaksymilianDemitraszek, line 6: Can you please explain the purpose of this class?

MaksymilianDemitraszek (changes requested):

protostar/cli/command_names_delayed_provider_test.py
MaksymilianDemitraszek: You could parametrize it and put it in conftest

protostar/configuration_file/configuration_file_v2_test.py
kasperski95: https://martinfowler.com/bliki/TestDouble.html

protostar/cli/command_names_delayed_provider_test.py
kasperski95: I'd have to pollute global conftest.py, but that fixture would be used only by two components.

protostar/protostar_cli.py
kasperski95: `CommandNamesProvider` property would have to operate on internals of `ProtostarCLI` which breaks the encapsulation.

> It is hard to reason about.

This is a delegate pattern (I need to change `CommandNamesProvider` to `CommandNamesProviderDelegate`)
https://dilloncodes.com/delegate-pattern-in-swift

protostar/configuration_file/configuration_file_v1.py
kasperski95: The class would have to know how to retrieve data from `ProtostarCLI`. It creates a cyclic dependency between `protostar.cli` and `protostar.configuration_file`. 

> (Double inheritance is hard to reason about)

In my mind:
`class CommandNamesProviderDelegate(Protocol)` == `interface CommandNamesProviderDelegate`

protostar/cli/command_names_delayed_provider.py
kasperski95: To transform ConfigurationFile V1 into V2, command names must be known, and they must be the same as command names provided to ProtostarCLI.

This class also breaks the dependency cycle:
`ConfigurationFile <- Commands <- ProtostarCLI <- ConfigurationFIle`

protostar/configuration_file/configuration_file_v2_test.py
MaksymilianDemitraszek: I would name it `Stub` instead, name `Dobule` does not seem very popular one

MaksymilianDemitraszek: I thought it has some property of doing something twice/ returning two commands

protostar/cli/command_names_delayed_provider_test.py
MaksymilianDemitraszek: If it's only two you can resolve it, If it's three I would consider polluting

kasperski95: Hmm, I can export the Double (Stub) from the `protostar.configuration_file` component instead of putting it to the global conftest. Maybe in the future. Resolving.

protostar/configuration_file/configuration_file_v2_test.py
kasperski95: The name is also used in NestJS docs.
https://docs.nestjs.com/fundamentals/testing#unit-testing

Anyway, I can rename it.

protostar/protostar_cli.py
MaksymilianDemitraszek: I think new name is worse, Delegate pattern is something that not every python programmer is familliar with

protostar/configuration_file/configuration_file_v1.py
MaksymilianDemitraszek: Ok let's leave it like this, but please do not rename it to Delegate, you can consider `Interface` or leaving it as it is

protostar/configuration_file/configuration_file_v2_test.py
MaksymilianDemitraszek: It may be but I believe stub is more popular one and less misleading

PR #897, opened 2022-09-28
Redesigning `protostar.toml` 9 — Migrator

kasperski95:
This PR adds Configuration File Migrator and corresponding command. The command isn't visible to the user yet.

### DIFF
--- a/protostar/configuration_file/configuration_file.py
+++ b/protostar/configuration_file/configuration_file.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any, Generic, Optional, TypeVar, Union
+from typing import Any, Generic, Optional, Protocol, TypeVar, Union
 
 from protostar.protostar_exception import ProtostarException
 from protostar.self import DeclaredProtostarVersionProviderProtocol, ProtostarVersion
@@ -89,3 +89,8 @@ class ContractNameNotFoundException(ProtostarException):
         super().__init__(
             f"Couldn't find `{contract_name}` in `{expected_declaration_location}`"
         )
+
+
+class ConfigurationFileMigratorProtocol(Protocol):
+    def run(self) -> None:
+        pass
--- a/protostar/configuration_file/configuration_file.py
+++ b/protostar/configuration_file/configuration_file.py
@@ -26,12 +26,13 @@ class ConfigurationFileContentBuilder(ABC):
         data: dict[str, Any],
         profile_name: Optional[str] = None,
     ) -> None:
-        pass
+        ...
 
     @abstractmethod
     def build(self) -> str:
         ...
 
+    @abstractmethod
     def get_file_extension(self) -> str:
         ...
 
--- a/protostar/configuration_file/configuration_file.py
+++ b/protostar/configuration_file/configuration_file.py
@@ -37,7 +37,6 @@ class ConfigurationFileContentConfigurator(Generic[ConfigurationFileModelT]):
     @abstractmethod
     def create_file_content(
         self,
-        content_builder: ConfigurationFileContentBuilder,
         model: ConfigurationFileModelT,
     ) -> str:
         ...
--- a/protostar/configuration_file/configuration_file.py
+++ b/protostar/configuration_file/configuration_file.py
@@ -33,7 +33,7 @@ class ConfigurationFileContentBuilder(ABC):
         ...
 
     @abstractmethod
-    def get_file_extension(self) -> str:
+    def get_content_format(self) -> str:
         ...
 
 
--- a/protostar/configuration_file/configuration_file.py
+++ b/protostar/configuration_file/configuration_file.py
@@ -50,6 +50,10 @@ class ConfigurationFile(
         ...
 
     @abstractmethod
+    def get_filepath(self) -> Path:
+        ...
+
+    @abstractmethod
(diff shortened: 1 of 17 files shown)


protostar/configuration_file/configuration_file.py
```diff
@@ -26,22 +26,28 @@ def set_section(
...
     ) -> None:
-        pass
+        ...
 
     @abstractmethod
     def build(self) -> str:
         ...
 
+    @abstractmethod
+    def get_file_extension(self) -> str:
```
MaksymilianDemitraszek, line 36: This class is a **Content**Builder` i don't thing this method belongs here, file extension is not a content

protostar/configuration_file/configuration_file_v2.py
```diff
@@ -72,6 +71,9 @@ def get_declared_protostar_version(self) -> Optional[ProtostarVersion]:
             return None
         return parse_protostar_version(version_str)
 
+    def get_filepath(self) -> Path:
+        return self._file_path
```
MaksymilianDemitraszek, line 75:
```suggestion
    def get_filepath(self) -> Path:
        return self._file_path
```
Why don't make this property public? Imo it is unnecessary biolderplate

protostar/configuration_file/configuration_file_v2.py
```diff
@@ -109,37 +111,49 @@ def read(
     ) -> ConfigurationFileV2Model:
         raise NotImplementedError("Operation not supported.")
 
-    def create_file_content(
+
+class ConfigurationFileV2ContentConfigurator(
```
MaksymilianDemitraszek, line 115:
```suggestion
class ConfigurationFileV2ContentFactory(
```

MaksymilianDemitraszek (changes requested):

MaksymilianDemitraszek:
Please provide an integration/e2e tests of the feature (those in this case are much more important than unit tests), you can do this in a separate PR though

kasperski95:
> Please provide an integration/e2e tests of the feature (those in this case are much more important than unit tests), you can do this in a separate PR though

This feature is not available for the user. Sociable unit tests used here test this functionality well.

protostar/configuration_file/configuration_file.py
kasperski95: `getContentFormat`?

MaksymilianDemitraszek: better

MaksymilianDemitraszek:
>This feature is not available for the user.
Sure just remember to add them in the PR you will make them available

protostar/self/conftest.py
```diff
@@ -0,0 +1,12 @@
+from .protostar_compatibility_with_project_checker import (
+    ProtostarVersionProviderProtocol,
+    parse_protostar_version,
+)
+
+
+class ProtostarVersionProviderDouble(ProtostarVersionProviderProtocol):
```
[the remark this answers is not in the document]
kasperski95: One may say that this is a `Fake` not a `Stub`. I'd rather stick to the convention `TestDouble` (I'll change the suffix from Double to TestDouble). 

> Why not use a mock? Would be much simpler

`MockerFixture` needs to be used per test case, which makes the test less readable. Also, mocked values are slightly harder to maintain because refactoring breaks such tests.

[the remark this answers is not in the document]
MaksymilianDemitraszek: I think `Double` is just a bad word because has vague meaning, If you prefer stick to `Fake` it is also fine

protostar/configuration_file/configuration_file_v2.py
kasperski95: `ConfigurationFile` callers communicate with concrete classes through interface. `get_filepath` needs to be defined in that interface.

protostar/configuration_file/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+protostar_version = "0.4.0"
+"""
+
+PROTOSTAR_TOML_V2_CONTENT = """
+[project]
+protostar-version = "0.5.0"
+"""
+
+
+class CommandNamesProviderTestDouble(CommandNamesProviderProtocol):
```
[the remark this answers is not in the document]
kasperski95: Explained above

protostar/self/conftest.py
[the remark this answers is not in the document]
kasperski95: > I think Double is just a bad word because has vague meaning

That's why I wanted to rename it `TestDouble`.  Stubs, Fakes, Mocks are often misused.

Renamed `TestDouble`s in [08a6b28](https://github.com/software-mansion/protostar/pull/897/commits/08a6b28366385040c2daaa114fac5dd235fabd23)

protostar/configuration_file/configuration_file.py
kasperski95: Done in [3612d93](https://github.com/software-mansion/protostar/pull/897/commits/3612d9376d39418b26cb3aa1c124ad5cc83fbe97)

protostar/configuration_file/configuration_file_v2.py
kasperski95: Done in [9f01b4c](https://github.com/software-mansion/protostar/pull/897/commits/9f01b4ce49e21154b6d553b17c2803d413761e9f)

PR #901, opened 2022-09-29
Redesigning `protostar.toml` 10 — Rolling integration — `ArgumentValueProvider`

kasperski95:
Replacing `ProtostarTOML` with `ConfigurationFile`.


protostar/configuration_file/configuration_file_v2.py
```diff
@@ -92,20 +91,24 @@ def get_contract_source_paths(self, contract_name: str) -> list[Path]:
             for path_str in contract_section[contract_name]
         ]
 
-    def get_command_argument(
+    def get_argument_value(
```
Arcticae, line 94: how is configuration file related to arguments? this class should not be handling that logic

protostar/composition_root.py
```diff
@@ -62,6 +66,13 @@ def build_di_container(script_root: Path, start_time: float = 0):
         protostar_toml_path.parent if protostar_toml_path is not None else cwd
     )
     protostar_toml_path = protostar_toml_path or project_root_path / "protostar.toml"
+
+    command_names_delayed_provider = CommandNamesDelayedProvider()
```
Arcticae, line 70: why is that class used here? it literally checks if the provider is not `None` - and in which case exactly `protostar_cli` is expected to be `None`?

protostar/start.py
```diff
@@ -1,39 +1,46 @@
...
     configuration_profile_name = (
         ArgumentParserFacade(ConfigurationProfileCLI(), disable_help=True)
         .parse(ignore_unrecognized=True)
         .profile
     )
-    argument_value_from_config_provider = ArgumentValueFromConfigProvider(
-        protostar_toml_reader,
-        configuration_profile_name,
+    argument_value_from_config_provider = (
+        ArgumentValueFromConfigProvider(
```
Arcticae, line 38: This composition of providers is really bizarre and i don't really understand why it exists

Arcticae: Nor will anyone else reading this code probably

Arcticae (changes requested):

protostar/configuration_file/configuration_file_v2.py
kasperski95: `ConfigurationFile` apart from the "normal" interface implements `ArgumentValueProviderProtocol`. 
That interface could be implemented in another class, however it would increase the complexity by introducing a new class and complicating code responsible for class instantiation.

protostar/start.py
kasperski95: The composition is logical, but the names could be better. I'll change that.

CLI needs to get data from the configuration file, if the configuration file exists.

Arcticae:
@MaksymilianDemitraszek can you take a look?

protostar/start.py
kasperski95: Renamed

protostar/configuration_file/configuration_file_v2.py
kasperski95: I'd had to pick what module depends on which module:
`protostar.cli` -> `protostar.configuration_file` or `protostar.configuration_file` -> `protostar.cli`. I picked the second one because the structure of the composition file is affected by CLI, not vice versa.

protostar/cli/argument_value_from_config_extractor.py
```diff
@@ -0,0 +1,65 @@
...
+    ) -> Optional[Any]:
+        ...
+
+    def get_shared_argument_value(
+        self, argument_name: str, profile_name: Optional[str] = None
+    ) -> Optional[Any]:
+        ...
+
+
+class ArgumentValueFromConfigExtractor:
```
MaksymilianDemitraszek, line 16: What is the purpose of this class? The name does not suggest it at all

protostar/cli/argument_value_from_config_extractor.py
```diff
@@ -0,0 +1,65 @@
+from typing import Any, Optional, Protocol
+
+
+class ArgumentValueFromConfigProviderProtocol(Protocol):
```
MaksymilianDemitraszek, line 4: Do we need this protocol? How many classes implement it?

protostar/cli/argument_value_from_config_extractor_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+import pytest
+
+from .argument_value_from_config_extractor import (
+    ArgumentValueFromConfigExtractor,
+    ArgumentValueFromConfigProviderProtocol,
+)
+
+
+class FakeArgumentValueProvider(ArgumentValueFromConfigProviderProtocol):
```
MaksymilianDemitraszek, line 11: I do not like that we have to write so much code to just stub test something. Can we do something about it?

MaksymilianDemitraszek: I believe it is just not worth to write an unit tests for integration classes. @Arcticae WDYT?

protostar/composition_root.py
MaksymilianDemitraszek: I think what @Arcticae meant is why do we hoist it as DI if we just hard code it? I agree with him we shouldn't do that

protostar/configuration_file/configuration_file_v2.py
MaksymilianDemitraszek: I think the point of changes made long time ago was to decompose the class `Project` into smaller classes. Creating one big class for everything is nonsense taking it into account and we can consider the time wasted.

MaksymilianDemitraszek (changes requested):
Code is overengineered, please simplify it.

protostar/start.py
MaksymilianDemitraszek: It is still hard to understand IMO why we wrap one provider into another. Names are too similar and purposes are not clear. I have no clue what is the difference between provider and extractor

protostar/cli/argument_value_from_config_extractor.py
kasperski95: Choosing correct argument (shared arguments, command specific argument, profiled shared argument, profiled command specific command).

`ArgumentValueFromConfigStrategy`?
@Arcticae @MaksymilianDemitraszek Suggestions?

protostar/cli/argument_value_from_config_provider_test.py
```diff
@@ -1,72 +1,81 @@
...
     ArgumentValueFromConfigProvider,
+    ArgumentValueProviderProtocol,
 )
-from protostar.protostar_toml.io.protostar_toml_reader import ProtostarTOMLReader
 
 
-@pytest.fixture(name="configuration_profile_name")
-def configuration_profile_name_fixture() -> Optional[str]:
-    return None
+class ArgumentValueProviderTestDouble(ArgumentValueProviderProtocol):
```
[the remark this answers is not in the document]
kasperski95: *`TestDouble` not *`Double`

https://en.wikipedia.org/wiki/Test_double

protostar/cli/argument_value_from_config_extractor.py
MaksymilianDemitraszek: This class it basically one method, I would merge it with ` ArgumentValueFromConfigProviderProtocol`. As we have only one strategy, there is no point in splitting it. (We would avoid this weird composition this way and code would be simpler)

protostar/cli/argument_value_from_config_extractor.py
kasperski95: This protocol is extended by the `ConfigurationFile` protocol. The purpose of this Protocol is inverting dependency.

from: `protostar.cli` -> `protostar.configuration_file`
to: `protostar.cli` <- `protostar.configuration_file`

protostar/start.py
MaksymilianDemitraszek: We should merge those providers into one, as i suggested in the comment above

protostar/cli/argument_value_from_config_extractor_test.py
Arcticae: it's not that bad if it already exists.

protostar/cli/argument_value_from_config_extractor.py
MaksymilianDemitraszek: Can you please address my question? 

> How many classes implement it?

protostar/configuration_file/configuration_file_v2.py
kasperski95: > I think the point of changes made long time ago was to decompose the class Project into smaller classes. Creating one big class for everything is nonsense taking it into account and we can consider the time wasted.

This class has 64 lines...

> ...decompose the class Project into smaller classes

`ProtostarTOML` sliced the Project class vertically and was a hybrid object. The current approach slices classes horizontally.

MaksymilianDemitraszek: > This class has 64 lines...
Method, wrapped into the class

> sliced the Project class vertically. The current approach slices classes horizontally.
I do not understand what do you mean by that

MaksymilianDemitraszek: @Arcticae what is your opinion?

protostar/composition_root.py
kasperski95: I am not sure if I get it. Should `set_command_names_provider` be part of `ConfigurationFile` interface and that class removed?

protostar/configuration_file/configuration_file_v2.py
kasperski95: - Vertical - by features 
- Horizontal - by technological concepts

Horizontal slicing works better because it makes supporting old and new configuration file at the same time easier.

protostar/composition_root.py
MaksymilianDemitraszek: We should not inject it as a dependency

protostar/cli/argument_value_from_config_extractor.py
kasperski95: This protocol is extended by the `ConfigurationFile` protocol. And the `ConfigurationFile` protocol is implemented by two classes.

protostar/configuration_file/configuration_file_v2.py
Arcticae: In my opinion this breaks SRP, which is going to result in worse code in terms of maintainability. Can't really tell if the "slicing" alteration is making things better on the whole though, it doesn't ocur to me that the approach was modified in an obvious way

protostar/cli/argument_value_from_config_extractor.py
MaksymilianDemitraszek: So why the abstraction extist if only one class extends it?

protostar/configuration_file/configuration_file_v2.py
kasperski95: > In my opinion this breaks SRP...

[SRP is often misunderstood.](https://dev.to/bourzayq_khalid/srp-is-the-most-misunderstood-principle--2b64)

"Horizontal slicing" doesn't violate SRP. Implementing `ArgumentValueProviderProtocol` methods in this class could be understood as a violation of SRP, however I did on the purpose to avoid creating too many classes.

protostar/cli/argument_value_from_config_provider_test.py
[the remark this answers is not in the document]
kasperski95: Done. This class doesn't exist anymore.

protostar/cli/argument_value_from_config_extractor.py
kasperski95: [use ArgumentValueResolver](https://github.com/software-mansion/protostar/pull/901/commits/84905fb42a7b05d3106fda89e5c0013dbdd99c5a)

I removed that protocol. This approach can potentially bite in the future.

protostar/cli/argument_value_from_config_extractor.py
kasperski95: Merged. Closing this thread in favor the one below.

protostar/start.py
kasperski95: Closing this thread in the favor of the one below/above.

protostar/cli/config_file_argument_resolver.py
```diff
@@ -0,0 +1,8 @@
+from typing import Any, Optional, Protocol
+
+
+class ConfigFileArgumentResolverProtocol(Protocol):
```
MaksymilianDemitraszek, line 4: I discussed it with @Arcticae. We shouldn't generalize abstract interfaces when only one class implements it. Point of abstraciton is to generalize something and we should be aware of complexity it introduces. 

Please:
- remove this protocol
- use mock of the exact class in tests instead

protostar/configuration_file/configuration_file_v2.py
MaksymilianDemitraszek: I think it would be easier and simple to create new class than to isolate an abstract protocol.

protostar/cli/argument_value_from_config_extractor.py
MaksymilianDemitraszek: I don't think we have to abstract protocol over that

MaksymilianDemitraszek: Resolving in favor of the new thread

protostar/cli/argument_value_from_config_provider_test.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: Resolving in favor of the new thread

protostar/cli/config_file_argument_resolver.py
kasperski95: Please provide an alternative solution that follows [acyclic dependencies principle](https://en.wikipedia.org/wiki/Acyclic_dependencies_principle).

MaksymilianDemitraszek: Can you point the cycle proposed solution creates?

MaksymilianDemitraszek: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

Arcticae: @kasperski95 is there a cycle when trying to mock the said class?

Arcticae (approved):
I think we should merge this as-is and eventually return to it if it's problematic

protostar/cli/config_file_argument_resolver.py
kasperski95: > Can you point the cycle proposed solution creates?

> @kasperski95 is there a cycle when trying to mock the said class?

I don't know the exact cycle (they are hard to reason about), but I know it exists because:
- it breaks Protostar and VS Code testing extension.
- `protostar.cli` imports from `protostar.configuration_file` and `protostar.configuration_file` imports from `protostar.cli`

> https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming), on the other hand your solution breaks other rule.

The point of that interface is not creating abstraction but creating boundaries, hence the rule of three doesn't really apply here.

> In essence, a system’s architecture is what defines the shape of the system. More specifically, a system’s architecture defines how the system is divided into components, how those components are arranged, what of kinds boundaries exist between different components and how the components communicate across those boundaries. Basically, it’s all about the way we are using boundaries to separate parts of the system that shouldn’t know too much about each other.
https://convincedcoder.com/2019/04/27/Software-architecture-boundaries/
 
> use mock of the exact class in tests instead

I can use them instead of stubs and use the rule of three here, (maybe in the future PRs). Mocks create more fragile tests and pollute test cases with implementation details.

> I think it would be easier and simpler to create a new class than to isolate an abstract protocol.

A developer would be responsible for keeping that new class and `ConfigurationFile` concrete class in sync.

protostar/configuration_file/configuration_file_v2.py
kasperski95: Responded to that in another thread.

PR #906, opened 2022-09-30
Fixed InstructionID issue, partial refactor of transaction profiler

MaksymilianDemitraszek:
This PR includes:
- Made Instruction ID alghoritm deterministic, removed randomness
- Added unit test for some functions
- Frozen all dataclasses for safety reasons

### DIFF
--- a/protostar/profiler/transaction_profiler.py
+++ b/protostar/profiler/transaction_profiler.py
@@ -64,10 +64,13 @@ def translate_callstacks(
     global_instructions: Dict[Tuple[GlobalFunctionID, int], Instruction],
     callstacks: List[List[Instruction]],
 ) -> List[List[Instruction]]:
-    translated : List[List[Instruction]]= []
+    translated: List[List[Instruction]] = []
     for call in callstacks:
-        translated.append(translate_callstack(current_contract, global_instructions, call))
+        translated.append(
+            translate_callstack(current_contract, global_instructions, call)
+        )
     return translated
+
 
 def build_global_functions(
     samples: List["ContractProfile"],
@@ -77,7 +80,7 @@ def build_global_functions(
         function_id_prefix = sample.callstack[-1] + "."
         for func in sample.profile.functions:
             global_name = function_id_prefix + func.id
-            global_functions[global_name] = replace(func, id = global_name)
+            global_functions[global_name] = replace(func, id=global_name)
     return global_functions
 
 
@@ -88,16 +91,13 @@ def get_instruction_id_offsets(
     for sample in samples:
         current_contract = sample.callstack[-1]
         previous = contract_id_offsets.get(current_contract, 0)
-        new = max([instr.id for instr in sample.profile.instructions]) + 1
-        contract_id_offsets[current_contract] = max(
-            new, previous
-        )
+        new = max(instr.id for instr in sample.profile.instructions) + 1
+        contract_id_offsets[current_contract] = max(new, previous)
     accumulator = 0
     for contract_name, size in contract_id_offsets.items():
         contract_id_offsets[contract_name] = accumulator
         accumulator += size
     return contract_id_offsets
-
 
 
 def build_global_instructions(
@@ -112,7 +112,9 @@ def build_global_instructions(
             global_instruction_id = instr.id + contract_id_offsets[current_contract]
             global_function_id = function_id_prefix + instr.function.id
             in_instructions[(global_function_id, instr.id)] = replace(
-                instr, id=global_instruction_id, function=global_functions[global_function_id]
+                instr,
+                id=global_instruction_id,
+                function=global_functions[global_function_id],
             )
     return in_instructions
 
@@ -139,7 +141,11 @@ def merge_profiles(samples: List["ContractProfile"]) -> TransactionProfile:
 
     # We build a tree from callstacks
     max_clst_len = max(len(s.callstack) for s in samples)
-    callstacks_from_upper_layer = translate_callstacks(samples[-1].callstack[-1], global_instructions, samples[-1].profile.contract_call_callstacks, )
+    callstacks_from_upper_layer = translate_callstacks(
+        samples[-1].callstack[-1],
+        global_instructions,
+        samples[-1].profile.contract_call_callstacks,
+    )
     for i in range(2, max_clst_len + 1):
         samples_in_layer = [s for s in samples if len(s.callstack) == i]
         new_callstacks_from_upper_layer: List[List[List[Instruction]]] = []
@@ -151,10 +157,15 @@ def merge_profiles(samples: List["ContractProfile"]) -> TransactionProfile:
                         value=smp.value,
                         callstack=translate_callstack(
                             current_contract, global_instructions, smp.callstack
-                        ) + upper,
+                        )
+                        + upper,
                     )
                 )
-            translated_callstacks = translate_callstacks(current_contract, global_instructions, runtime_sample.profile.contract_call_callstacks)
+            translated_callstacks = translate_callstacks(
+                current_contract,
+                global_instructions,
+                runtime_sample.profile.contract_call_callstacks,
+            )
             new_callstacks_from_upper_layer.append(
                 [c + upper for c in translated_callstacks]
             )
@@ -174,7 +185,11 @@ def merge_profiles(samples: List["ContractProfile"]) -> TransactionProfile:
         )
 
     # We build a tree from callstacks
-    callstacks_from_upper_layer = translate_callstacks(samples[-1].callstack[-1], global_instructions, samples[-1].profile.contract_call_callstacks, )
+    callstacks_from_upper_layer = translate_callstacks(
+        samples[-1].callstack[-1],
+        global_instructions,
+        samples[-1].profile.contract_call_callstacks,
+    )
     for i in range(2, max_clst_len + 1):
         samples_in_layer = [s for s in samples if len(s.callstack) == i]
         new_callstacks_from_upper_layer: List[List[List[Instruction]]] = []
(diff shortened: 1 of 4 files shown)


protostar/profiler/transaction_profiler.py
```diff
@@ -0,0 +1,226 @@
...
+        callstacks_from_upper_layer = list(
+            itertools.chain(*new_callstacks_from_upper_layer)
+        )
+
+    return TransactionProfile(
+        functions=list(global_functions.values()),
+        instructions=list(global_instructions.values()),
+        step_samples=step_samples,
+        memhole_samples=memhole_samples,
+    )
```
kasperski95, line 226: This function is way too long.

protostar/profiler/transaction_profiler_test.py
```diff
@@ -0,0 +1,223 @@
...
+                        [],
+                        [
+                            Instruction(10, 0, replace(funct), 0),
+                            Instruction(100, 0, replace(funct), 0),
+                        ],
+                        [],
+                        [],
+                        [],
+                    ),
+                ),
```
kasperski95, line 103: This going to be hard to maintain. Can you hide this somehow behind a readible testing API?

kasperski95 (commented):
<Quick feedback>

protostar/profiler/transaction_profiler.py
MaksymilianDemitraszek: Agreed, I slowly decomposing it (removed few block from the top in this PR) I will proceed further in the following PR

protostar/profiler/transaction_profiler_test.py
MaksymilianDemitraszek: I will think about it, but it is gonna be hard. It looked much better before linting 😓

kasperski95: Removing the trailing comma might help.

protostar/profiler/transaction_profiler.py
```diff
@@ -0,0 +1,226 @@
...
+    current_contract: "ContractFilename",
+    global_instructions: Dict[Tuple[GlobalFunctionID, int], Instruction],
+    callstacks: List[List[Instruction]],
+) -> List[List[Instruction]]:
+    translated: List[List[Instruction]] = []
+    for call in callstacks:
+        translated.append(
+            translate_callstack(current_contract, global_instructions, call)
+        )
+    return translated
```
kasperski95, line 72: Style: Consider top-down approach to vertical formatting, e.g.`translate_callstacks` should be above `translate_callstack`.

> Source code is like a newspaper article. The farther down we go, the more detailed the article gets.
https://levelup.gitconnected.com/javascript-clean-code-vertical-formatting-d15097f5f30e

protostar/profiler/transaction_profiler_test.py
```diff
@@ -0,0 +1,223 @@
...
+                ("b.C", 150): Instruction(1151, 0, glob_func["b.C"], 0),
+                ("b.C", 100): Instruction(1101, 0, glob_func["b.C"], 0),
+                ("a.B", 100): Instruction(100, 0, glob_func["a.B"], 0),
+            },
+        ),
+        ([], {}),
+    ],
+)
+def test_build_global_instructions(samples, result):
+    assert build_global_instructions(glob_func, samples) == result
```
kasperski95, line 179: Missing types. Tests can be treated as a documentation. I start reviewing from here, and I don't know what a Sample is. Is this term defined? If so, why where is defined?

Maybe samples defined here should be moved to the conftest file and given a name.

kasperski95 (approved):
LGTM.

I struggle with the understanding of this code because it's written in bottom-up way.

protostar/profiler/transaction_profiler.py
MaksymilianDemitraszek: I am used to opposite approach bcs of c, I always write my code bottom up and I prefer it this way :p Why should decide on one format and use it globally in the project

protostar/profiler/transaction_profiler.py
```diff
@@ -0,0 +1,226 @@
...
+        new = max(instr.id for instr in sample.profile.instructions) + 1
+        contract_id_offsets[current_contract] = max(new, previous)
+    accumulator = 0
+    for contract_name, size in contract_id_offsets.items():
+        contract_id_offsets[contract_name] = accumulator
+        accumulator += size
+    return contract_id_offsets
+
+
+def build_global_instructions(
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: I do not agree, this function builds the structure based on some resources, it much better represents what it does

protostar/profiler/transaction_profiler.py
```diff
@@ -0,0 +1,226 @@
+# pylint: disable=invalid-name
+from dataclasses import dataclass, replace
+import itertools
+from typing import TYPE_CHECKING, List, Dict, Tuple
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: I think we use List, Dict in many places, I would prefer to be consistent.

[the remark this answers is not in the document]
kasperski95: @mkaput would request changes for that reason only.
https://github.com/software-mansion/protostar/pull/728#pullrequestreview-1107071664

I think we should migrate to new types in one way or another.

protostar/profiler/transaction_profiler.py
kasperski95: We need to define the style. It can be done here https://github.com/software-mansion/protostar/issues/928.

protostar/profiler/transaction_profiler.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: Oh, it's python  3.9 feature, it makes sense then why we haven't used it before. I will change it

protostar/profiler/transaction_profiler.py
MaksymilianDemitraszek: I will change the ordering

protostar/profiler/transaction_profiler_test.py
MaksymilianDemitraszek: >Maybe samples defined here should be moved to the conftest file and given a name.

Those are used only in this file, I don't think it is a good idea

protostar/profiler/transaction_profiler.py
```diff
@@ -0,0 +1,211 @@
...
+        ContractProfile,
+        ContractFilename,
+    )
+
+Address = int
+GlobalFunctionID = str
+GlobalInstructionID = str
+
+GlobalFunction = Function
+GlobalInstruction = Instruction
```
kasperski95, line 24: Can functions be local?

protostar/starknet/cheatable_execute_entry_point.py
```diff
@@ -54,9 +54,12 @@
 logger = logging.getLogger(__name__)
 
 
+ContractFilename = str
+
+
 @dataclass(frozen=True)
 class ContractProfile:
-    callstack: List[str]
+    callstack: List[ContractFilename]
```
kasperski95, line 62: I find it confusing because `Sample::callstack` has a different type (`List[Instruction]`).

protostar/starknet/cheatable_execute_entry_point.py
```diff
@@ -36,7 +36,7 @@
 from starkware.starknet.business_logic.state.state import StateSyncifier
 from protostar.profiler.pprof import serialize, to_protobuf
 from protostar.profiler.contract_profiler import RuntimeProfile, build_profile
-from protostar.profiler.profile import (
+from protostar.profiler.transaction_profiler import (
```
kasperski95, line 39: Imports could be relative.

protostar/profiler/transaction_profiler_test.py
```diff
@@ -0,0 +1,223 @@
...
+)
+
+
+@pytest.mark.parametrize(
+    "samples,result",
+    [
+        (
+            [
+                ContractProfile(["a", "b"], ContractEntryPoint(0, 0), replace(profile)),
+                ContractProfile(["c", "b"], ContractEntryPoint(0, 0), replace(profile)),
```
kasperski95, line 27: Keyword arguments could improve readability.

protostar/profiler/transaction_profiler_test.py
```diff
@@ -0,0 +1,223 @@
...
+    [Function("f1", "d", 1), Function("f2", "d", 2)], [], [], [], []
+)
+
+
+@pytest.mark.parametrize(
+    "samples,result",
+    [
+        (
+            [
+                ContractProfile(["a", "b"], ContractEntryPoint(0, 0), replace(profile)),
```
kasperski95, line 26: Why is `replace` necessary here?

kasperski95 (approved):

protostar/profiler/transaction_profiler.py
MaksymilianDemitraszek: Yep, local in terms contract, global in terms of transaction

protostar/starknet/cheatable_execute_entry_point.py
MaksymilianDemitraszek: Because there are two different callstacks, contract callstack(cairo functions) and transaction callstack(called contracts). I will rename it to transaction_callstack

protostar/profiler/transaction_profiler.py
kasperski95: So `GlobalFunction` == `Entrypoint`?

protostar/profiler/transaction_profiler_test.py
MaksymilianDemitraszek: All of them use the same structure, I want to avoid them having the same reference

protostar/profiler/transaction_profiler_test.py
MaksymilianDemitraszek: I will add them in some places, but not everywhere, would vastly increase the size of the code

MaksymilianDemitraszek: https://github.com/software-mansion/protostar/issues/978

protostar/starknet/cheatable_execute_entry_point.py
MaksymilianDemitraszek: https://github.com/software-mansion/protostar/issues/978

PR #907, opened 2022-09-30
Skip checking for upgrades on `upgrade` command

Arcticae:


kasperski95 (approved):
Please add a test case.

PR #915, opened 2022-10-03
Removing GitPython from tests

Radinyn:


protostar/git/git_repository.py
```diff
@@ -0,0 +1,262 @@
...
+    InvalidGitRepositoryException,
+    wrap_git_exception,
+)
+
+
+@dataclass
+class Submodule:
+    name: str
+    url: str
+    version: Optional[str] = None
```
kasperski95, line 17: Nit: `Submodule::version` doesn't make sense. Consider `PackageInfo.from_submodule` to allow these two dataclasses to vary independently.

kasperski95 (dismissed):

protostar/git/git_repository.py
Radinyn: I don't think it's particularly important at the moment. I might do it in a separate PR, perhaps related to package version enforcing.

kasperski95:
10:59:09 [ERROR] Installation failed
10:59:09 [ERROR] Error while executing Git command:
Command '['git', 'commit', '-m', 'add cairo_contracts']' returned non-zero exit status 1.
10:59:09 [INFO] Execution time: 0.96 s

kasperski95 (approved):
This PR fixes the issue with installing the specific version. However, I've managed to crash protostar by running:
```
protostar install ...
protostar remove ...
protostar install ...
```
That problem can be solved in the different PR.

Issue #914, opened 2022-10-03
No state deepcopy


kasperski95:
https://github.com/Shard-Labs/starknet-devnet/pull/293

Solpatium:
I've added this code to `__init__.py`:
```python
from copy import copy
from starkware.starknet.services.api.contract_class import ContractClass
# Deep copy of a ContractClass takes a lot of time, but it should never be mutated.
def simpler_copy(self, memo):  # pylint: disable=unused-argument
    """
    A dummy implementation of ContractClass.__deepcopy__
    """
    return copy(self)


setattr(ContractClass, "__deepcopy__", simpler_copy)
```

and test execution time in https://github.com/software-mansion-labs/GoL2 dropped from 108.36s to 76.91s (29%). In other project with 178 tests (some are quite big) it dropped from 104.53s to 86.63s (17%). The gains should be much higher in case of small tests.

PR #924, opened 2022-10-04
Monkeypatch ContractClass.__deepcopy__ to improve performance

Solpatium:
Related to https://github.com/software-mansion/protostar/issues/914. 
Ariel confirmed that instances of ContractClass are not mutated.


kasperski95 (approved):
LGTM, if Ariel confirmed that instances of ContractClass are not mutated.

PR #929, opened 2022-10-05
Fix fee arg type

kasperski95:


protostar/cli/argument_parser_facade.py
```diff
@@ -196,3 +182,26 @@ def _add_argument(
             **kwargs,
         )
         return argument_parser
+
+    @staticmethod
+    def _map_type_to_arg_type(argument_type: InputAllowedType) -> Callable[[str], Any]:
```
Arcticae, line 187: That's always bugged me - can't be this done with dict access? You wouldn't have to do false assertions and have so many ifs. Also - more pythonic that way

Arcticae (approved):

PR #936, opened 2022-10-05
Refactor `SignerCommandUtils`

kasperski95:


Arcticae:
Ah i already approved. I guess you're lucky today 😛

kasperski95:
> I'm not sure we need this. Can you explain why? Then i'll approve

- I want to intercept transactions in the `GatewayClient` to log them to the user instead of doing that manually because current approach is error prone. In order to do so, I need to create a subclass of `GatewayClient` and fix the code everywhere where `GatewayClient` is used instead in one place. I noticed `GatewayFacade` is leaking, which I am trying to fix with bottom-up approach.
- The PR extracts a model `AccountConfig` which can be used as a building block. For example, `DevnetAccount` can extend `AccountConfig` and be passed directly to the gateway facade. Changes to `AccountConfig` won't affect `GatewayFacade` interface.
- `AccountConfig` guarantees address and signer are not None.
- `get_signer` was a method with 3 responsibilities
- Marek complained about `Util` suffix.

protostar/starknet_gateway/signer.py
```diff
@@ -0,0 +1,36 @@
+from dataclasses import dataclass
+
+from starknet_py.net.models import StarknetChainId
+from starknet_py.net.signer import BaseSigner
+from starknet_py.net.signer.stark_curve_signer import KeyPair, StarkCurveSigner
+
+from protostar.protostar_exception import ProtostarException
+
+ProtostarBaseSigner = BaseSigner
+ProtostarDefaultSigner = StarkCurveSigner
```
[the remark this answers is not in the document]
kasperski95: I'll remove them.

protostar/cli/signer_factory.py
```diff
@@ -0,0 +1,100 @@
...
+    ),
+    Command.Argument(
+        name="signer-class",
+        description="Custom signer class module path.",
+        type="str",
+    ),
+]
+
+
+def create_account_config_from_args(
```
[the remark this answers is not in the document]
kasperski95: Hmm, I should have picked a better title for this PR. It's more about extracting `AccountConfig` model. I left the `args` here because this function uses `create_custom_signer` and `get_private_key` which is IO.

protostar/starknet_gateway/signer.py
```diff
@@ -0,0 +1,33 @@
+from dataclasses import dataclass
```
mkaput, line 1: these classes are only used in `signer_factory.py`

could these be moved there?

protostar/cli/signer_factory.py
```diff
@@ -0,0 +1,97 @@
...
+
+def create_custom_signer(signer_class_path: str) -> BaseSigner:
+    *module_names, class_name = signer_class_path.split(".")
+    module = ".".join(module_names)
+    signer_module = importlib.import_module(module)
+    # pylint: disable=invalid-name
+    SignerClass = getattr(signer_module, class_name)
+    if not issubclass(SignerClass, BaseSigner):
+        raise InvalidSignerClassException()
+    return SignerClass()
```
mkaput, line 63:
```suggestion
    signer_factory = getattr(signer_module, class_name)
    
    try:
        signer = signer_factory()
    except TypeError as ex:
        raise InvalidSignerClassException() from ex
        
    if not isinstance(signer, BaseSigner):
        raise InvalidSignerClassException()
        
    return signer
```

1. This way users can pass custom factory functions instead of classes directly.
1. Why `try ... except`?: https://stackoverflow.com/a/5028449/3657395

protostar/cli/signer_factory.py
```diff
@@ -0,0 +1,97 @@
...
+        signer=create_stark_curve_signer(
+            account_address=args.account_address,
+            private_key=get_private_key(args.private_key_path),
+            chain_id=chain_id,
+        ),
+    )
+
+
+def create_custom_signer(signer_class_path: str) -> BaseSigner:
+    *module_names, class_name = signer_class_path.split(".")
```
mkaput, line 56: no need for immediate action: I recall somewhere that a convention for specifying Python module items looks like this: `module.path.a.b.c:item_name`. Gunicorn does this for example: https://gunicorn.org/

mkaput (changes requested):

PR #943, opened 2022-10-07
Replacing `utils` 1 - move everything except IO

Radinyn:


kasperski95 (commented):
I'd move `protostar_directory` to `protostar.self`. I plan to extract `VersionManager` to a separate file. `VersionManager` will be used directly by the `CompatibilityChecker` in the future.

PR #945, opened 2022-10-07
Replace `utils`

Radinyn:


Arcticae (approved):
Nit: I kind of dislike name `self` since it's somewhat replacing `utils`. it's not really saying anything about the files purpose - if that's the only file in the directory we can just leave it in the root instead of hiding it in directory. Do as you wish though - the refactor is fine as-is

kasperski95 (approved):
LGTM

> Nit: I kind of dislike name self since it's somewhat replacing utils. it's not really saying anything about the files purpose

@Arcticae I created the module `self`. It can be renamed later if we find a better name.

> - if that's the only file in the directory we can just leave it in the root instead of hiding it in directory. Do as you wish though - the refactor is fine as-is

There's one more file there but `VersionManager` needs to be extracted to a new file anyway, so it kinda justifies the reason of existence. Keeping files in the root makes `protostar` folder look like utils, and high-level concepts are mixed with low-level concepts.

PR #947, opened 2022-10-07
Parameterized tests

### LINKED ISSUE #623: Parameterized tests
Use Hypothesis's [`@example` decorator](https://hypothesis.readthedocs.io/en/latest/reproducing.html?highlight=example#hypothesis.example)

API and implementation design:
1. Introduce a new cheatcode `example(*args: Any, **kwargs: Any)`. This cheatcode will add *a single* example to explore. Semantics very simlar to `given`. Also only available in setup cases.
2. Examples will be stored in `TestConfig` and consumed in `FuzzTestExecutionEnvironment`. Like we already do with `given` cheatcode.
3. A new `TestMode` will be introduced: `TestMode.PARAMETERIZED`. It will also resolve to `FuzzTestExecutionEnvironment` and valid transitions are:
```
STANDARD -> PARAMETERIZED
STANDARD -> FUZZ
PARAMETERIZED -> FUZZ
FUZZ -> PARAMETERIZED should be noop and yield FUZZ anyway
```
4. Don't forget to write docs. Inclusion in tutorial might also be nice (how to ensure that the edge case we found is always tested).

karol-bisztyga:
Introducing the possibility to parameterize tests with desired, specific examples. Inspired by hypothesis' [`example` decorator](https://hypothesis.readthedocs.io/en/latest/reproducing.html#hypothesis.example).

fix #623


protostar/testing/cheatcodes/example_cheatcode.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def build(self) -> ExampleCallable:
+        return self.example
+
+    def example(self, *args: Any, **kwargs: Any) -> None:
+        self.test_config.convert_mode_to(TestMode.FUZZ)
+
+        assert bool(args) != bool(
+            kwargs
+        ), "you have to provide either args or kwargs to the example cheatcode"
```
kasperski95, line 36: [Functionality] `AssertionError` tells the user to report the issue. Raising `CheatcodeException` would be more user-friendly.

protostar/testing/cheatcodes/example_cheatcode.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @property
+    def name(self) -> str:
+        return "example"
+
+    def build(self) -> ExampleCallable:
+        return self.example
+
+    def example(self, *args: Any, **kwargs: Any) -> None:
+        self.test_config.convert_mode_to(TestMode.FUZZ)
```
kasperski95, line 32: [Functionality] Shouldn't test mode be changed to `TestMode.PARAMETERIZED`?

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -37,6 +37,22 @@
 )
 
 
+def multiply_decorator(other_dec, args):
```
kasperski95, line 40:
```suggestion
def multiply_decorator(other_dec: Callable, args: list):
```

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -37,6 +37,22 @@
...
+    def inner(func):
+        current_fn = func
+        if not args:
+            return current_fn
+        args.reverse()
```
kasperski95, line 45: [Tests] Why `args` reversed? Can you add a unit test for this decorator? Tests can be used as a form of documentation.

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -167,7 +199,7 @@ async def test(**inputs: Any):
         except InvalidArgument as ex:
             # This exception is sometimes raised by Hypothesis during runtime when user messes up
             # strategy arguments. For example, invalid range for `integers` strategy is caught here.
-            raise CheatcodeException("given", str(ex)) from ex
+            raise CheatcodeException("unknown", str(ex)) from ex
```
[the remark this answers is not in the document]
kasperski95: [Functionality] Maybe `given or example` would be better to help the user with debugging.

protostar/testing/test_config.py
```diff
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import Dict
+from typing import Dict, List
```
kasperski95, line 3: [Consistency] We started using Python 3.9 types: `dict`, `list`.

tests/integration/fuzzing/parameterized_test.cairo
```diff
@@ -0,0 +1,25 @@
+%lang starknet
+
+from starkware.cairo.common.math import assert_le
+
+@external
+func setup_examples() {
+    %{
+        max_examples(5)
+        # todo test no example
```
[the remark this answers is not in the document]
kasperski95: Yep, such a test case is not needed.

tests/integration/fuzzing/parameterized_test.cairo
```diff
@@ -0,0 +1,25 @@
...
+        example(1, 2)
+        example(2, 3)
+        example(a=3, b=4)
+        example(b=6, a=5)
+        given(a = strategy.integers(15, 20), b = strategy.integers(10, 14))
```
kasperski95, line 14: [Functionality] I am confused why is `given` cheatcode necessary here. It looks like a hypothesis requirement. I think Protostar should throw an exception if `example` is used without `given` telling the user that given is necessary.

protostar/testing/environments/fuzz_test_execution_environment.py
kasperski95, line 40: [Style] This function could be moved to a different file.

protostar/testing/test_config.py
```diff
@@ -13,6 +13,11 @@
 from .testing_seed import Seed, random_seed
 
 
+class ExampleData:
+    def __init__(self):
+        self.mapping = {}
+
+
```
kasperski95, line 20: [Design] Why is this class necessary?

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -127,7 +144,22 @@ def build_and_run_test(
...
+            ]
+            assert len(functions) == 1
+            function = functions[0]
+            function_inputs_names = [input["name"] for input in function["inputs"]]
+            examples = []
+            for example_item in self.state.config.fuzz_examples:
+                if isinstance(example_item, dict):
+                    examples.append(example_item)
+                    continue
+                examples.append(dict(zip(function_inputs_names, list(example_item))))
```
kasperski95, line 160: [Style] This could be a method.

tests/integration/fuzzing/parameterized_test.cairo
```diff
@@ -0,0 +1,25 @@
+%lang starknet
+
+from starkware.cairo.common.math import assert_le
+
+@external
+func setup_examples() {
+    %{
+        max_examples(5)
```
kasperski95, line 8: [Tests] What happens when `max_examples` is lower than usage of `example`?

tests/integration/fuzzing/parameterized_test.cairo
```diff
@@ -0,0 +1,25 @@
...
+        example(b=6, a=5)
+        given(a = strategy.integers(15, 20), b = strategy.integers(10, 14))
+    %}
+    return ();
+}
+
+@external
+func test_examples{syscall_ptr: felt*, range_check_ptr}(a: felt, b: felt) {
+    assert_le(0, a);
+    assert_le(0, b);
```
kasperski95, line 22: [Tests] Test checking if the specified example is actually provided to the test case is missing.

tests/integration/fuzzing/parameterized_test.cairo
mkaput: `max_examples` does not apply to `example`-caused runs. it only affects `given`.

This is all controlled by a concept of _phases_ in Hypothesis. `example` and `given` control separate phases. Another one is shrinking.

https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.Phase

tests/integration/fuzzing/parameterized_test.cairo
mkaput: I think you don't have to use `given` here, from Protostar's perspective.

Under the hood we might want to detect usage of `example` without `given (this is trivial due to how TestConfig works), and apply a `given` decorator with only `explicit` phase enabled.

protostar/testing/test_config.py
mkaput: you can use `PythonData` type alias here instead

protostar/testing/environments/fuzz_test_execution_environment.py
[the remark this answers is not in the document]
mkaput: Hmm I think we should rather shift towards raising more general `FuzzingError` here.

protostar/testing/cheatcodes/example_cheatcode.py
mkaput: `given` only accepts `kwargs`. `example` should do that too.

protostar/testing/cheatcodes/example_cheatcode.py
```diff
@@ -0,0 +1,40 @@
...
+
+        assert bool(args) != bool(
+            kwargs
+        ), "you have to provide either args or kwargs to the example cheatcode"
+        assign_args = args
+        if kwargs:
+            assign_args = kwargs
+        self.test_config.fuzz_examples.append(assign_args)
```
mkaput, line 40:
```suggestion
        self.test_config.fuzz_examples |= kwargs
```

protostar/testing/test_config.py
```diff
@@ -89,6 +94,8 @@ class TestConfig:
         default_factory=dict
     )
 
+    fuzz_examples: List[ExampleData] = field(default_factory=list)
```
mkaput, line 97:
```suggestion
    fuzz_examples: PythonData = field(default_factory=dict)
```

protostar/testing/environments/fuzz_test_execution_environment.py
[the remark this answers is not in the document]
karol-bisztyga: We can do that, although it doesn't look entirely right.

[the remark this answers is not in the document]
mkaput: `FuzzingError`? It actually can be right, depending on what perspective are you looking from.

Until now, we were pretending to say that strategies are statically fully determined during `given` cheatcode evaluation, which is **not** true. By raising `FuzzingError` we simply surface the fact that the strategies do some lazy initialization.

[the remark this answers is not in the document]
mkaput: Also, if we think about this more: strategies are silly only related to the `given` cheatcode 😅 

`example` operates on concrete values only.

protostar/testing/cheatcodes/example_cheatcode.py
karol-bisztyga: It should, I didn't cover that as, at this point, it was just a POC and I was going to do some stuff later

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: The `args` should be reversed to keep the correct order of invoking. When you write your test setup like this:
```
example(1)
example(2)
example(3)
given(some strategy...)
```
you expect this test to be run with the following data in the following order:
```
test(1)
test(2)
test(3)
for x in data_from_strategy: test()
```

Now, the `multiply_decorator` wraps the current function with an additional `example` (basically it might be some other decorator, but in this case, we have the `example`) function with a specified set of arguments, so it would work like that (for the test I added in the beginning):
```
example(example(example(given(some strategy...), 1), 2), 3)
```
So you would get something like this
```
test(3)
test(2)
test(1)
for x in data_from_strategy: test()
```
That's why I put the `reverse` there.

> Can you add a unit test for this decorator?

Sure

protostar/testing/test_config.py
karol-bisztyga: It's not, will be removed

tests/integration/fuzzing/parameterized_test.cairo
karol-bisztyga: `max_examples` only refers to `given`, `example` works separately.

kasperski95: It's worth to point out that later in the docs.

tests/integration/fuzzing/parameterized_test.cairo
karol-bisztyga: This is correct, we do not have to use `given` cheatcode. Basically, hypothesis requires `given` when we use `example` but in protostar, `given` seems to be used automatically when not specified explicitly, so we're not going to get this error from hypothesis.

protostar/testing/cheatcodes/example_cheatcode.py
karol-bisztyga: Are you sure about this? Doesn't seem to work, plus if `args` are provided, this code misses them. Doesn't this operator `|=` work only for the pair of objects of the same type? Here, we have `args, kwargs` that are `tuple` and `dict` respectively and `fuzz_examples` is essentially a list.

protostar/testing/test_config.py
karol-bisztyga: I think we need a list here.

tests/integration/fuzzing/parameterized_test.cairo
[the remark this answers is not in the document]
karol-bisztyga: [remove redundant todo from cairo test file](https://github.com/software-mansion/protostar/pull/947/commits/2fd0a437cad4063c34d64430e8f18c646533478c)

protostar/testing/environments/fuzz_test_execution_environment.py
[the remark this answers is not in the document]
karol-bisztyga: [change error message to more descriptive](https://github.com/software-mansion/protostar/pull/947/commits/a0e49741e12012e2a28febc07d675fadfad20b92)

protostar/testing/cheatcodes/example_cheatcode.py
karol-bisztyga: [use cheatcode exception instead of assert](https://github.com/software-mansion/protostar/pull/947/commits/49a8c53ee21304110aa266c771336b06b33d55a4)

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: [move multiply decorator to a separate file](https://github.com/software-mansion/protostar/pull/947/commits/7443eba804bd76a54ffd481ae41bb185d5872ffa)

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: [add test for multiply decorator](https://github.com/software-mansion/protostar/pull/947/commits/361e96ec4a40e7e9b41092acfd7c4813ef13bb0b)

protostar/testing/test_config.py
karol-bisztyga: [use built-in types for dict and tuple](https://github.com/software-mansion/protostar/pull/947/commits/8c50898fc9488104046d067b20440d3335dfd6bc)

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: [move multiply decorator to a separate file](https://github.com/software-mansion/protostar/pull/947/commits/7443eba804bd76a54ffd481ae41bb185d5872ffa)

protostar/testing/test_config.py
karol-bisztyga: [Remove ExampleData class](https://github.com/software-mansion/protostar/pull/947/commits/7da6be7293f367eef033fe97e6cff9827f102998)

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: [extract code to produce_examples](https://github.com/software-mansion/protostar/pull/947/commits/cc5625620677e0ba6e22ab0be2c1e349adb1f700)

tests/integration/fuzzing/parameterized_test.cairo
karol-bisztyga: [check number of test runs](https://github.com/software-mansion/protostar/pull/947/commits/7b224466a0986a856f952a105af472d4b2ca7737)

protostar/testing/cheatcodes/example_cheatcode.py
mkaput: @karol-bisztyga please take a look at how `given` is implemented and replicate its behaviour here. `given` does not take any `args` for simplicity

protostar/testing/cheatcodes/example_cheatcode.py
mkaput: this is strictly connected to my other suggestions to remove `args` entirely here and to use `PythonData` :)

applying everything you get `dict[str, Any]` on both sides of the `|=` operator

protostar/testing/test_config.py
mkaput: ah yes, you're right. use `List[PythonData]`

tests/integration/fuzzing/parameterized_test.cairo
mkaput: @karol-bisztyga you can if you manipulate the `phases` setting https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.phases

kasperski95:
@mkaput feel free to mark conversation started by me as resolved

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -128,6 +147,7 @@ def build_and_run_test(
     ):
         try:
 
+            @multiply_decorator(example, self.produce_examples(function_name))
```
mkaput, line 150: I find this a little bit overengineered, because of the trial to generalize the `multiply_decorator` function.

What you are doing here seems to be boiling down to a very trivial reducing for loop:

```python
f = test
for example in reversed(self.state.config.fuzz_examples):
    f = hypothesis.example(**example)(f)

return f
```

this is a very rare use case and I don't see a need to generalise it, and by inlining this you save a lot of code that sets up inputs data structures etc.

(simplified to only handle kwargs, but this is not a point of this suggestion)

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -49,6 +50,7 @@ def __init__(self, state: TestExecutionState):
 
     async def execute(self, function_name: str) -> FuzzTestExecutionResult:
         abi = self.state.contract.abi
+        # here are params
```
mkaput, line 53:
```suggestion
```

protostar/testing/test_config.py
mkaput: use `PythonData` instead of `dict`

protostar/testing/environments/fuzz_test_execution_environment.py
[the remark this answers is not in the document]
mkaput: I find the use of `or` word confusing here and not professionally looking. Take into account that this string is presented as **cheatcode name** to the user! It'd be better to raise FuzzingError here, really IMHO

mkaput (changes requested):

protostar/testing/test_config.py
karol-bisztyga: > you can use `PythonData` type alias here instead, see my other review comment

Thanks for the suggestion but this was really redundant.

protostar/testing/environments/fuzz_test_execution_environment.py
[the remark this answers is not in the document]
karol-bisztyga: Sounds good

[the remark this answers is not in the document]
karol-bisztyga: [raise fuzzing error instead of a cheatcode exception](https://github.com/software-mansion/protostar/pull/947/commits/aeabdb33a47b3da8c9377257b767f1347b5cba88)

protostar/testing/test_config.py
karol-bisztyga: [allow only keyword based arguments](https://github.com/software-mansion/protostar/pull/947/commits/ca556714d1697625088c577d841cc52c4300b044)

tests/integration/fuzzing/parameterized_test.cairo
karol-bisztyga: [update docs](https://github.com/software-mansion/protostar/pull/947/commits/dd62519a5048383692bf43306551af778b51bb28)

protostar/testing/test_config_test.py
```diff
@@ -8,6 +8,9 @@
     [
         (TestMode.STANDARD, TestMode.FUZZ, True),
         (TestMode.FUZZ, TestMode.STANDARD, False),
+        (TestMode.STANDARD, TestMode.PARAMETERIZED, False),
```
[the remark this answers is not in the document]
karol-bisztyga: [allow conversion from standard to parameterized](https://github.com/software-mansion/protostar/pull/947/commits/4d05801e5171f1405f63bebbfacf4d00279ce315)

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: [remove redundant comment](https://github.com/software-mansion/protostar/pull/947/commits/132281c70cf42214ab103b2a2584ab3c9efe4d41)

website/docs/tutorials/07-testing/02-cheatcodes/example.md
```diff
@@ -0,0 +1,44 @@
+# `example`
+
+```python
+def example(self, *args: Any, **kwargs: Any) -> None:
```
MaksymilianDemitraszek, line 4: I wonder if `parametrize` wouldn't be a better name as we are already inspired by pytest heavily

website/docs/tutorials/07-testing/02-cheatcodes/given.md
```diff
@@ -31,3 +31,6 @@ func test_less_equal_compare{syscall_ptr : felt*, range_check_ptr}(a : felt, b :
     return ();
 }
 ```
+
+Note: There is also the [`example`](./example.md) cheatcode that tests certain cases.
+This may come handy if there is a need to test specific edge cases.
```
MaksymilianDemitraszek, line 36:
```suggestion
:::info
There is also the [`example`](./example.md) cheatcode that tests explicitly provided cases. It may come handy if you don't want to rely on randomization.
:::
```

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: Not sure why I just wanted to keep this as a decorator with the annotation syntax `@xxx`

website/docs/tutorials/07-testing/02-cheatcodes/example.md
```diff
@@ -0,0 +1,44 @@
...
+Tests with the data specified in `example` will always be run before applying the data from `given`.
+
+Multiple `example`s can be used for one test. In such a case, all of them will be run sequentially,
+only after that allowing `given` to run.
+
+`example` is not limited by [`max_examples`](./max-examples.md) and is not connected to it in any way.
+
+:::warning
+This cheatcode is only available in [setup cases](../README.md#setup-case).
+:::
```
MaksymilianDemitraszek, line 20: I wouldn't present it as an addition to `given` cheatcode. It is a standalone feature. Please make the first sentence explanation what the cheatcode does as it is the most important information. 
I provided an simplified description, lmk what do you think

```suggestion
Parametrizes test with explicitly provided data. You can provide multiple examples for one test. In such a case, they run sequentially.

Even when examples are provided protostar still runs fuzz tests against the test case. Protostar does it after running all of the examples.

:::warning
This cheatcode is only available in [setup cases](../README.md#setup-case).
:::
:::warning
`example` is not limited by [`max_examples`](./max-examples.md) and is not connected to it in any way.
:::

```

website/docs/tutorials/07-testing/02-cheatcodes/example.md
```diff
@@ -0,0 +1,44 @@
...
+
+```cairo title="Example"
+@external
+func setup_less_equal_compare() {
+    %{
+        example(1, 2)
+        example(2, 3)
+
+        given(a = strategy.integers(15, 20), b = strategy.integers(10, 14))
```
MaksymilianDemitraszek, line 29:
```suggestion
        # it affects the fuzzing strategy applied after all of the examples
        given(a = strategy.integers(15, 20), b = strategy.integers(10, 14))
```

website/docs/tutorials/07-testing/02-cheatcodes/example.md
```diff
@@ -0,0 +1,44 @@
...
+:::warning
+This cheatcode is only available in [setup cases](../README.md#setup-case).
+:::
+
+```cairo title="Example"
+@external
+func setup_less_equal_compare() {
+    %{
+        example(1, 2)
+        example(2, 3)
```
MaksymilianDemitraszek, line 27:
```suggestion
```

website/docs/tutorials/07-testing/02-cheatcodes/example.md
```diff
@@ -0,0 +1,44 @@
...
+
+        given(a = strategy.integers(15, 20), b = strategy.integers(10, 14))
+
+        example(a=3, b=4)
+        example(b=6, a=5)
```
MaksymilianDemitraszek, line 32:
```suggestion
```

website/docs/tutorials/07-testing/02-cheatcodes/example.md
```diff
@@ -0,0 +1,44 @@
...
+    %}
+    return ();
+}
+
+@external
+func test_examples{syscall_ptr: felt*, range_check_ptr}(a: felt, b: felt) {
+    assert_le(0, a);
```
MaksymilianDemitraszek, line 39: [Nitpick] Maybe it is worth to include it:
```suggestion
    # Tested against:
    # (1,2)
    # (3,4)
    # fuzzing: a = strategy.integers(15, 20), b = strategy.integers(10, 14) 
    assert_le(0, a);
```

MaksymilianDemitraszek (changes requested):
Reviewed only the documentation

website/docs/tutorials/07-testing/02-cheatcodes/example.md
karol-bisztyga: @mkaput What do you think? You're the original author of the issue for this feature.
`parametrize` would come from pytest
`example` from hypothesis

website/docs/tutorials/07-testing/02-cheatcodes/example.md
karol-bisztyga: > Even when examples are provided protostar still runs fuzz tests against the test case. Protostar does it after running all of the examples.

Actually, I handled this in the latest commit and it doesn't work like that anymore. Since we want to make `example` a standalone cheatcode, it should be possible to run it without `given` and just get the examples we want. Thanks for the rest, I will come up with a new version.

website/docs/tutorials/07-testing/02-cheatcodes/example.md
karol-bisztyga: I'd remove the other example `example(a=3, b=4)`, I think it's good to add an example with reversed args (`b` is first, then `a`)

website/docs/tutorials/07-testing/02-cheatcodes/example.md
karol-bisztyga: I didn't see it in other docs (but maybe I didn't look close enough), but I don't mind adding it.

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -130,7 +135,6 @@ def build_and_run_test(
     ):
         try:
 
-            @multiply_decorator(example, self.state.config.fuzz_examples)
```
mkaput: nitpick

```python
@self.decorate_with_examples
```

should work

but I think this would be even better:

```python
def many_examples(examples):
    def inner(f):
        for ex in reversed(examples):
            f = example(ex)(f)
        return f
    return inner

# here
@many_examples(self.state.config.fuzz_examples)
...
```

no need to test `many_examples` in this scenario, integration tests do the job already

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -127,6 +133,18 @@ def build_and_run_test(
         given_strategies: Dict[str, SearchStrategy],
     ):
         try:
+            phases = (
+                Phase.explicit,
+                Phase.reuse,
+                Phase.generate,
+                Phase.target,
+                Phase.shrink,
+            )
```
mkaput, line 142:
```suggestion
            phases = tuple(Phase)
```

protostar/testing/test_config_test.py
```diff
@@ -8,6 +8,9 @@
     [
         (TestMode.STANDARD, TestMode.FUZZ, True),
         (TestMode.FUZZ, TestMode.STANDARD, False),
+        (TestMode.STANDARD, TestMode.PARAMETERIZED, True),
+        (TestMode.PARAMETERIZED, TestMode.FUZZ, True),
+        (TestMode.FUZZ, TestMode.PARAMETERIZED, True),
```
mkaput, line 13: please add a test which checks that 
```python
TestMode.FUZZ.convert_to(TestMode.PARAMETERIZED) is TestMode.FUZZ
```

tests/integration/fuzzing/parameterized_test.cairo
mkaput: now that you manipulate `phases`, you should be able to write a test as suggested by @kasperski95 :)

tests/integration/fuzzing/fuzzing_test.py
```diff
@@ -122,3 +122,24 @@ async def test_should_not_share_state(
...
+        expected_passed_test_cases_names=[
+            "test_examples",
+            "test_examples_without_given",
+        ],
+    )
+
+    assert len(testing_summary.passed) == 2
+    passed_set = {passed.fuzz_runs_count for passed in testing_summary.passed}
+    assert 2 in passed_set
+    assert 7 in passed_set
```
mkaput, line 145: this will be unneeded when you assert exact values in test case Cairo code

website/docs/tutorials/07-testing/02-cheatcodes/example.md
mkaput: Pytest's `parameterize` has different API than ours. Their approach is more like `examples` (notice plural form), when you declare multiple input sets at once. Here we define inputs one by one hence I think `example` is better suit. 

also `parameters` doesn't sound as nice as `example` imho here.

mkaput (changes requested):

website/docs/tutorials/07-testing/02-cheatcodes/example.md
karol-bisztyga: [update docs](https://github.com/software-mansion/protostar/pull/947/commits/75b02de8f52d7eced8f64ae8687db4947c59bb45)

website/docs/tutorials/07-testing/02-cheatcodes/example.md
karol-bisztyga: [update docs](https://github.com/software-mansion/protostar/pull/947/commits/75b02de8f52d7eced8f64ae8687db4947c59bb45)

website/docs/tutorials/07-testing/02-cheatcodes/example.md
karol-bisztyga: [update docs](https://github.com/software-mansion/protostar/pull/947/commits/75b02de8f52d7eced8f64ae8687db4947c59bb45)

website/docs/tutorials/07-testing/02-cheatcodes/example.md
karol-bisztyga: [update docs](https://github.com/software-mansion/protostar/pull/947/commits/75b02de8f52d7eced8f64ae8687db4947c59bb45)

website/docs/tutorials/07-testing/02-cheatcodes/given.md
karol-bisztyga: [update docs](https://github.com/software-mansion/protostar/pull/947/commits/75b02de8f52d7eced8f64ae8687db4947c59bb45)

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: the default value for this is actually `set(Phase) - {Phase.explain}`

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga:
```
@self.decorate_with_examples
```
Yes, this works.

As for the other example you provided (with `many_examples`), it doesn't work. For me it looks like there are a lot of ways of doing this, we already managed to have a small amount of code that does a job and I wouldn't spend too much time on this at this point. I think the most beneficial approach would be to just choose one of the possibilities and move on.

tests/integration/fuzzing/fuzzing_test.py
karol-bisztyga: I don't know, I'd leave this but if you see a redundancy here, I'm ok with removing any of these checks.

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: [use the example decorator as annotation](https://github.com/software-mansion/protostar/pull/947/commits/525b1dc8eeb3377c4700c9dd146a9e354208200a)

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: [simplify phases](https://github.com/software-mansion/protostar/pull/947/commits/3882f41e145d8acf1f71d031b4159c18b0a163e1)

protostar/testing/test_config_test.py
karol-bisztyga: [forbid converting fuzz test to parameterized test](https://github.com/software-mansion/protostar/pull/947/commits/7026c691fe9502aff8de957e10b3975f253f285c)

tests/integration/fuzzing/parameterized_test.cairo
karol-bisztyga: [add a check for the exact example value](https://github.com/software-mansion/protostar/pull/947/commits/d60dcaa9c152af075bfeb0964477879fcb4da6a6)

protostar/testing/environments/fuzz_test_execution_environment.py
mkaput: yep, hence I was nitpicking :)

mkaput (commented):
looks like we're close 👏🏻

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: Neither of these approaches turned out to be completely clean.
- `settings.default` makes `pyright` complain, just like using `not_set` (different error being yield ofc)
- I could not really operate on the `settings object`, I wanted to use the default value for the `phases` and modify it conditionally. Unfortunately: `AttributeError: settings objects are immutable`.

I think the best way to go here is to just use `# type: ignore` and move on.

karol-bisztyga: [use default phases](https://github.com/software-mansion/protostar/pull/947/commits/8e0e9577459959a5e70c83373ca172e84a07ed29)

mkaput: as discussed privately, there's a way to "mutate" settings, let's do that here

mkaput (approved):
Code is LGTM 🚀 
@MaksymilianDemitraszek please conclude docs review :)

website/docs/tutorials/07-testing/02-cheatcodes/example.md
MaksymilianDemitraszek: Ok, let's leave example then

website/docs/tutorials/07-testing/02-cheatcodes/example.md
```diff
@@ -0,0 +1,43 @@
...
+
+```python
+def example(self, *args: Any, **kwargs: Any) -> None:
+```
+
+Parametrizes test with explicitly provided data.
+You can provide multiple examples for one test.
+In such a case, they are run sequentially.
+
+`example` can be used next to the [`given`](./given.md) cheatcode.
```
MaksymilianDemitraszek, line 11: Is protostar still going to fuzz when there is no `given` cheatcode?
If yes -> this explanation doesn't capture this behaviour
If not -> behaviour is inconsistent with standard fuzz tests

website/docs/tutorials/07-testing/02-cheatcodes/example.md
```diff
@@ -0,0 +1,43 @@
...
+        example(1, 2)
+        # it affects the fuzzing strategy applied after all of the examples
+        given(a = strategy.integers(15, 20), b = strategy.integers(10, 14))
+        example(b=6, a=5)
+    %}
+    return ();
+}
+
+@external
+func test_examples{syscall_ptr: felt*, range_check_ptr}(a: felt, b: felt) {# Tested against:
```
MaksymilianDemitraszek, line 34:
```suggestion
func test_examples{syscall_ptr: felt*, range_check_ptr}(a: felt, b: felt) {
    # Tested against:
```

website/docs/tutorials/07-testing/02-cheatcodes/example.md
mkaput: > If not -> behaviour is inconsistent with standard fuzz tests

hmm you're quite right, but I wonder how to enforce doing parameterized testing without fuzzing in other way 🤔

mkaput: We want to support all 4 combinations:
1. standard tests
2. fuzz-only tests
3. parameterized-only tests
4. parameterized-and-then-fuzzed tests

Each of your ideas blocks some of these 4 combos. Current API allowed everything:

1. is trivial
2. use only `given`
3. use only `example`
4. use both `example` and `given` in single test

I think maybe this isn't actually so inconsistent. If you use a cheatcode, then you explicitly ask Protostar to do run specified mode, and you can combine modes naturally.

karol-bisztyga: > Is protostar still going to fuzz when there is no `given` cheatcode?

No, in such a case only `examples` will be taken under consideration.

Ok, so what's your suggestion here? We discussed somewhere in the comments that if we use `example`s without `given`, then we only want to have these examples executed, unlike when we do not use either of these cheatcodes - in this case, there's going to be a default strategy for `given` provided implicitly by protostar.

I'm going to add an explicit statement about that.

karol-bisztyga: [add an explicit statement about using example without given](https://github.com/software-mansion/protostar/pull/947/commits/b1acac1d73d7e7a36dbfa541ac3e53e124749a77)

karol-bisztyga: Sorry, I couldn't see the newest responses here.

It all makes sense to me anyway. Are there any more doubts here?

MaksymilianDemitraszek: I still have doubts being honest. We made a mistake designing initial verision of fuzzing, making it default strategy. No framework does that as far as I am aware. I would prefer to make fuzzing always explicit.

I created a issue to discuss that:

Approving, because the issue has nothing to do with your PR I believe.
Thanks for explicit statement, it makes things much cleaner.

MaksymilianDemitraszek (approved):

website/docs/tutorials/07-testing/02-cheatcodes/example.md
mkaput: Resolving this conversation in favour of Maks's task

PR #952, opened 2022-10-10
Update 01-cli.md

ClementWalter:
Suggest the use of _e.g._ (meaning "for example"), instead of _i.e._ (meaning "that is/which means") for wallet suggestion


mkaput:
Submitted #996, as for now, let's proceed merging that.

PR #967, opened 2022-10-10
Refactor arg types

kasperski95:
The PR splits common and Protostar specific types in order to prepare the code for extracting a new module from `protostar.cli` and break a cyclic dependency in https://github.com/software-mansion/protostar/pull/949.


protostar/cli/cli_app.py
```diff
@@ -1,16 +1,20 @@
...
-from protostar.cli.command import Command
+from .arg_type import ArgTypeName
+from .argument import Argument
+from .command import Command
 
+ArgTypeNameT = TypeVar("ArgTypeNameT", bound=ArgTypeName)
 
-class CLIApp:
+
+class CLIApp(Generic[ArgTypeNameT]):
```
mkaput, line 10: I find making this generic just for the sake of testing superfluous:
1. A lot of complex changes in this PR is caused solely by this.
2. The tests could be perfectly fine if they used well known argument types.
3. Tests are leaning towards testing internals of `CLIApp` instead of testing it like a black box.
4. It kinda doesn't make sense by intuition?

docs_generator/reference_docs_generator.py
```diff
@@ -11,7 +11,7 @@ def save_to_markdown_file(filepath: Path, content: str):
         with open(filepath, "w", encoding="utf_8") as file:
             file.write(content)
 
-    def __init__(self, cli_app: CLIApp) -> None:
+    def __init__(self, cli_app: CLIApp[Any]) -> None:
```
mkaput, line 14: This highlights the issue with `CLIApp` being generic IMHO.

mkaput (changes requested):
partial review

mkaput:
I don't really like that we reimplement argument parsing on our own, instead of using a battery included library like https://typer.tiangolo.com/ or https://click.palletsprojects.com/

kasperski95:
> I don't really like that we reimplement argument parsing on our own, instead of using a battery included and battle tested library like https://typer.tiangolo.com/ or https://click.palletsprojects.com/.

> I'm raising this concern now because this PR makes our argparsing even more esoteric.

The configuration file and CLI reference docs relies on the current system.

protostar/cli/cli_app.py
kasperski95: > I find making this generic just for the sake of testing superfluous:

I am not making this generic for the sake of testing, but in order to decouple CLI logic from Protostar specific code (e.g. `ProtostarArgType`, `ProtostarCLI`), extracting a new module in the future PR and preserving type correctness.

> The tests could be perfectly fine if they used well known argument types.

It's not about tests.

> Tests are leaning towards testing internals of CLIApp instead of testing it like a black box.

Refactoring tests isn't in the scope of this PR.

docs_generator/reference_docs_generator.py
kasperski95: Fixed

mkaput (commented):
wdyt about abandoning type safety of being generic on `ArgTypeName` over simply enforcing it on all CliApp instances?

docs_generator/reference_docs_generator.py
```diff
@@ -22,6 +22,7 @@ def generate_cli_reference_markdown(self) -> str:
             result += self._generate_args_markdown(self.cli_app.root_args)
 
         if len(self.cli_app.commands) > 0:
+            # pyright: reportUnknownLambdaType=false
```
mkaput, line 25: is this relevant anymore?

protostar/commands/deploy_command.py
```diff
@@ -38,7 +39,7 @@ def example(self) -> Optional[str]:
         return "protostar deploy ./build/main.json --network alpha-goerli"
 
     @property
-    def arguments(self) -> List[Command.Argument]:
+    def arguments(self):
         return [
             Command.Argument(
```
mkaput, line 44: Why not use ProtostarArgument everywhere to be consistent?

kasperski95:
> wdyt about abandoning type safety of being generic on ArgTypeName over simply enforcing it on all CliApp instances?

Removed generic from `CLIApp`.

protostar/cli/argument.py
```diff
@@ -0,0 +1,20 @@
...
+    description: str
+    type: ArgTypeNameT
+    is_positional: bool = False
+    is_required: bool = False
+    is_array: bool = False
+    default: Any = None
+    example: Optional[str] = None
+    short_name: Optional[str] = None
+
+    def copy_with(self, **changes) -> "Argument":
```
mkaput, line 19: 1. hmm why not simply use `dataclasses.replace` everywhere by convention?
2. if not, could this return `-> Self` ?

protostar/commands/deploy_command.py
kasperski95: I want to keep the PR small. I'll do it later.

docs_generator/reference_docs_generator.py
kasperski95: It is.

protostar/cli/argument.py
kasperski95: It's old code.
1) Flutter's `copy_with` seemed more natural to me back then.
2) I didn't know about `Self` type back then.

I try to keep this PR focused.

protostar/commands/deploy_command.py
mkaput: that's fine

protostar/cli/argument.py
mkaput: ok, gotcha

mkaput (approved):
lgtm

PR #973, opened 2022-10-11
Add .gitignore to cache

karol-bisztyga:
Addign `.gitignore` to the protostar cache as requested in https://github.com/software-mansion/protostar/pull/895#discussion_r991032941

### DIFF
--- a/protostar/self/cache_io.py
+++ b/protostar/self/cache_io.py
@@ -10,6 +10,10 @@ class CacheIO:
     def __init__(self, project_root_path: Path):
         self._cache_path = project_root_path / Path(self._CACHE_DIR_NAME)
         self._cache_path.mkdir(exist_ok=True)
+        gitignore_path = Path(self._cache_path / ".gitignore")
+        if not gitignore_path.exists():
+            gitignore_path.touch()
+            gitignore_path.write_text("*\n", encoding="utf-8")
 
     def write(self, name: str, value: dict) -> None:
         Path(self._cache_path / (name + self._EXTENSION)).write_text(


protostar/self/cache_io.py
```diff
@@ -10,6 +10,10 @@ class CacheIO:
     def __init__(self, project_root_path: Path):
         self._cache_path = project_root_path / Path(self._CACHE_DIR_NAME)
         self._cache_path.mkdir(exist_ok=True)
+        gitignore_path = Path(self._cache_path / ".gitignore")
+        if not gitignore_path.exists():
+            gitignore_path.touch()
+            gitignore_path.write_text("*\n", encoding="utf-8")
```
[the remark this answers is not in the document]
mkaput: This doesn't retrofit onto projects that are already created. The approach I suggested works smoothly in more cases.

mkaput (approved):

protostar/self/cache_io.py
```diff
@@ -10,6 +10,10 @@ class CacheIO:
     def __init__(self, project_root_path: Path):
         self._cache_path = project_root_path / Path(self._CACHE_DIR_NAME)
         self._cache_path.mkdir(exist_ok=True)
+        gitignore_path = Path(self._cache_path / ".gitignore")
+        if not gitignore_path.exists():
+            gitignore_path.touch()
```
mkaput, line 15:
```suggestion
```

nitpick: https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text write_text creates the file if missing

protostar/self/cache_io.py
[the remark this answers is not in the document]
kasperski95: Ok, let's stick with the Marek's approach. I'd move this logic to the write method. Constructors should be as light as possible.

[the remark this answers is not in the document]
karol-bisztyga: Are you okay with having a flag and creating the `.gitignore` only once during the `CacheIO` object lifetime? I don't think we should do an additional `write_text` every time we call the `write` method (`write_text` has to do file i/o ops which are slow and it is not necessary to check for the `.gitinore` every time).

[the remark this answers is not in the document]
kasperski95: You can keep constructor light and checking file existence once per object lifetime by introducing a new property:
```python
# write method...
if not self._was_gitignore_existenece_checked not gitignore_path.exists():
    self._was_gitignore_existenece_checked = True
    # ...
```

kasperski95 (commented):
(Responded in the thread.)

protostar/self/cache_io.py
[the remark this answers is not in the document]
karol-bisztyga: [move creating file away from the ctor](https://github.com/software-mansion/protostar/pull/973/commits/7ec8d2301a3df5005ede42175c8482a4a262d9cf)

protostar/self/cache_io.py
karol-bisztyga: [remove touch](https://github.com/software-mansion/protostar/pull/973/commits/170756a31b730e90fb1707f6a8b6fa48db37d22f)

protostar/self/cache_io.py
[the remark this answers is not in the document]
kasperski95: > That was exactly what I meant when I said "flag"

Sorry, I overlooked that part.

kasperski95 (approved):

protostar/self/cache_io.py
```diff
@@ -10,8 +10,13 @@ class CacheIO:
     def __init__(self, project_root_path: Path):
         self._cache_path = project_root_path / Path(self._CACHE_DIR_NAME)
         self._cache_path.mkdir(exist_ok=True)
+        self._gitignore_path = Path(self._cache_path / ".gitignore")
+        self._gitignore_checked = False
 
     def write(self, name: str, value: dict) -> None:
+        if self._gitignore_checked and not self._gitignore_path.exists():
```
mkaput, line 17: shouldn't this look like that?

```suggestion
        if not self._gitignore_checked and not self._gitignore_path.exists():
```

mkaput (changes requested):

mkaput (approved):
Hmm could we have a test that checks this behaviour? Try to write one, if it happens to be overly complex then feel free to merge this as-is.

protostar/self/cache_io_test.py
```diff
@@ -21,6 +21,9 @@ def test_cache_simple(tmp_path):
 
     assert cache_io.read(cache_name) is None
 
+    # pylint: disable=protected-access
+    assert cache_io._gitignore_path.read_text(encoding="utf-8") == "*\n"
```
mkaput, line 25: Either:
1. Consider making this field public
2. Preferred: assume the path in the test (use hardcoded value instead of reading the field)

mkaput (changes requested):

protostar/self/cache_io_test.py
mkaput, line 25: I think it would be also nice to verify that modification timestamp of the file does not change during subsequent writes.

mkaput (changes requested):

protostar/self/cache_io_test.py
karol-bisztyga: Why exactly is this code a problem?
1. I can do that
2. I think this is a no-go, because `_gitignore_path` evaluates to `Path(self._cache_path / ".gitignore")`, which evaluates to `Path((project_root_path / Path(self._CACHE_DIR_NAME)) / ".gitignore")`, so we'd eventually have to make something public anyway (in this case `_CACHE_DIR_NAME`)

I think that accessing private members in tests is okay - you want to test if everything is okay at the same having a clean structure in the prod.

mkaput: > I think that accessing private members in tests is okay - you want to test if everything is okay at the same having a clean structure in the prod.

I don't fully share this opinion, as I believe in writing tests that test a black box which exposes a public API. I think exposing a property pointing to the cache directory would make sense.

mkaput:
As discussed privately, the behaviour of this file management has to be changed to handle cases where user edits it further in intuitive way.

protostar/self/cache_io_test.py
karol-bisztyga: [check if the gitignore is not modified when already created](https://github.com/software-mansion/protostar/pull/973/commits/c62c8ddcce7665e812bccb440f87331c811d407d)

protostar/self/cache_io.py
```diff
@@ -10,8 +10,12 @@ class CacheIO:
     def __init__(self, project_root_path: Path):
         self._cache_path = project_root_path / Path(self._CACHE_DIR_NAME)
         self._cache_path.mkdir(exist_ok=True)
+        self._gitignore_path = Path(self._cache_path / ".gitignore")
+        self._gitignore_checked = False
```
mkaput, line 14:
```suggestion
```

karol-bisztyga: [remove redundant line](https://github.com/software-mansion/protostar/pull/973/commits/bf5a56bdcbcfb6b0bcf5ca9b2f45b14f48011e6a)

mkaput (approved):

PR #977, opened 2022-10-12
Add Uint256 strategy

Radinyn:


protostar/testing/fuzzing/strategies/uint256.py
```diff
@@ -20,7 +20,9 @@ def _get_high(uint256: int):
 
 class Uint256StrategyDescriptor(StrategyDescriptor):
     def build_strategy(self, cairo_type: CairoType) -> SearchStrategy[int]:
-        if not isinstance(cairo_type, TypeStruct):  # TODO: FIND BETTER CHECK
+        if not isinstance(cairo_type, TypeStruct) and str(cairo_type.scope).endswith(
```
mkaput, line 23: I would test fully qualified path, so that user-defined structs named the same won't be picked up by this

protostar/testing/fuzzing/strategies/uint256.py
```diff
@@ -0,0 +1,32 @@
...
+from starkware.cairo.lang.compiler.ast.cairo_types import CairoType, TypeStruct
+
+from protostar.testing.fuzzing.exceptions import SearchStrategyBuildError
+from protostar.testing.fuzzing.strategy_descriptor import StrategyDescriptor
+
+
+def _get_low(uint256: int):
+    mask = (1 << 128) - 1  # 128*"1"
+    mask = mask << 128  # 128*"1"+128*"0"
+    mask = ~mask  # 128*"0"+128*"1"
```
mkaput, line 13: won't `2 ** 128 -1` give what you want?

protostar/testing/fuzzing/strategies/uint256.py
Radinyn: After manual testing full path seems to be pretty inconsistent, I think the current approach is good enough.

protostar/testing/fuzzing/strategy_descriptor.py
```diff
@@ -1,11 +1,11 @@
...
-from typing import Callable
+from typing import Callable, Any
 
 from hypothesis.strategies import SearchStrategy
 from starkware.cairo.lang.compiler.ast.cairo_types import CairoType
 
-MappingFunction = Callable[[int], int]
-FilterFunction = Callable[[int], bool]
+MappingFunction = Callable[[Any], Any]
+FilterFunction = Callable[[Any], bool]
```
Arcticae, line 8: Having `Any` here disables basically all checks for the args. Is that the only possible way?

protostar/testing/fuzzing/strategy_inference.py
```diff
@@ -11,6 +15,8 @@ def infer_strategy_from_cairo_type(
 ) -> StrategyDescriptor:
     if isinstance(cairo_type, TypeFelt):
         return FeltsStrategyDescriptor()
+    if isinstance(cairo_type, TypeStruct) and str(cairo_type.scope).endswith("Uint256"):
```
Arcticae, line 18: Can identification be performed in some other way? Like the structure itself, not the name?

Arcticae (approved):
LGTM. Some nitpicks below

Arcticae:
I'd also ask @mkaput since he's the fuzzing specialist 😬

protostar/testing/fuzzing/strategy_descriptor.py
mkaput: Eventually these types were doomed to use `Any` (they originally were when I wrote that, but during code review we changed this to `int`).

Technically we could use some generics here, @Radinyn you could experiment with that here.

protostar/testing/fuzzing/strategy_inference.py
mkaput: @Arcticae See this comment thread: https://github.com/software-mansion/protostar/pull/977#discussion_r995783576

protostar/testing/fuzzing/strategies/uint256.py
mkaput: What's inconsistent here? Do you observe different paths to the type? 🤨

tests/integration/fuzzing_strategies/uint256_test.cairo
```diff
@@ -0,0 +1,14 @@
...
+@external
+func test_uint256_fail{range_check_ptr}(a: Uint256) {
+    alloc_locals;
+    let b: Uint256 = Uint256(0, 1);
+    let condition: felt = uint256_le(b, a);
+    if (condition == 1) {
+        assert 1 = 0;
+    }
+    return ();
+}
```
mkaput, line 14:
```suggestion
@external
func setup_uint256{range_check_ptr}() {
    %{
        max_examples(3)
        given(a=strategies.uint256(min_value=2))
    %}
    return ();
}

@external
func test_uint256{range_check_ptr}(a: Uint256) {
    tempvar b: Uint256 = Uint256(0, 1);
    tempvar condition: felt = uint256_le(b, a);
    assert condition = 1;
    return ();
}
```

tests/integration/fuzzing_strategies/uint256_test.cairo
```diff
@@ -0,0 +1,14 @@
+%lang starknet
+
+from starkware.cairo.common.uint256 import Uint256, uint256_le
+
+@external
+func test_uint256_fail{range_check_ptr}(a: Uint256) {
```
mkaput, line 6: I would like to see a test that uses `map` and `filter`.

protostar/testing/fuzzing/strategies/uint256.py
```diff
@@ -0,0 +1,27 @@
...
+    def build_strategy(self, cairo_type: CairoType) -> SearchStrategy[tuple[int, int]]:
+        if not isinstance(cairo_type, TypeStruct) or not str(cairo_type.scope).endswith(
+            "Uint256"
+        ):
+            raise SearchStrategyBuildError(
+                "Strategy 'uint256' can only be applied to Uint256 parameters."
+            )
+
+        return integers(min_value=0, max_value=(1 << 256) - 1).map(
+            lambda x: (_get_low(x), _get_high(x))
```
mkaput, line 26: Can you use a `NamedTuple` here? In the future I would love to have a possibility to operate on these pairs like on normal integers in `map` and `filter` functions

mkaput (changes requested):

protostar/testing/fuzzing/strategies/uint256.py
Radinyn: It turns out that `starknet_py.cairo.felt.is_uint256` is using CairoTypes from ABI which has a different naming scheme.

protostar/testing/fuzzing/strategy_descriptor.py
Radinyn: I mean, although the user doesn't see typing, it makes more sense in my opinion to use `Any`, after all the mapping function doesn't have to return the same type.

protostar/testing/fuzzing/strategies/uint256.py
```diff
@@ -13,15 +18,31 @@ def _get_high(uint256: int):
...
+    )
+
+
 class Uint256StrategyDescriptor(StrategyDescriptor):
+    def __init__(
+        self, min_value: Optional[int] = None, max_value: Optional[int] = None
+    ):
+        min_value = MIN_UINT256 if min_value is None else min_value
+        max_value = MAX_UINT256 if max_value is None else max_value
+        assert MIN_UINT256 <= min_value <= max_value <= MAX_UINT256
```
mkaput, line 33: This assertion is not raising meaningful exceptions

mkaput (changes requested):

mkaput (changes requested):
This code looks OK but as discussed internally, we should defer creating of `uint256` tuple after all filters and maps and one_ofs.

Radinyn:
As discussed on the meeting, will be merged in it's current state.

MaksymilianDemitraszek:
@Radinyn Are docs updated about this? Also if it is a breaking change we should mention it in changelog

MaksymilianDemitraszek:
@Radinyn Can you create task for this and add it to this milestone?

PR #981, opened 2022-10-13
Suggest adding cairo-path to the protostar.toml after adding a package

### LINKED ISSUE #236: Suggest adding `cairo-path` to the `protostar.toml` after adding a package

karol-bisztyga:
fix #236

### DIFF
--- a/protostar/commands/install/install_command.py
+++ b/protostar/commands/install/install_command.py
@@ -101,8 +101,8 @@ class InstallCommand(ProtostarCommand):
                 destination=libs_path,
                 tag=package_info.version,
             )
-            ConfigurationFile.create_appending_cairo_path_suggestion(
-                self._logger, self._project_root_path, libs_path, str(package_info.name)
+            self._logger.info(
+                ConfigurationFile.create_appending_cairo_path_suggestion()
             )
         else:
             pull_package_submodules(
--- a/protostar/commands/install/install_command.py
+++ b/protostar/commands/install/install_command.py
@@ -1,6 +1,6 @@
 from logging import Logger
 from pathlib import Path
-from typing import Callable, List, Optional
+from typing import Callable, Optional
 import tomli
 import tomli_w
 
@@ -138,6 +138,5 @@ class InstallCommand(ProtostarCommand):
             f"`{protostar_toml}` file"
         )
         self._logger.info(
-            f"\n[{shared_command_configs_key}]\n"
-            + tomli_w.dumps(shared_command_configs)
+            f"\n[{shared_command_configs_key}]\n{tomli_w.dumps(shared_command_configs)}"
         )
--- a/protostar/commands/install/install_command.py
+++ b/protostar/commands/install/install_command.py
@@ -1,8 +1,6 @@
 from logging import Logger
 from pathlib import Path
 from typing import Callable, Optional
-import tomli
-import tomli_w
 
 from protostar.cli import ProtostarArgument, ProtostarCommand
 from protostar.commands.install.install_package_from_repo import (
@@ -12,6 +10,7 @@ from protostar.commands.install.pull_package_submodules import pull_package_subm
 from protostar.io.log_color_provider import LogColorProvider
 from protostar.package_manager import extract_info_from_repo_id
 from protostar.protostar_toml.protostar_project_section import ProtostarProjectSection
+from protostar.configuration_file import ConfigurationFile
 
 EXTERNAL_DEPENDENCY_REFERENCE_DESCRIPTION = """- `GITHUB_ACCOUNT_NAME/REPO_NAME[@TAG]`
     - `OpenZeppelin/cairo-contracts@v0.4.0`
@@ -102,7 +101,9 @@ class InstallCommand(ProtostarCommand):
                 destination=libs_path,
                 tag=package_info.version,
             )
-            self.suggest_adding_cairo_path(libs_path, str(package_info.name))
+            ConfigurationFile.create_appending_cairo_path_suggestion(
+                self._logger, self._project_root_path, libs_path, str(package_info.name)
+            )
         else:
             pull_package_submodules(
                 on_submodule_update_start=lambda package_info: self._logger.info(
@@ -117,26 +118,3 @@ class InstallCommand(ProtostarCommand):
                 repo_dir=self._project_root_path,
                 libs_dir=libs_path,
             )
-
-    def suggest_adding_cairo_path(self, libs_path, package_name):
-        protostar_toml = "protostar.toml"
-        shared_command_configs_key = "protostar.shared_command_configs"
-
-        data = tomli.loads(
-            (self._project_root_path / protostar_toml).read_text(encoding="utf-8")
-        )
-        shared_command_configs = data.get(shared_command_configs_key)
-        if not shared_command_configs:
-            shared_command_configs = {}
-        shared_command_configs["cairo_path"] = (
-            ["..."] if shared_command_configs.get("cairo_path") else []
-        )
-        shared_command_configs["cairo_path"].append(f"{libs_path}/{package_name}/src")
-
-        self._logger.info(
-            "You may want to add your new library's path to the 'cairo_path' section of your "
-            f"`{protostar_toml}` file"
-        )
-        self._logger.info(
-            f"\n[{shared_command_configs_key}]\n{tomli_w.dumps(shared_command_configs)}"
-        )
--- a/protostar/commands/install/install_command.py
+++ b/protostar/commands/install/install_command.py
@@ -1,6 +1,8 @@
 from logging import Logger
 from pathlib import Path
-from typing import Callable, Optional
+from typing import Callable, List, Optional
+import tomli
+import tomli_w
 
 from protostar.cli import ProtostarArgument, ProtostarCommand
 from protostar.commands.install.install_package_from_repo import (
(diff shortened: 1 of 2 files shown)


protostar/commands/install/install_command.py
```diff
@@ -114,3 +117,27 @@ def install(
                 repo_dir=self._project_root_path,
                 libs_dir=libs_path,
             )
```
karol-bisztyga, line 119: should we also handle this situation?

karol-bisztyga:
We discussed this case with @kasperski95 and decided to give the user just a hint about adding a proper entry to the `protostar.toml`. @mkaput what do you think?

mkaput:
> We discussed this case with @kasperski95 and decided to give the user just a hint about adding a proper entry to the `protostar.toml`. @mkaput what do you think?

Looks ok. We will validate this on prod 😉

protostar/commands/install/install_command.py
karol-bisztyga: Ok, I think I understand this now. So if we run `install xxx`, then package `xxx` is being installed, but if we run just `install`, then the already installed packages are being updated. If I got this right, then we don't have to do anything here. Correct me if I'm wrong.

karol-bisztyga:
Do we need a test for this?

protostar/commands/install/install_command.py
```diff
@@ -114,3 +117,26 @@ def install(
...
+        )
+        shared_command_configs["cairo_path"].append(f"{libs_path}/{package_name}/src")
+
+        self._logger.info(
+            "You may want to add your new library's path to the 'cairo_path' section of your "
+            f"`{protostar_toml}` file"
+        )
+        self._logger.info(
+            f"\n[{shared_command_configs_key}]\n{tomli_w.dumps(shared_command_configs)}"
+        )
```
kasperski95, line 142: [Design] I'd move some code to the Configuration File module e.g. `ConfigurationFile::create_appending_cairo_path_suggestion`.

kasperski95 (commented):
Quick review

kasperski95 (commented):
https://github.com/software-mansion/protostar/pull/981/files#r995587563

protostar/commands/install/install_command.py
karol-bisztyga: [move suggesting function to the ConfigurationFile](https://github.com/software-mansion/protostar/pull/981/commits/2bd3099efe10c13ed4982fd6a56509ebd4288d03)

karol-bisztyga: Sorry I didn't see this

protostar/configuration_file/configuration_file.py
```diff
@@ -86,6 +88,32 @@ def set_command_names_provider(
     ):
         pass
 
+    @classmethod
+    def create_appending_cairo_path_suggestion(
+        cls, logger, project_root_path, libs_path, package_name
+    ):
```
kasperski95, line 94: [Design] `cls` is not used. This method shouldn't log anything but return a string. The method implementation should be located in concrete classes `ConfigurationFileV1` and `ConfigurationFileV2` unless the error message is generic.
 
```suggestion
    @abstractmethod
    def create_appending_cairo_path_suggestion(self) -> str:
```

protostar/configuration_file/configuration_file.py
```diff
@@ -86,6 +88,32 @@ def set_command_names_provider(
...
+        )
+        shared_command_configs["cairo_path"].append(f"{libs_path}/{package_name}/src")
+
+        logger.info(
+            "You may want to add your new library's path to the 'cairo_path' section of your "
+            f"`{protostar_toml}` file"
+        )
+        logger.info(
+            f"\n[{shared_command_configs_key}]\n{tomli_w.dumps(shared_command_configs)}"
+        )
```
kasperski95, line 115: [Complexity/Functionality] Let's make it simpler and generic. Current suggestion could be invalid if the library doesn't have src directory.

For example:
> You may want to update the cairo-path configuration in the configuration file. 
https://docs.swmansion.com/protostar/docs/tutorials/dependencies-management#adding-a-dependency

kasperski95 (changes requested):

protostar/configuration_file/configuration_file.py
karol-bisztyga: > `cls` is not used

`cls` was required so this method can be in this class as far as I remember. We can use the `@staticmethod` here.

> This method shouldn't log anything but return a string

OK

> The method implementation should be located in concrete classes `ConfigurationFileV1` and `ConfigurationFileV2` unless the error message is generic.

Sorry, but I don't understand. It is generic, it doesn't contain anything specific for the concrete implementations, therefore it introduces mutual code for every implementation and thus should not be distributed into different classes.
What would be the purpose of putting separate methods doing basically the same thing in separate derived classes?

protostar/configuration_file/configuration_file.py
karol-bisztyga: I think it's good to provide a small example, not only a link to the docs, if you think otherwise, let me know.

karol-bisztyga: [make the message more generic](https://github.com/software-mansion/protostar/pull/981/commits/bcd081a9ebfdd25525c4f73224afbdbd1ae2366c)

protostar/configuration_file/configuration_file.py
```diff
@@ -86,6 +86,20 @@ def set_command_names_provider(
...
+    @staticmethod
+    def create_appending_cairo_path_suggestion() -> str:
+        return "\n".join(
+            [
+                "You may want to add your new library's path to the 'cairo_path' section of your `protostar.toml` file",
+                "[protostar.shared_command_configs]",
+                "cairo_path = [",
+                '    "...",',
+                '    "path/to/the/library/source/files"',
+                "]",
```
kasperski95, line 98: This config is specific to the ConfigurationFileV1. You can remove the example. The generic message and a link to the documentation should be good enough.

protostar/configuration_file/configuration_file.py
```diff
@@ -86,6 +86,20 @@ def set_command_names_provider(
     ):
         pass
 
+    @staticmethod
+    def create_appending_cairo_path_suggestion() -> str:
+        return "\n".join(
+            [
+                "You may want to add your new library's path to the 'cairo_path' section of your `protostar.toml` file",
```
kasperski95, line 93: You can avoid hard-coding the configuration file name by using `self.get_filepath().name`

kasperski95 (changes requested):

protostar/configuration_file/configuration_file.py
kasperski95: > Sorry, but I don't understand. It is generic, it doesn't contain anything specific for the concrete implementations...

"shared_command_configs" is specific to V1

protostar/configuration_file/configuration_file.py
karol-bisztyga: I see, alright

protostar/configuration_file/configuration_file.py
karol-bisztyga: For this, we'd have to have a `ConfigurationFile` object as this method is static. I'm going to replace this message with something like this:
```
"You may want to add your new library's path to the 'cairo_path' section of your configuration file (usually `protostar.toml`).",
```

protostar/configuration_file/configuration_file.py
karol-bisztyga: [address feedback - make the message more generic](https://github.com/software-mansion/protostar/pull/981/commits/7eab993b8be6705e14f2a27944838915254a6ca9)

protostar/configuration_file/configuration_file.py
kasperski95: > I think it's good to provide a small example, not only a link to the docs, if you think otherwise, let me know.

A message with an example could be a bit too intrusive (a user needs to learn this once and do not forget about it) and harder to implement (especially because we are migrating the configuration file).

protostar/configuration_file/configuration_file.py
```diff
@@ -86,6 +86,17 @@ def set_command_names_provider(
     ):
         pass
 
+    @staticmethod
+    def create_appending_cairo_path_suggestion() -> str:
+        return "\n".join(
+            [
+                "You may want to add your new library's path to the 'cairo_path' "
```
kasperski95, line 93:
```suggestion
                "You may want to add your new library's project root path to the 'cairo_path' "
```

It's not exactly the library's path. If that were a case, we wouldn't need this suggestion.

protostar/configuration_file/configuration_file.py
```diff
@@ -86,6 +86,17 @@ def set_command_names_provider(
...
+                "section of your configuration file (usually `protostar.toml`).",
```
kasperski95, line 94:
```suggestion
                "section of your configuration file.",
```
We can relatively easy get the name of the configuration file (but this method can't be static). We should either print the exact name or remove the guess.
Instead of the file name, a relative file path could be printed so that the user can `cmd+click` it.

kasperski95, line 94: `cairo_path` isn't a section. `cairo-path` is an attribute.

> "You may want to update your `cairo-path` configuration in {configuration_file_name}."
> "For more information ..."

kasperski95 (changes requested):

protostar/configuration_file/configuration_file.py
karol-bisztyga: Ok, in your other comment you said

> We should either print the exact name or remove the guess.

So, with the "guess" removed, we can just address the config file in general as "your config file", right?

karol-bisztyga: [update msg](https://github.com/software-mansion/protostar/pull/981/commits/a998dc15508fe8fab245e741553b08787011c826)

Issue #796, opened 2022-09-12
Collect fee estimation data from test cases


mkaput:
We should provide fee estimation as one of test statistics.

PR #1013, opened 2022-10-18
Estimate fee for test cases

karol-bisztyga:
#796


kasperski95:
@MaksymilianDemitraszek Can you take a look on this PR?

protostar/testing/test_case_runners/estimate_fee.py
```diff
@@ -0,0 +1,37 @@
+from typing import Optional
+
+from protostar.testing.starkware.execution_resources_summary import (
+    ExecutionResourcesSummary,
+)
+
+
+def estimate_fee(
```
MaksymilianDemitraszek, line 8: Please use existing functions for that, discussed on Slack

MaksymilianDemitraszek (changes requested):

mkaput:
> I had a problem with fuzz tests. In the `execution_resources` we have:
> 
> * `builtin_name_to_count_map::range_check_builtin`
> * `n_steps`
> 
> In normal tests, we have just single `value`s. In fuzz tests we have different values (`series`) for `n_steps` and for some of them, the `range_check_builtin` is present, for some it is not. The problem is, we do not know for which tests the `range_check_builtin` is there and for which is not. The `range_check_builtin` `series` is not filled with `0` or `None` or anything else that would indicate that for a certain case we do not have it. Therefore, we will not be able to always know whether a specific case has the `range_check_builtin`.
> 
> We could come up with heuristics for that but I can spot a misleading example right away - that would be when `range_check_builtin` is exactly as long as half of the `n_steps` length (in other words, just half of all the cases). Let's go with an example:
> 
> We have the following `series` for `n_steps`: `[78, 78, 78, 131, 131, 131]` We have the following `series` for `range_check_builtin`: `[1, 1, 1]`
> 
> Now, we cannot know if the range checks are for the cases with `n_steps` of value `78` or `131`.
> 
> That would be possible if we had `series` for `range_check_builtin` (for example): `[0, 0, 0, 1, 1, 1]`
> 
> I see two options here:
> 
> * we do not estimate the fuzz tests
> * maybe there's a way to fill the `range_check_builtin` `series` properly for our use-case.

There is a way to fill series of builtins properly, but it requires total rewrite of ExecutionResourcesSummary class:

1. Let ExecutionResourcesSummary store a list of `samples`, each sample describes execution resources consumed by a single test case (so for non-fuzz tests there will be only one sample ever).
2. Shift the logic of aggregating sampled values outside this. It should be possible to keep the logic within measurements objects, but add a logic to automatically traverse a list of sample objects and build aggregates outside ERS

protostar/testing/test_case_runners/estimate_fee.py
karol-bisztyga: [use existing function](https://github.com/software-mansion/protostar/pull/1013/commits/8b719e9450f2ceffc27fc9b0ec651be237597938)

karol-bisztyga:
@mkaput I handled this differently in [also add estimated fee for fuzz tests](https://github.com/software-mansion/protostar/pull/1013/commits/b648c8932e213660f0fe0e6f84e04effdfa8d687), it was possible after I used an existing function for this task in [use existing function](https://github.com/software-mansion/protostar/pull/1013/commits/46f745ab33051ae4174018eff2b1c5aa8d7ba918) as Maks suggested

protostar/starknet/forkable_starknet.py
```diff
@@ -50,15 +55,34 @@ async def deploy(
...
+            result=(),
+            main_call_events=[],
+        )
+
+        starknet_contract = WrappedStarknetContract(
+            state=self.state,
+            abi=get_abi(contract_class=contract_class),
+            contract_address=address,
+            deploy_call_info=deploy_call_info,
+            actual_resources=execution_info.actual_resources,
```
MaksymilianDemitraszek, line 84: Aren't this result in us printing the `deploy` of the test transaction fee? Not the transaction itself

MaksymilianDemitraszek (changes requested):

protostar/testing/starkware/test_execution_state.py
```diff
@@ -55,13 +55,18 @@ async def from_test_suite_definition(
...
     def fork(self) -> Self:
-        return dataclasses.replace(
+        result = dataclasses.replace(
             super().fork(),
             config=deepcopy(self.config),
             context=deepcopy(self.context),
             output_recorder=self.output_recorder.fork(),
             stopwatch=self.stopwatch.fork(),
         )
+        if result.contract:
```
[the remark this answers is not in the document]
karol-bisztyga: done with [remove wrapped starknet contract and move actual_resources to to Exec…](https://github.com/software-mansion/protostar/pull/1013/commits/5238e631fab86a87eb2b56f79ca034f10085ff18)

protostar/starknet/wrapped_starknet_contract.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(
+        self,
+        state: StarknetState,
+        abi: AbiType,
+        contract_address: CastableToAddress,
+        deploy_call_info: StarknetCallInfo,
+        actual_resources: ResourcesMapping,
+    ):
+        super().__init__(state, abi, contract_address, deploy_call_info)
+        self.actual_resources = actual_resources
```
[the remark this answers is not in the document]
karol-bisztyga: [remove wrapped starknet contract and move actual_resources to to Exec…](https://github.com/software-mansion/protostar/pull/1013/commits/5238e631fab86a87eb2b56f79ca034f10085ff18) is this ok?

protostar/testing/starkware/execution_resources_summary.py
```diff
@@ -74,6 +74,7 @@ class ExecutionResourcesSummary:
     n_steps: Statistic = field(default_factory=CountStatistic)
     n_memory_holes: Statistic = field(default_factory=CountStatistic)
     builtin_name_to_count_map: Dict[str, Statistic] = field(default_factory=dict)
+    estimated_fee: float = field(default_factory=float)
```
Arcticae, line 77: Is that one going to be 0 by default? I think that might be misleading

protostar/testing/testing_summary.py
```diff
@@ -71,6 +72,20 @@ def log(
...
+                isinstance(result, (PassedTestCaseResult, PassedFuzzTestCaseResult))
+                and result.estimated_fee
+            ):
+                estimated_fees[result.test_case_name] = result.estimated_fee
+        if estimated_fees:
+            logger.info(log_color_provider.bold("Estimated fees:".ljust(header_width)))
+            for test_name, estimated_fee in estimated_fees.items():
+                logger.info(
+                    f'    - for a single run of the test "{test_name}": '
+                    f"{estimated_fee} gas ({estimated_fee * 1_000_000_000} WEI)"
```
Arcticae, line 87: i think that the result of that should be output in gas only, gas prices can vary.

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -171,12 +172,15 @@ async def test(**inputs: Any):
...
 
+                is_first_run = runs_counter_initial_value == runs_counter.count
                 run_no = next(runs_counter)
                 with self.state.output_recorder.redirect(("test", run_no)):
                     with with_reporter(protostar_reporter):
                         try:
                             this_run_resources = await self.execute_test_case(
-                                function_name, **inputs
+                                function_name=function_name,
+                                calculate_estimated_fee=is_first_run,
```
Arcticae, line 182: This might prove pretty useless when we want to detect price spikes in some corner cases of the fuzzed inputs

protostar/starknet/forkable_starknet.py
```diff
@@ -51,14 +54,32 @@ async def deploy(
...
+            result=(),
+            main_call_events=[],
+        )
+
+        starknet_contract = StarknetContract(
+            state=self.state,
+            abi=get_abi(contract_class=contract_class),
+            contract_address=address,
+            deploy_call_info=deploy_call_info,
+        )
```
Arcticae, line 82: Why is this code copied over from superclass?

protostar/testing/starkware/execution_resources_summary.py
karol-bisztyga: Yeah, we probably should go with optional here

protostar/testing/testing_summary.py
karol-bisztyga: fixed in [new approach](https://github.com/software-mansion/protostar/pull/1013/commits/3efa247b9c0b19218279da5b017f02f38aff4ebf)

protostar/testing/environments/fuzz_test_execution_environment.py
karol-bisztyga: in [new approach](https://github.com/software-mansion/protostar/pull/1013/commits/3efa247b9c0b19218279da5b017f02f38aff4ebf) I changed the mechanism of fee estimation for fuzzing. Now, the biggest fee encountered is shown in the summary.

protostar/starknet/forkable_starknet.py
karol-bisztyga: it's removed in [new approach](https://github.com/software-mansion/protostar/pull/1013/commits/3efa247b9c0b19218279da5b017f02f38aff4ebf)

protostar/starknet/forkable_starknet.py
karol-bisztyga: [new approach](https://github.com/software-mansion/protostar/pull/1013/commits/3efa247b9c0b19218279da5b017f02f38aff4ebf)

protostar/testing/starkware/execution_resources_summary.py
karol-bisztyga: [make estimated_fee optional](https://github.com/software-mansion/protostar/pull/1013/commits/00572a026aeb8dcf327b3ba8a327d2e80bf00bf3)

protostar/testing/testing_summary.py
```diff
@@ -71,6 +71,17 @@ def log(
             log_color_provider.bold("Seed: ".ljust(header_width))
             + str(self.testing_seed)
         )
+        estimated_fees: dict[str, float] = {}
```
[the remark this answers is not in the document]
karol-bisztyga: I don't think this is related, can I create a separate issue for this?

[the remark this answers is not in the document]
karol-bisztyga: [add estimated fee to a single test result](https://github.com/software-mansion/protostar/pull/1013/commits/4b49deda608038ab2191d5c3e0dc98af7a9f1b45)

BTW should we keep the this log? `- for a single run of the test "{test_name}": {estimated_fee} gas'`

protostar/testing/test_results.py
```diff
@@ -33,6 +33,7 @@ class TimedTestCaseResult(TestCaseResult, TimedTestResult):
 @dataclass(frozen=True)
 class PassedTestCaseResult(TimedTestCaseResult):
     execution_resources: Optional[ExecutionResourcesSummary]
+    estimated_fee: Optional[float] = None
```
MaksymilianDemitraszek, line 36: Why don't make it part of `ExecutionResources`?

protostar/testing/environments/execution_environment.py
```diff
@@ -26,11 +42,39 @@ async def perform_execute(
         self,
         function_name: str,
         *args: Any,
+        estimate_fee: bool = False,
```
MaksymilianDemitraszek, line 45: I think we should expose it with `--estimate-fees` flag

protostar/testing/test_results.py
MaksymilianDemitraszek, line 36: Why don't make it part of `ExecutionResourcesSummary` ?

MaksymilianDemitraszek (changes requested):

protostar/testing/environments/execution_environment.py
```diff
@@ -26,11 +42,39 @@ async def perform_execute(
...
+                async_state=self.state.starknet.state.state, loop=loop
+            )
+            tracker_state = UpdatesTrackerState(state=sync_state)
+            actual_resources = calculate_tx_resources(
+                state=tracker_state,
+                resources_manager=resources_manager,
+                call_infos=[call_info],
+                tx_type=TransactionType.INVOKE_FUNCTION,
+            )
+            estimated_fee = calculate_tx_fee(
```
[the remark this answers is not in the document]
kasperski95: How do you want to test "test case" fee estimation with "feeder_client.simulate_transaction". Estimations are going to vary. What is the acceptable error-margin?

karol-bisztyga:
@kasperski95 should we close this?

PR #979, opened 2022-10-12
Unify max-fee interfaces across commands and cheatcodes

karol-bisztyga:
fix #935

PR #1015, opened 2022-10-18
Fix unify max-fee

karol-bisztyga:
CI Fix for #979


mkaput (approved):

protostar/commands/invoke_command.py
```diff
@@ -124,8 +124,7 @@ async def invoke(
             contract_address=contract_address,
             function_name=function_name,
             inputs=inputs,
-            max_fee=max_fee if isinstance(max_fee, int) else None,
-            auto_estimate_fee=max_fee == "auto",
+            max_fee=max_fee if max_fee is not None else "auto",
```
kasperski95, line 127:
```suggestion
            max_fee=max_fee
```
I think `max_fee` shouldn't be `None` at this point.

kasperski95 (approved):

PR #1019, opened 2022-10-18
Do not lit a fire so much in some cheatcodes docs

mkaput:
I find these warning boxes to be misleading, as there is no danger here. We only want users to pay attention.


MaksymilianDemitraszek (approved):
<3

Issue #962, opened 2022-10-10
Support new `DeployAccount` transaction type (preferably as a CLI command)


Arcticae:
https://starkware.notion.site/starkware/StarkNet-0-10-1-07b5c6d433464c61bf035037b8113b7b

kasperski95:
Closed by https://github.com/software-mansion/protostar/pull/1105

PR #1031, opened 2022-10-19
[0.7] Fix type errors

kasperski95:
Related #962


Arcticae:
This is a duplicate of #966

kasperski95:
@Arcticae It's not a duplicate. The diff was wrong after merging the previous PR. Can you take a look again?

PR #1016, opened 2022-10-18
Basic scripting runtime

mkaput:
cc #997

---

**Stack**:
- #1059
- #1017
- #1016 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #1034, opened 2022-10-19
Prototype initial deploy scripting API

mkaput:
cc #998

---

**Stack**:
- #1035
- #1034 ⬅
- #1017
- #1016

### DIFF
--- a/protostar/scripting/deployments/declare.py
+++ b/protostar/scripting/deployments/declare.py
@@ -1,0 +1,18 @@
+from dataclasses import dataclass
+from typing import Optional
+
+from protostar.scripting.typing import Fee
+
+
+@dataclass
+class DeclaredContract:
+    class_hash: int
+
+
+def declare(
+    contract: str,
+    *,
+    wait_for_acceptance: Optional[bool] = None,
+    max_fee: Optional[Fee] = None,
+) -> DeclaredContract:
+    raise NotImplementedError
--- a/protostar/scripting/deployments/invoke.py
+++ b/protostar/scripting/deployments/invoke.py
@@ -1,0 +1,14 @@
+from typing import Optional
+
+from protostar.scripting.typing import CairoOrPythonData, Fee
+
+
+def invoke(
+    contract_address: int,
+    function_name: str,
+    inputs: Optional[CairoOrPythonData],
+    *,
+    wait_for_acceptance: Optional[bool] = None,
+    max_fee: Optional[Fee] = None,
+) -> None:
+    raise NotImplementedError
--- a/protostar/scripting/typing.py
+++ b/protostar/scripting/typing.py
@@ -1,0 +1,11 @@
+import protostar.starknet.data_transformer as _dt
+import protostar.starknet_gateway.gateway_facade as _gf
+
+PythonData = _dt.PythonData
+CairoOrPythonData = _dt.CairoOrPythonData
+FromPythonTransformer = _dt.FromPythonTransformer
+ToPythonTransformer = _dt.ToPythonTransformer
+
+ContractFunctionInputType = _gf.ContractFunctionInputType
+Wei = _gf.Wei
+Fee = _gf.Fee
--- a/protostar/scripting/__init__.py
+++ b/protostar/scripting/__init__.py
@@ -1,1 +1,2 @@
-from .version import __version__
+from .typing import *
+from .version import *
--- a/protostar/scripting/deployments/__init__.py
+++ b/protostar/scripting/deployments/__init__.py
@@ -1,0 +1,3 @@
+from .call import call
+from .declare import declare, DeclaredContract
+from .invoke import invoke
--- a/protostar/scripting/deployments/call.py
+++ b/protostar/scripting/deployments/call.py
@@ -1,0 +1,11 @@
+from typing import Optional
+
+from protostar.scripting.typing import CairoOrPythonData
+
+
+def call(
+    contract_address: int,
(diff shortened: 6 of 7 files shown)


mkaput:
question: should I include a series of tests showcasing what I see here? These tests would assert that `NotImplemented` is raised

Arcticae (commented):
Overall - LGTM, some thoughts:
As for the tests - i don't think there's a need to have them
As for the implementation stubs - are those the same we already have in migrations? If so, shouldn't we just be using those implementations?

Arcticae (commented):
Overall - LGTM, some thoughts:
As for the tests - i don't think there's a need to have them
As for the implementation stubs - are those the same we already have in migrations? If so, shouldn't we just be using those implementations?

Arcticae (commented):
Overall - LGTM, some thoughts:
As for the tests - i don't think there's a need to have them
As for the implementation stubs - are those the same we already have in migrations? If so, shouldn't we just be using those implementations?

kasperski95:
I am not sure if reusing migration cheatcode signatures is a good idea due. A user needs to import every function.

At the other, we can try mimicking CLI interface so that the user can reuse that knowledge.
If you know how the CLI looks like, you know how to use configuration file (more or less) and you'll know how to use scripts. CLI could be treated as a source of truth. Or `protostar.scripting` should be a source of truth for `protostar.commands`. Maybe `protostar.scripting` should be renamed to `protostar.use_cases`?

> question: should I include a series of tests showcasing what I see here? These tests would assert that NotImplemented is raised

We don't have to be that strict.

mkaput:
> A user needs to import every function.
> 
> At the other, we can try mimicking CLI interface so that the user can reuse that knowledge. If you know how the CLI looks like, you know how to use configuration file (more or less) and you'll know how to use scripts. CLI could be treated as a source of truth. Or `protostar.scripting` should be a source of truth for `protostar.commands`. Maybe `protostar.scripting` should be renamed to `protostar.use_cases`?

@kasperski95 my idea for this API in long term is:

Non-transactional code:
```python
from protostar.scripting.deployments import *

call(something)
invoke(something)
declare(something)
# I don't know how to model deploy yet
```

Transactional code:
```python
from protostar.scripting.deployments import *

with multicall(...) as mc:
    mc.call(something)
    mc.invoke(something)
    mc.declare(something)
    # I don't know how to model deploy yet
```

protostar/scripting/deployments/declare.py
```diff
@@ -0,0 +1,18 @@
...
+
+@dataclass
+class DeclaredContract:
+    class_hash: int
+
+
+def declare(
+    contract: str,
+    *,
+    wait_for_acceptance: Optional[bool] = None,
```
mkaput, line 15: I wonder if we shouldn't make `wait_for_acceptance` be `True` by default, and non-optional, i.e. use `wait_for_acceptance: bool = True` everywhere

protostar/scripting/typing.py
```diff
@@ -0,0 +1,11 @@
...
+import protostar.starknet_gateway.gateway_facade as _gf
+
+PythonData = _dt.PythonData
+CairoOrPythonData = _dt.CairoOrPythonData
+FromPythonTransformer = _dt.FromPythonTransformer
+ToPythonTransformer = _dt.ToPythonTransformer
+
+ContractFunctionInputType = _gf.ContractFunctionInputType
+Wei = _gf.Wei
+Fee = _gf.Fee
```
mkaput, line 11: I wonder if we couldn't re-export these from `protostar.scripting` directly 🤔

MaksymilianDemitraszek:
```
with multicall(...) as mc:
    mc.call(something)
    mc.invoke(something)
    mc.declare(something)
```
I think you can multicall only invokes, but contextmanager API is fine

protostar/scripting/deployments/invoke.py
```diff
@@ -0,0 +1,14 @@
+from typing import Optional
+
+from protostar.scripting.typing import CairoOrPythonData, Fee
+
+
+def invoke(
```
MaksymilianDemitraszek, line 6: I wonder if we shouldn't have separate function for deploy to make it simpler to user, even though it is invoke under the hood

protostar/scripting/deployments/declare.py
kasperski95: Maybe `wait_for_acceptance` should be a generic function.

protostar/scripting/deployments/invoke.py
mkaput: Yes, I agree. But I don't know what signature would make sense yet and is more kind of addition, not necessity initially.

protostar/scripting/deployments/declare.py
mkaput: > Maybe `wait_for_acceptance` should be a generic function.

I don't understand, could you elaborate please?

kasperski95:
```py
declare_tx = declare("main")
awaited_declare_tx = wait_for_acceptance(declare_tx)
assert isinstance(awaited_declare_tx, DeclareTx)
```

mkaput: Isn't this too noisy?

Also this sounds like futures, maybe we should reuse that ones?

mkaput: FYI I can allow top-level awaits in these scripts

mkaput:
new PR: #1059

PR #1051, opened 2022-10-20
Use `encode_shortstring` function from Starknet.py instead of custom code

mkaput:


mkaput:
@Radinyn could you please take a look why tests started to fail here? It's all Git-related stuff 🤨

PR #1053, opened 2022-10-20
Implement `call` command

### LINKED ISSUE #1014: Add `call` command

Arcticae:
closes #1014


protostar/commands/call_command.py
```diff
@@ -0,0 +1,98 @@
...
+            ProtostarArgument(
+                name="function",
+                description="The name of the function being called.",
+                type="str",
+                is_required=True,
+            ),
+            ProtostarArgument(
+                name="inputs",
+                description="Inputs to the function being called, represented by a list of space-delimited values.",
+                type="felt",
```
MaksymilianDemitraszek, line 55:
```suggestion
                type="felt",
```
felt or int?

MaksymilianDemitraszek (approved):
LGTM but @kasperski95 should take a look as well

protostar/commands/call_command.py
Arcticae: inputs are of the same type as in `invoke`. so i guess yes - felt

mkaput (approved):

protostar/commands/call_command.py
```diff
@@ -0,0 +1,98 @@
...
+        return None
+
+    @property
+    def arguments(self):
+        return [
+            *NetworkCommandUtil.network_arguments,
+            ProtostarArgument(
+                name="contract-address",
+                description="The address of the contract being called.",
+                type="str",
```
kasperski95, line 43: The gateway facade expects `int` type from the contract address.

kasperski95 (approved):

protostar/commands/call_command.py
Arcticae: Yes i created a task for that. `GatewayClient` accepts a hex string right now and it will parse it automatically so it will work fine right now.

PR #1056, opened 2022-10-20
Enable git file transport for all unit tests

Arcticae:


mkaput (changes requested):
I don't like the fact that we operate on local user's global Git configs from our tests...

Maybe we should call this command as a step in our CI?

mkaput (approved):

Issue #477, opened 2022-07-15
Simplify `protostar.toml`


kasperski95:
As a StarkNet contract developer, I want `protostar.toml` be more intuitive, so that I don't have to check docs every time I want to modify something.

The current structure is the result of introducing "CLI based configuration" later in the project.

- remove `["protostar.project"]` section
  - `install`, `update`, `remove` commands should accept `libs_path` as an argument
- rename  `["protostar.shared_command_configs"]` to something shorter:
  - `[protostar.shared]`, [`protostar.general`], [`protostar.default`], [`protostar.project`](my fav)
  - this section should be auto generated by the `protostar init`
    - `protostar init` should include a comment explainig this section 
- remove `["protostar.config"]`
  - move `protostar_version` to [`protostar.project`] or shared/general whatever
  - rename `protostar_version` to `min_protostar_version`
  - initially `protostar_version` was meant to be named `version` and indicate version of the `protostar.toml`
- remove double quotes from sections e.g. `["protostar.test"]` -> `[protostar.test]` 
- protostar implementing this task should throw an error if the declared protostar_version is lower
- use kebab case everywhere
- remove `["protostar.contracts"]`?

```toml
# In this section you can set any argument from any command. <= comment generated by protostar init
# e.g.
# ...
[protostar.project] # <= protostar.config + protostar.project + protostar.shared_command_configs
min-protostar-version="9.9.9" # <= extra field not available from the CLI
libs-path="..."
no-color=true
cairo-path=["...", "..."]

# Some comment explaining this section generated by `protostar init`
[protostar.contracts] # <= consider redesigning `build command` so that this section can be removed and to allow building specific contracts rather than the whole project 
# ...

# commands configuration...
[protostar.test]
# ...
```

MaksymilianDemitraszek:
> remove ["protostar.project"] section
>        install, update, remove commands should accept libs_path as an argument

This is a good idea, we may provide it in the default config

>rename ["protostar.shared_command_configs"] to something shorter:
> [protostar.shared], [protostar.general], [protostar.default], [protostar.project](my fav)

`[protostar.project]` is the best imo too 

> rename `protostar_version` to `min_protostar_version`

I dont think it's very nice, it is more like a `source_version` (which is terrible name). `min_protostar_version` does not suggest puropose of this field properly.

Generally very nice proposal!

mkaput:
> > rename protostar_version to min_protostar_version
> 
[...]

PR #1058, opened 2022-10-21
Redesigning `protostar.toml` 13 — Rolling integration — Package management commands

kasperski95:
Add `lib-path` argument to package management commands. This argument is used after migrating to the `ConfigurationFileV2`. Make lib directory optional.

Related #477

### DIFF
--- a/protostar/cli/common_arguments.py
+++ b/protostar/cli/common_arguments.py
@@ -2,6 +2,7 @@ from .protostar_argument import ProtostarArgument
 
 lib_path_arg = ProtostarArgument(
     name="lib-path",
-    description="Directory containing project dependencies",
+    description="Directory containing project dependencies. "
+    "This argument is/will after migrating the configuration file.",
     type="path",
 )
--- a/protostar/cli/common_arguments.py
+++ b/protostar/cli/common_arguments.py
@@ -3,6 +3,6 @@ from .protostar_argument import ProtostarArgument
 LIB_PATH_ARG = ProtostarArgument(
     name="lib-path",
     description="Directory containing project dependencies. "
-    "This argument is/will after migrating the configuration file.",
+    "This argument is used with the configuration file V2.",
     type="path",
 )
--- a/protostar/cli/common_arguments.py
+++ b/protostar/cli/common_arguments.py
@@ -1,6 +1,6 @@
 from .protostar_argument import ProtostarArgument
 
-lib_path_arg = ProtostarArgument(
+LIB_PATH_ARG = ProtostarArgument(
     name="lib-path",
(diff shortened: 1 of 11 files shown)


protostar/cli/common_arguments.py
```diff
@@ -0,0 +1,8 @@
+from .protostar_argument import ProtostarArgument
+
+lib_path_arg = ProtostarArgument(
+    name="lib-path",
+    description="Directory containing project dependencies. "
+    "This argument is/will after migrating the configuration file.",
```
Arcticae, line 6: This seems a bit wrong

protostar/cli/common_arguments.py
```diff
@@ -0,0 +1,8 @@
+from .protostar_argument import ProtostarArgument
+
+lib_path_arg = ProtostarArgument(
```
Arcticae, line 3: Shouldn't this have a default of value "lib"?

Arcticae, line 3: Constants should be all uppercase

tests/integration/protostar_fixture.py
```diff
@@ -1,5 +1,4 @@
 import asyncio
```
Arcticae, line 1: Please revert formatting-only changes

Arcticae (changes requested):
I understand that removing this argument from config file is technically simplifying the `.toml`, but this instead requires passing this on every single invocation of a command. I think this is totally unnecessary and will result in a worse UX.

tests/integration/protostar_fixture.py
kasperski95: [revert formatting](https://github.com/software-mansion/protostar/pull/1058/commits/0fb106927da80f05ce082d01368ed980a4acd741)

protostar/cli/common_arguments.py
kasperski95: [use constant case](https://github.com/software-mansion/protostar/pull/1058/commits/009c1322db2dabc6584dd819891e0ddbe131be08)

kasperski95: We can't use default if the ConfigurationFileV1 is used. Protostar needs to print the warning if the user decides to pass the argument explicitly and the Configuration File isn't migrated. LibPathResolver handles the default value logic for the time being.

protostar/cli/common_arguments.py
kasperski95: [update description](https://github.com/software-mansion/protostar/pull/1058/commits/9e7079d021f1ed070be5939059f2b83f78125d86)

Arcticae (approved):

PR #1059, opened 2022-10-21
Prototype initial deploy scripting API

mkaput:
cc #998

---

**Stack**:
- #1059 ⬅
- #1017
- #1016

### DIFF
--- a/protostar/scripting/__init__.py
+++ b/protostar/scripting/__init__.py
@@ -1,1 +1,2 @@
-from .version import __version__
+from .build import *
+from .version import *
--- a/protostar/scripting/build.py
+++ b/protostar/scripting/build.py
@@ -1,0 +1,28 @@
+from os import PathLike
+from typing import Union
+
+from starknet_py.net.client_models import ContractClass
+from starknet_py.utils.sync.sync import make_sync
+
+__all__ = (
+    "ContractIdentifier",
+    "ContractName",
+    "ContractSourcePath",
+    "build",
+    "build_sync",
+)
+
+ContractName = str
+ContractSourcePath = PathLike
+ContractIdentifier = Union[ContractName, ContractSourcePath]
+
+
+async def build(
+    contract: ContractIdentifier,
+    *,
+    disable_hint_validation: bool = False,
+) -> ContractClass:
+    raise NotImplementedError
+
+
+build_sync = make_sync(build)
--- a/protostar/scripting/version.py
+++ b/protostar/scripting/version.py
@@ -1,3 +1,5 @@
 from protostar.scripting_runtime.api_context import api_context
 
+__all__ = ("__version__",)
+
 __version__: str = api_context().protostar_version


mkaput:
This is reincarnation of #1034. See review comments there.

kasperski95 (commented):
#### running the script
```bash
...
├── protostar.toml
└── scripts
    └── conftest.py
    └── custom_signer.py
    └── deploy_project.py
    └── deploy_project_test.py

protostar deploy-project --skip-e2e
protostar -p mainnet deploy-project
```

#### protostar.toml
```toml
# ...
[deploy-project]
account-address = 123

[profile.mainnet.deploy-project]
account-address = 321
network-name = "mainnet"
```

#### scripts/deploy_project.py
```python
from protostar.scripting import *
from .custom_signer import CustomSigner


@option('--account-address')
@option('--network-name', default="devnet")
@option('--skip-e2e', default=False)
def main(account_address: int, network_name, skip_e2e: bool):
    network = Network.from_known_network(network_name)
    account = Account(
        address=account_address,
        private_key=load_private_key_from_env(),
        signer=CustomSigner(chain_id=network.chain_id),
    )

    project = use_protostar_project()
    project.install_packages()
    project.build()
    if not skip_e2e:
        testing_result = project.run_e2e_tests(target="scripts/deploy_test.py")
        assert testing_result.success
    gateway = project.create_gateway(network)
    deploy_project(gateway, account)


def deploy_project(g: ProtostarStarknetGateway, account: Account):
    # this feels like a Starknet.py territory
    if not g.check_contract_existence(account.address):
        g.send(DeployAccountTransaction(contract="account", salt=123, max_fee="auto"))
    declare_res = g.send(DeclareTransaction(contract="main", account=account, max_fee="auto"))
    declare_res2 = g.send(DeclareTransaction(contract="main2", account=account, max_fee="auto"))
    ud = g.get_universal_deployer()
    invoke_tx1 = ud.create_deploy_contract_invoke_tx(
        class_hash=declare_res.class_hash,
        inputs={"initial_balance": 42},
        account=account, 
        salt=123,
        max_fee="auto",
    )
    res1 = g.send(invoke_tx1)
    invoke_tx2 = ud.create_deploy_contract_invoke_tx(
        class_hash=declare_res2.class_hash,
        inputs={"my_other_contract_address": res1.contract_address},
        account=account, 
        salt=123,
        max_fee=1000000,
    )
    res2 = g.send(invoke_tx2)
    g.send([
        InvokeTransaction(res1.contract_address, "activate", account, max_fee="auto"),
        InvokeTransaction(res2.contract_address, "activate", account, max_fee="auto"),
    ])
    res3 = g.send(Call(res2.contract_address, "get_my_contract_address"))
    assert res3.result == res2.contract_address
```

#### running e2e tests
```sh
protostar test ./scripts
```

#### scripts/deploy_project_test.py
```py
# E2E tests in pytest? Maybe protostar.scripting should be a PyPi package
from protostar.scripting import Network, use_protostar_project
from .conftest import DevnetAccount
from .deploy_project import deploy_project


def test_deploying(devnet: Network, account: DevnetAccount):
    project = use_protostar_project()
    gateway = project.create_gateway(devnet)

    deploy_project(gateway, account)
```

Arcticae (approved):

Arcticae:
I kinda don't understand why we need the __all__ thingy - is that necessary? Do we want to just safeguard ourselves from users importing weird things, when we forget to remove the imports?

mkaput:
> I kinda don't understand why we need the **all** thingy - is that necessary? Do we want to just safeguard ourselves from users importing weird things, when we forget to remove the imports?

You're close. It's also used to define the interface of the library. Importing stuff outside `__all__` will be unsupported. Additionally, `from protostar.scripting import *` will have defined behaviour controlled by us.

mkaput:
I have cut off most of the stuff from this PR because after some thought, it doesn't make any sense to add one more layer of simplification over Starknet.py APIs. If we can simplify stuff or expand it, why not simply add this to Starknet.py and let more people benefit from it?

Because of this, this PR is now limited only to the `build` function, which also has a sync counterpart `build_sync` -- in the similar fashion Starknet.py does. The result was also changed to return `ContractClass` for easier usage with Starknet.py functions.

Some more points, based on the example @kasperski95 did, which has some nice ideas:
1. I don't want to provide a global `main` function like the one Nile expects -- it's not Pythonic. People do `if __name__ == "__main__":` here. If somebody wants to use `async` then let people call `asyncio.run` in that block.
2. I don't understand the reasoning behind `use_protostar_project` function -- we can simply avoid this by exposing global functions.
3. I do very like the idea of the `@option` decorator! I didn't think of something like this. Without the `main` function thingy, I imagine following API for the user:

```python
from protostar.scripting import option

account_address: int = option("account-address")
```
4. I don't want to expose `GatewayFacade`, let it be Protostar internal stuff. After conversation with @MaksymilianDemitraszek we think it's best to simply let users create `GatewayClient` or `FullNodeClient` however they want. Let `scripting` be solely focusing on providing users an ability to run Python code without virtualenv.

Re E2E tests:
1. I think we shouldn't allow writing tests in Python. It will be hard to integrate `pytest` with our test runner. It will be hard to publish `protostar.scripting` to PyPi because it will use some private `multiprocessing.Manager` based communication scheme with Protostar process.
2. My current idea is to provide **one** cheatcode in tests: `run_script(path: PathLike, **kwargs)` which will run the script within a script. For this to work, we will need to provide an ability to define output variables from the scripts, like the ones in Terraform.
3. People may want to run E2E tests on `testnet` or any other custom network. We cannot assume people will use our own provided devnet. Perhaps, we may add a cheatcode to start devnet instance (from `__setup__` hook)?

Arcticae (approved):

kasperski95:
> I don't understand the reasoning behind use_protostar_project function -- we can simply avoid this by exposing global functions.

monorepos?

> I think we shouldn't allow writing tests in Python. It will be hard to integrate pytest with our test runner. It will be hard to publish protostar.scripting to PyPi because it will use some private multiprocessing.Manager based communication scheme with Protostar process.

I think we shouldn't allow writing e2e tests in Cairo. All cheatcodes except fuzzing cheatcodes won't work, and fuzzing cheatcodes will be useless because e2e tests are slower.

Arcticae:
@kasperski95 what exactly would be the problem in monorepo?

mkaput:
> monorepos

Yeah, I thought that users create one Protostar project and write multiple contracts within it if they are all related to each other.

Deployment scripts are going to be strictly bound to a single Protostar project IMHO.

mkaput:
> I think we shouldn't allow writing e2e tests in Cairo. All cheatcodes except fuzzing cheatcodes won't work, and fuzzing cheatcodes will be useless because e2e tests are slower.

From cheatcodes perspective you're right, but the trick is what should happen post Cairo1 world, when Starknet will abandon Python as lingua franca. With future proofing in mind, in my opinion it makes sense to stick with writing tests in Cairo, because everything will uniformly play together within one Cairo-centric ecosystem. Otherwise, users will still have to set up virtual environments and other fuss, only to get E2E tests working.

And cheatcodes availability problem is fixable in other ways. For example, by using the `TestMode` mechanic we already have we can ban using cheatcodes not working in E2E context and vice versa.

mkaput:
Alternative idea:

```python
build(contract_identifier, ...)
```

->

```python
from protostar.scripting import contracts

contracts[contract_identifier].build(...)
```

wdyt?

mkaput:
We are closing this PR. Internally, we have been brainstorming about this feature during last week, and we have decided to take a different path here because we found many flaws in this idea.

PR #1064, opened 2022-10-21
Add "address" argument type

### LINKED ISSUE #1054: Create a new type to accomodate various formats of addresses
I want to have a drop-in type of argument type which accepts both hex and int as an input, and coerces it to int automatically. 
This would be usable in addresses (invoke/call).

Arcticae:
closes #1054


protostar/argument_parser/arg_type_test.py
```diff
@@ -38,3 +39,10 @@ def test_regex_arg_type():
     result = map_type_name_to_parser("regexp")(".*")
 
     assert isinstance(result, Pattern)
+
+
+def test_address_arg_type():
+    parser = map_protostar_type_name_to_parser("address")
+    result = parser("0xdeadbeef123")
+    assert isinstance(result, int)
+    assert result == 15302363377955
```
kasperski95, line 48: This test should be in the `cli` module in the `protostar_arg_type_test.py` file. `protostar.cli` depends on `protostar.argument_parser`.

protostar/cli/protostar_arg_type.py
```diff
@@ -33,3 +37,9 @@ def parse_fee_arg_type(arg: str) -> Fee:
     if arg == "auto":
         return arg
     return int(arg)
+
+
+def parse_address_arg_type(arg: str) -> int:
+    if arg.startswith("0x"):
+        return int(arg, 16)
+    return int(arg)
```
kasperski95, line 45: What is the address convention? Maybe Protostar should tell the user to provide addresses as a hexadecimal number, and other formats won't be supported in the future.

protostar/commands/invoke_command.py
```diff
@@ -44,7 +44,7 @@ def arguments(self):
             ProtostarArgument(
```
kasperski95, line 44: I think "account-address" arg in the `SignableCommandUtil` should also be updated.

protostar/cli/protostar_arg_type.py
Arcticae: Both formats should be accepted, since some tools export the address as an integer

protostar/commands/invoke_command.py
Arcticae: done

mkaput (approved):

protostar/cli/protostar_arg_type.py
kasperski95: Today, I asked the Starknet.py team how should address be represented in the console. They recommended using hex representation. I think we should at least give users an initiative to represent addresses consistently, e.g. by printing a warning. 

https://swmansion.slack.com/archives/C02MBPPR2HY/p1667474916413619

kasperski95 (approved):
LGTM

protostar/cli/protostar_arg_type.py
Arcticae: We talked about the console output, not the inputs passed to the program

Arcticae: I think that's a bit overzealous

PR #1080, opened 2022-10-26
Remove rollback from migrations

kasperski95:
Remove rollback functionality from migrations. This functionality creates more confusion than good.

Related #1001

### DIFF
--- a/protostar/commands/migrate_command.py
+++ b/protostar/commands/migrate_command.py
@@ -59,11 +59,6 @@ class MigrateCommand(ProtostarCommand):
                 is_positional=True,
             ),
             ProtostarArgument(
-                name="rollback",
-                description="Run `rollback` function in the migration script.",
-                type="bool",
-            ),
-            ProtostarArgument(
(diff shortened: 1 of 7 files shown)


Arcticae (approved):

mkaput (changes requested):
Blocking this PR because it removes a feature without deprecation notice.

Also, I think we can simply not touch migrations code and wait for deployment scripts to be finalized. Then we can remove all migrations code together in the future.

Wdyt?

kasperski95:
> Blocking this PR because it removes a feature without deprecation notice.

Screenshot 2022-11-02 at 10 45 34

> Also, I think we can simply not touch migrations code and wait for deployment scripts to be finalized. Then we can remove all migrations code together in the future.

There's no point of adding new features to the migrations. However, I wouldn't avoid touching migrations if the change is fast to make, and adds a value to the project even in the shorter perspective.

mkaput (approved):
You convinced me @kasperski95 😉

PR #1081, opened 2022-10-26
Redesigning `protostar.toml` 18 — Connect compatibility checker

kasperski95:
Display warnings if a Protostar project is not compatible with Protostar.
Related #477

### DIFF
--- a/protostar/composition_root.py
+++ b/protostar/composition_root.py
@@ -40,6 +40,7 @@ from protostar.configuration_file import (
 from protostar.io import InputRequester, log_color_provider
 from protostar.migrator import Migrator, MigratorExecutionEnvironment
 from protostar.protostar_cli import ProtostarCLI
+from protostar.self import ProtostarCompatibilityWithProjectChecker
 from protostar.self.protostar_directory import ProtostarDirectory, VersionManager
 from protostar.starknet_gateway import GatewayFacadeFactory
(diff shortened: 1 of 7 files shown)


Arcticae (changes requested):
I don't understand why we validate this. This only makes the user confused when he upgrades protostar, and changes in a new version might not be even breaking the behavior of config file.

kasperski95:
> I don't understand why we validate this. This only makes the user confused when he upgrades protostar, and changes in a new version might not be even breaking the behavior of config file.

The behavior of the configuration file doesn't have to change, but the behavior of cheatcodes can. The change doesn't have to be breaking to introduce unexpected behavior.

Arcticae:
I'm not convinced. Going to leave it up to @mkaput

mkaput (approved):
@kasperski95 intentions are convincing to me, but I definitely would love to hear community feedback on this. Indeed, for some people this will be only unwanted burden at current stage of Protostar when users probably catch up new releases quickly.

kasperski95:
Currently, the checker will only print a warning. It shouldn't be intrusive, yet it gives an initiative to keep the declaration valid. Without the checker, there's no point of keeping `protostar-version` in the configuration file because it can be invalid.

mkaput:
Yup, I agree with your point and this code is fine to me!

PR #1082, opened 2022-10-27
Suggest chain-ids values

### LINKED ISSUE #1074: Make `chain-id` easier to use (with devnet)
Currently, `chain-id` argument is required if the user interacts with devnet, and Protostar rejects chain ids different from the ones used by mainnet and testnet.

kasperski95:
Show supported chain ID values if invalid is provided. Supported values are hard to memorize. A user needs to provide chain ID explicitly when using Protostar with devnet.

Closes #1074

### DIFF
--- a/protostar/cli/network_command_util.py
+++ b/protostar/cli/network_command_util.py
@@ -71,7 +71,7 @@ class NetworkCommandUtil:
         return NetworkConfig.build(
             gateway_url=self._args.gateway_url,
             network=self._args.network,
-            chain_id=self.normalize_chain_id(self._args.chain_id),
+            chain_id=self._normalize_chain_id(self._args.chain_id),
         )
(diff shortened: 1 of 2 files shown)


mkaput (approved):
In general this code looks OK, but I'm not sure this solves users' problem:

1. People expect an ability to pass hex values there from what I remember
2. Shouldn't we also allow users to pass human understandable names?

kasperski95:
> People expect an ability to pass hex values there from what I remember

Starknet.py doesn't support it. We would have to change the Starknet.py's `StarkCurveSigner` to accept integer chain ids, and I'd wanted to bother with this in this PR. It can be improved in the future.

> Shouldn't we also allow users to pass human understandable names?

We could, but maybe in the future PR.

PR #1089, opened 2022-11-01
outdated output-dir option removed from migration examples

benceharomi:
To meet the changes introduced in `v0.6.0` the examples for the migrate command has been updated:

- the `output-dir` flag is changed to `compiled-contracts-dir`


Arcticae (changes requested):
This is not how it works, although the docs are outdated right now. `output-dir` has been removed overall, and `compiled-contracts-dir` is an unrelated option. It serves the purpose of passing the directory we host the compiled contracts, so migrations can use them for declare & deploy. LMK if i should update, or if that's something you'd like to tackle :)

benceharomi:
> 
Thank you for the clarification, I updated the changes accordingly.

Arcticae:
Thank you for your contribution @benceharomi !

PR #1106, opened 2022-11-06
Type generator from for CLI commands

kasperski95:
Add type generator for CLI commands. Use it in the `InitCommand`.

This PR was done partially as a side project to learn compiler related concepts.


mkaput (changes requested):
Hmm, this is definitely interesting approach and I guess it was pretty nice side project 😄 

But from project's point of view, I don't think we should merge this. To me this seems like additional maintenance burden which to my intuition will unfortunately only be slowing our pace of development instead of bringing expected improvements. It's nice that this gives us type safety and autocompletion, but I can imagine we would have to constantly modify these scripts to accommodate new features we introduce in our CLI.

I think the correct path would be to go the other way around! These `*Input` dataclasses are super nice! Why not generate CLI from these? Hint hint: that's how https://typer.tiangolo.com/ works.

Blocking this PR, awaiting others' opinions.

kasperski95:
> Why not generate CLI from these?

It would be harder to integrate with the existing code.

> But from project's point of view, I don't think we should merge this. To me this seems like additional maintenance burden which to my intuition will unfortunately only be slowing our pace of development instead of bringing expected improvements. It's nice that this gives us type safety and autocompletion, but I can imagine we would have to constantly modify these scripts to accommodate new features we introduce in our CLI.

The problem is, we don't have tests for every single argument. This is a place in the application we could rely on types more heavily instead. Translating args to typed args manually is hard to maintain, because we have two sources of information that need to be kept in sync.

> BTW I would happily merge these small tweaks you did to InitCommand code.

Tweaks depend on the generated `*Input` type.

mkaput:
> It would be harder to integrate with the existing code.

So perhaps, something is wrong with existing code?

> The problem is, we don't have tests for every single argument. This is a place in the application we could rely on types more heavily instead. Translating args to typed args manually is hard to maintain, because we have two sources of information that need to be kept in sync.

We do have E2E tests though, which I believe cover a lot of stuff? At least an amount that we are pretty confident into making changes there.

kasperski95:
Ok, I'll close it. The alternative approach would require more work, hence it's not going to happen soon.

> So perhaps, something is wrong with existing code?

The current approach has certain flaws, but I wouldn't say it's wrong. The alternative approach would change the design that would affect all commands. This PR introduces a change that is opt-in.

> We do have E2E tests though, which I believe cover a lot of stuff? At least an amount that we are pretty confident into making changes there.

Tests don't cover everything.

PR #1122, opened 2022-11-07
Block explorer links

### LINKED ISSUE #899: Add StarkScan as block explorer when deploying
As a StarkNet smart contract developer, I want to have a link to the StarkScan ([Twitter](https://twitter.com/StarkscanCo/status/1575170281416454154)) block explorer ([testnet](https://testnet.starkscan.co/), [mainnet](https://starkscan.co/)) when deploying a contract so that I can choose to use either voyager or starkscan.

The photo below shows how the voyager link is given when deploying a contract. Would be nice to have a link to starkscan too. 

image

kasperski95:
Introduce Block Explorer classes. Integrate with existing commands.
Closes #899


protostar/cli/protostar_arg_type.py
```diff
@@ -52,3 +55,8 @@ def parse_hex_or_decimal(arg: str) -> int:
     if arg.startswith("0x"):
         return int(arg, 16)
     return int(arg)
+
+
+def parse_block_explorer_type(arg: str) -> SupportedBlockExplorerName:
+    assert arg in SUPPORTED_BLOCK_EXPLORER_NAMES
```
mkaput, line 61: could we raise ValueError here?

protostar/cli/protostar_arg_type_test.py
```diff
@@ -39,3 +39,15 @@ def test_address_arg_type():
...
+
+def test_block_explorer_type():
+    parser = map_protostar_type_name_to_parser("block_explorer")
+    result = parser("starkscan")
+    assert result == "starkscan"
+
+
+def test_fail_on_wrong_block_explorer():
+    parser = map_protostar_type_name_to_parser("block_explorer")
+    with pytest.raises(AssertionError):
```
mkaput, line 52: pls match error message here

protostar/starknet_gateway/block_explorer/fake_block_explorer.py
```diff
@@ -0,0 +1,18 @@
...
+
+class FakeBlockExplorer(BlockExplorer):
+    def create_link_to_transaction(self, tx_hash: TransactionHash) -> URL:
+        return ""
+
+    def create_link_to_contract(self, contract_address: ContractAddress) -> URL:
+        return ""
+
+    def create_link_to_class(self, class_hash: ClassHash) -> URL:
+        return ""
```
mkaput, line 18: I think we can avoid this class?

Why not assume voyager as default block explorer even in tests? Simply avoid testing block explorer output in assertions.

protostar/starknet_gateway/block_explorer/block_explorer_factory.py
```diff
@@ -0,0 +1,29 @@
...
+SUPPORTED_BLOCK_EXPLORER_NAMES: list[SupportedBlockExplorerName] = [
+    "starkscan",
+    "viewblock",
+    "voyager",
+]
+
+
+def create_block_explorer(
+    block_explorer_name: SupportedBlockExplorerName,
+    network: Optional[PredefinedNetwork],
```
mkaput, line 20: I think `network` could be just an argument to all `create_*` functions and thus you could make `BlockExplorer` instances parameterless.

protostar/starknet_gateway/block_explorer/stark_scan_block_explorer_test.py
```diff
@@ -0,0 +1,61 @@
+import pytest
```
mkaput, line 1: I think this test does not bring any value. Tested functions are trivial by definition? Are we sure we want this?

protostar/starknet_gateway/block_explorer/stark_scan_block_explorer.py
```diff
@@ -0,0 +1,28 @@
...
+    TransactionHash,
+)
+
+
+class StarkScanBlockExplorer(BlockExplorer):
+    def __init__(self, network: PredefinedNetwork) -> None:
+        super().__init__()
+        network_to_domain: dict[PredefinedNetwork, URL] = {
+            "mainnet": "https://starkscan.co",
+            "testnet": "https://testnet.starkscan.co",
```
mkaput, line 17: What about testnet2?

mkaput (changes requested):

protostar/starknet_gateway/block_explorer/fake_block_explorer.py
mkaput: discussed this privately

website/docs/cli-reference.md
```diff
@@ -53,6 +53,11 @@ Required.
 Path to compiled contract.
 #### `--account-address ADDRESS`
 Account address.
+#### `--block-explorer BLOCK_EXPLORER=starkscan`
+Generated links will point to that block explorer. Available values:
+- starkscan
+- viewblock
+- voyager
```
Arcticae, line 60: I think users should be somehow aware that we can add one on their request (after verification, if the project is mature enough). We should either include it here or somewhere in the docs

Arcticae (approved):
Looks good codewise. I am wondering if we should either hurry with this or wait for @mkaput to finish his formatter output thing, but either way if that isn't conflicting - i think it's mergable as-is.

mkaput:
I would prefer to have this merged before I get with formatter stuff onto it

protostar/starknet_gateway/block_explorer/stark_scan_block_explorer.py
kasperski95: It's out of scope of this PR. Starknet.py needs to introduce it.

website/docs/cli-reference.md
kasperski95: This issue was requested by users. I am not sure if it's needed.

protostar/starknet_gateway/block_explorer/stark_scan_block_explorer.py
mkaput: Starknet.py team says they will rather not introduce anything to PredefinedNetwork. But I think this doesn't mean we should also not?

protostar/starknet_gateway/block_explorer/block_explorer_factory.py
kasperski95: I'd have to then pass network to formatters.

protostar/cli/protostar_arg_type_test.py
kasperski95: Changed to `ValueError`. argparse generates an error message for ValueErrors.

```
... argument --block-explorer: invalid parse_block_explorer_type value: 'abc'
```

protostar/starknet_gateway/block_explorer/stark_scan_block_explorer_test.py
kasperski95: Removed tests

protostar/starknet_gateway/block_explorer/block_explorer_factory.py
mkaput: OK, makes sense

protostar/starknet_gateway/block_explorer/stark_scan_block_explorer.py
kasperski95: Then we need to decouple this concept from Starknet.py. That's a different issue.

PR #1131, opened 2022-11-09
Set up output formatting infrastructure

mkaput:
commit-id:92df4f6a

---

**Stack**:
- #1146
- #1132
- #1137
- #1131 ⬅

### DIFF
--- a/protostar/io/output.py
+++ b/protostar/io/output.py
@@ -1,0 +1,45 @@
+import json
+import sys
+from abc import ABC, abstractmethod
+
+from .log_color_provider import LogColorProvider
+
+
+class Message(ABC):
+    @abstractmethod
+    def format_human(self, fmt: LogColorProvider) -> str:
+        ...
+
+
+class StructuredMessage(Message, ABC):
+    @abstractmethod
+    def format_dict(self) -> dict:
+        ...
+
+
+class Messenger(ABC):
+    @abstractmethod
+    def __call__(self, message: Message):
+        ...
+
+
+class HumanMessenger(Messenger):
+    def __init__(self, log_color_provider: LogColorProvider):
+        self._log_color_provider = log_color_provider
+
+    def __call__(self, message: Message):
+        human_message = message.format_human(fmt=self._log_color_provider)
+        print(human_message)
+
+
+class JsonMessenger(Messenger):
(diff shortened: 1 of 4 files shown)


protostar/cli/output_command_util.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any
+
+from .protostar_argument import ProtostarArgument
+from ..io import Messenger, JsonMessenger, HumanMessenger, LogColorProvider
+
+
+class OutputCommandUtil:
+    output_arguments = [
```
Arcticae, line 8: [nitpick]: this should be uppercase since it's a constant

Arcticae (approved):

protostar/cli/output_command_util.py
mkaput: done

Arcticae (approved):

kasperski95 (approved):
LGTM

protostar/io/output.py
```diff
@@ -0,0 +1,45 @@
...
+        ...
+
+
+class Messenger(ABC):
+    @abstractmethod
+    def __call__(self, message: Message):
+        ...
+
+
+class HumanMessenger(Messenger):
```
MaksymilianDemitraszek, line 26:
```suggestion
class ConsoleMessenger(Messenger):
```
Or `LoggerMessanger` maybe?

MaksymilianDemitraszek: > What's wrong with Human?

Sounds like we would have been sending messages from Mars (I think @Radinyn is a fan of such things) 😄 .

But seriously, we do not use human in other places of the project, we just call it `format` for example in `protostar/commands/test/test_result_formatter.py` so it is kinda inconsistent.  It should be marked Nitpick though.

MaksymilianDemitraszek: On the other hand we had no different types of `format` before. I think maybe human is alright but we have to be consistent with it.

mkaput: `protostar/commands/test/test_result_formatter.py` is up to be redone in scope of this project so it will naturally use `human`

mkaput:
Replaced `OutputCommandUtil` with `MessengerFactory`

PR #1132, opened 2022-11-09
Implement structured output for `format` command

mkaput:
commit-id:1d51ef78

---

**Stack**:
- #1154
- #1146
- #1132 ⬅

### DIFF
--- a/protostar/formatter/formatting_result.py
+++ b/protostar/formatter/formatting_result.py
@@ -1,18 +1,26 @@
-from typing import List
+from abc import ABC
 from dataclasses import dataclass
 from pathlib import Path
 
-from protostar.io.log_color_provider import log_color_provider
+from protostar.io import Message, LogColorProvider
 
 
 @dataclass
-class FormattingResult:
+class FormattingResult(Message, ABC):
     filepath: Path
+    checked_only: bool
+
+    def _format_path(self, fmt: LogColorProvider):
+        return fmt.colorize("GRAY", str(self.filepath))
 
 
 @dataclass
 class BrokenFormattingResult(FormattingResult):
     exception: Exception
+
+    def format_human(self, fmt: LogColorProvider) -> str:
+        header = _pad_header(_get_broken_header(fmt), self.checked_only, fmt)
+        return f"{header} {self._format_path(fmt)}\n{str(self.exception)}"
 
 
 @dataclass
@@ -21,6 +29,12 @@ class CorrectFormattingResult(FormattingResult):
     Indicates that file was already properly formatted when encountered.
     """
 
+    def format_human(self, fmt: LogColorProvider) -> str:
+        header = _pad_header(
+            _get_correct_header(self.checked_only, fmt), self.checked_only, fmt
+        )
+        return f"{header} {self._format_path(fmt)}"
+
 
 @dataclass
 class IncorrectFormattingResult(FormattingResult):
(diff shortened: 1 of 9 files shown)


protostar/formatter/formatting_result.py
```diff
@@ -1,82 +1,96 @@
...
 
-from protostar.io.log_color_provider import log_color_provider
+from protostar.io import Message, LogColorProvider
 
 
 @dataclass
-class FormattingResult:
+class FormattingResult(Message, ABC):
     filepath: Path
+    check: bool
```
kasperski95, line 11: Why is `check` needed here?

mkaput: 1. I need it to determine what header to use for correct & incorrect results (mind that these data structures are now self-containing, they have all information they need for presentation and they do not have to reference "parents")
2. I would love to put `check` only in correct & incorrect and skip it for `broken`, but padding of headers essentially forces me to make `check` more general because it is needed in each variant.

kasperski95: `Check` is an input parameter. The name `check` doesn't make sense within result's context.

protostar/formatter/formatting_summary.py
```diff
@@ -23,30 +26,44 @@ def extend(self, formatting_result: FormattingResult):
...
+        return {
+            "type": "format",
+            "check": self.check,
+            "summary": {
+                "total": len(self),
+                "unchanged": len(self.correct),
+                "formatted": len(self.incorrect),
+                "broken": len(self.broken),
+            },
+        }
```
kasperski95, line 69: Why is `"type": "format"` needed? I don't see why Protostar duplicates input data.

protostar/formatter/formatting_result.py
mkaput: It actually does! It's good to know from the result whether the operation modified stuff or not IMHO

protostar/formatter/formatting_summary.py
mkaput: https://github.com/software-mansion/protostar/discussions/1123

Radinyn (approved):

protostar/formatter/formatting_summary.py
```diff
@@ -23,30 +26,32 @@ def extend(self, formatting_result: FormattingResult):
...
+    def __len__(self) -> int:
         return len(self.broken) + len(self.correct) + len(self.incorrect)
 
     def any_unformatted_or_broken(self, check: bool):
         return bool(len(self.broken) + (len(self.incorrect) if check else 0))
 
+    def format_human(self, fmt: LogColorProvider) -> str:
+        total = len(self)
+        if total == 0:
+            return "No files found"
```
Arcticae, line 38: I think this case is untested

Arcticae (approved):
LGTM

protostar/formatter/formatting_summary.py
mkaput: done

protostar/formatter/formatting_result.py
```diff
@@ -1,82 +1,72 @@
...
+        return fmt.colorize("GRAY", str(self.filepath))
 
 
 @dataclass
 class BrokenFormattingResult(FormattingResult):
     exception: Exception
 
+    def format_human(self, fmt: LogColorProvider) -> str:
+        header = _pad_header(_get_broken_header(fmt), self.check, fmt)
+        return f"{header} {self._format_path(fmt)}\n{str(self.exception)}"
```
kasperski95, line 23: I'd probably create separate classes in `protostar.commands`. It's a tradeoff between [low coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)) and boilerplate. In layered architectures, application code should be decoupled from I/O.

#### protostar.commands
```py
class BrokenFormattingResultMessage(Message[BrokenFormattingResult]):
    def format_human(self, obj: BrokenFormattingResult) -> str:
        return "..."

    def format_object(self, obj: BrokenFormattingResult) -> dict
        return {}
```

kasperski95 (commented):
I am not a fan of this approach because it reintroduces a pattern that required refactoring in the past.

protostar/formatter/formatting_result.py
mkaput: discussed this privately

protostar/formatter/formatting_result.py
mkaput: done

PR #1133, opened 2022-11-09
Added `max-steps` flag

Radinyn:


protostar/commands/test/test_command.py
```diff
@@ -137,6 +137,11 @@ def arguments(self):
                 type="int",
                 description="Set a seed to use for all fuzz tests.",
             ),
+            ProtostarArgument(
+                name="max-steps",
+                type="int",
+                description="Set the Cairo step limit.",
```
mkaput, line 143:
```suggestion
                description="Set Cairo execution steps limit.",
```

tests/e2e/test_testing.py
```diff
@@ -274,3 +274,18 @@ def test_skipping(protostar: ProtostarFixture, copy_fixture: CopyFixture):
     result = protostar(["test", "tests"])
     assert "SKIP" in result
     assert "REASON" in result
+
+
+@pytest.mark.usefixtures("init")
+def test_max_steps(protostar: ProtostarFixture, copy_fixture: CopyFixture):
```
mkaput, line 280: Can we use integration tests for this instead? E2E tests are super slow..

protostar/starknet/cheatable_execute_entry_point.py
```diff
@@ -89,6 +91,9 @@ class CheatableExecuteEntryPoint(ExecuteEntryPoint):
     contract_callstack: list[str] = []
     profiling = False
 
+    # None results in default Cairo value
+    max_steps: Optional[int] = None
```
mkaput, line 95:
```suggestion
    max_steps: Optional[int] = None
    "``None`` means default Cairo value."
```

protostar/starknet/cheatable_execute_entry_point.py
```diff
@@ -300,6 +305,26 @@ def append_runtime_profile(
...
         )
 
+    async def execute_for_testing(
+        self,
+        state: State,
+        general_config: StarknetGeneralConfig,
+        resources_manager: Optional[ExecutionResourcesManager] = None,
+        tx_execution_context: Optional[TransactionExecutionContext] = None,
+    ) -> CallInfo:
+
```
mkaput, line 315:
```suggestion
```

protostar/starknet/cheatable_execute_entry_point.py
```diff
@@ -300,6 +305,26 @@ def append_runtime_profile(
...
+        if self.max_steps is not None:
+
+            # Providing a negative value to Protostar results in infinite steps,
+            # this is here to mimic default Cairo behavior
+            value = None if self.max_steps < 0 else self.max_steps
+            new_config.__dict__["invoke_tx_max_n_steps"] = value
+
+        return await super().execute_for_testing(
+            state, new_config, resources_manager, tx_execution_context
+        )
```
mkaput, line 326:
```suggestion
        # Providing a negative value to Protostar results in infinite steps,
        # this is here to mimic default Cairo behavior.
        if self.max_steps is not None and self.max_steps >= 0:
            general_config = dataclasses.replace(
                general_config,
                invoke_tx_max_n_steps=value
            )

        return await super().execute_for_testing(
            state, general_config, resources_manager, tx_execution_context
        )
```

I think this should work?

mkaput (approved):
Looks good to me, no critical comments

protostar/starknet/cheatable_execute_entry_point.py
Radinyn: It will result in a default value (aka not infinity) for negative numbers

mkaput: You can tweak the condition here, but mostly I wanted to express usage of `dataclasses.replace` here

Radinyn: `dataclasses.replace` seems to be blocked by the typeguard, it allows setting this value to int only, and sometimes None is needed

tests/e2e/test_testing.py
Radinyn: Done

protostar/starknet/cheatable_execute_entry_point.py
mkaput: `# type: ignore`?

tests/integration/testing_steps/testing_steps_test.py
```diff
@@ -0,0 +1,43 @@
...
+):
+    testing_summary = await run_cairo_test_runner(
+        Path(__file__).parent / "testing_steps_test.cairo", max_steps=10
+    )
+
+    assert_cairo_test_cases(
+        testing_summary, expected_failed_test_cases_names=["test_max_steps"]
+    )
+
+    assert "OUT_OF_RESOURCES" in format_test_result(testing_summary.failed[0])
```
mkaput, line 23: Are you able to avoid using `format_test_result` here? It would be much better to extract the failure exception and match on its message. The test would be less dependent on unrelated stuff.

mkaput (changes requested):

protostar/starknet/cheatable_execute_entry_point.py
Radinyn: `typeguard` raises `TypeError` either way

tests/integration/testing_steps/testing_steps_test.py
Radinyn: done

protostar/starknet/cheatable_execute_entry_point.py
mkaput: OK, please leave a `# NOTE:` why you did this code that way though.

mkaput (approved):

PR #1141, opened 2022-11-09
Use pytest-xdist to further parallelize tests

mkaput:
https://pytest-xdist.readthedocs.io/en/latest/distribution.html

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -96,6 +96,8 @@ jobs:
       - name: Integration tests
         run: |
           poetry run poe test_integration
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: 4
       - name: Convert coverage to XML
         run: |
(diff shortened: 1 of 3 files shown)


Arcticae:
@mkaput Does this work properly with coverage?

mkaput:
> @mkaput Does this work properly with coverage?

Probably not, but I primarily want to experiment if this helps at all, because default `auto` mode does not gain anything...

kasperski95 (commented):
If codecov is a problem, I'd remove it in favor of faster checks.

mkaput:
Now coverage is disabled @Arcticae @kasperski95

Arcticae:
@mkaput is this one ready? does it introduce a speedup?

mkaput:
Yep, not as much as I anticipated, but still few minutes better.

Arcticae (approved):

PR #1146, opened 2022-11-10
Support indicators in structured output and implement one for `build`…

mkaput:
… command

commit-id:d9ef14ef

---

**Stack**:
- #1160
- #1166
- #1165
- #1154
- #1146 ⬅

### DIFF
--- a/protostar/commands/build_command.py
+++ b/protostar/commands/build_command.py
@@ -1,11 +1,18 @@
-from argparse import Namespace
-from logging import Logger
 from pathlib import Path
-from typing import List, Optional
+from typing import List, Optional, Any
 
-from protostar.cli import ActivityIndicator, ProtostarArgument, ProtostarCommand
+from protostar.cli import (
+    ProtostarArgument,
+    ProtostarCommand,
+    MessengerFactory,
+)
 from protostar.compiler import ProjectCompiler, ProjectCompilerConfig
-from protostar.io import log_color_provider
+from protostar.io import LogColorProvider, Message
+
+
+class BuildActivityMessageTemplate(Message):
+    def format_human(self, fmt: LogColorProvider) -> str:
+        return "Building projects' contracts"
 
 
 class BuildCommand(ProtostarCommand):
@@ -16,10 +23,14 @@ class BuildCommand(ProtostarCommand):
         default="build",
     )
 
-    def __init__(self, project_compiler: ProjectCompiler, logger: Logger) -> None:
+    def __init__(
+        self,
+        project_compiler: ProjectCompiler,
+        messenger_factory: MessengerFactory,
+    ):
         super().__init__()
         self._project_compiler = project_compiler
-        self._logger = logger
+        self._messenger_factory = messenger_factory
 
     @property
     def example(self) -> Optional[str]:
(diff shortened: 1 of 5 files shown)


protostar/commands/build_command.py
```diff
@@ -50,31 +65,26 @@ def arguments(self):
...
         ]
 
-    async def run(self, args: Namespace):
-        await self.build(
-            output_dir=args.compiled_contracts_dir,
-            disable_hint_validation=args.disable_hint_validation,
-            relative_cairo_path=args.cairo_path,
-        )
+    async def run(self, args: Any):
+        write = self._messenger_factory.from_args(args)
```
kasperski95, line 69:
```suggestion
        messanger = self._messenger_factory.from_args(args)
```
It's strange `messanger_factory` produces `write`.

protostar/io/output.py
```diff
@@ -39,7 +62,28 @@ def __call__(self, message: Message):
...
+
+        rendered = message_template.format_dict()
+
+        self._print({**rendered, "event": "started"})
+        try:
+            yield
+            self._print({**rendered, "event": "ok"})
+        except BaseException:
+            self._print({**rendered, "event": "failed"})
+            raise
```
kasperski95, line 83: (We need to figure out how to display errors (ProtostarException), but this is out of scope of this PR.)

kasperski95 (approved):
LGTM

protostar/commands/build_command.py
mkaput: Yeah, but it's also strange to have `messenger(...)` function (noun vs verb)

protostar/io/output.py
mkaput: #1155

protostar/commands/build_command.py
kasperski95: `write_fn_factory`? I don't know. Let's leave it as it is.

protostar/io/output.py
```diff
@@ -39,7 +62,28 @@ def __call__(self, message: Message):
                 f"JSON output is not supported for messages of type {type(message).__name__}."
             )
 
-        dictionary = message.format_dict()
-        json.dump(dictionary, sys.stdout, allow_nan=False, separators=(",", ":"))
+        obj = message.format_dict()
+        self._print(obj)
+
+    @contextmanager
+    def activity(self, message_template: Message) -> Generator[None, None, None]:
```
Arcticae, line 69: I'm wondering if that's something that users would like to see in the output... I think the activity indicator thing is useful for having some UX thing making people aware that something is happening, whilst when having the json output i'd rather have the output in one plain json.

mkaput: I think I have a good reason for having these messages: measuring timing of the process in the tooling. With these messages, one is able to determine time cost of compilation minus Protostar booting time.

But nonetheless: I just copied what `cargo test` does in its JSON output. Maybe it doesn't really make sense here.

Another point of this PR was to move activity indicators to structured output framework. Maybe this command could simply skip JSON output as for now, like we did for `protostar format`?

Arcticae: I think we can skip, since the output of the build is not that important - we know the preliminary contract names (and thus output names)

mkaput: done

Arcticae (approved):

PR #1169, opened 2022-11-16
Use global `logging` instead of passing `Logger` as dependency

mkaput:
commit-id:a62e6eee

### DIFF
--- a/protostar/starknet_gateway/gateway_facade.py
+++ b/protostar/starknet_gateway/gateway_facade.py
@@ -1,5 +1,5 @@
 import dataclasses
-from logging import Logger, getLogger
+from logging import getLogger, LoggerAdapter
 from pathlib import Path
 from typing import (
     Any,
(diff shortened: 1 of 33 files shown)


protostar/starknet_gateway/gateway_facade.py
```diff
@@ -292,8 +289,7 @@ async def declare_v0(
             result = await self._gateway_client.declare(tx, token)
             register_response(dataclasses.asdict(result))
             if wait_for_acceptance:
-                if self._logger:
-                    self._logger.info("Waiting for acceptance...")
+                logging.info("Waiting for acceptance...")
```
Arcticae, line 292: We want to stop logging sometimes, right? That's why we didn't pass the logger higher up in some cases

Arcticae (approved):
Generally approve, but the conditional logging worries me a bit, we'll start logging different stuff in migrations when we merge this.

protostar/starknet_gateway/gateway_facade.py
mkaput: Was this an expected behaviour? From the code it looked to me rather like someone being lazy when implementing...

Arcticae (approved):

PR #1150, opened 2022-11-14
Add `calculate-account-address` command

kasperski95:
Add `calculate-account-address` command. Introduce and use `Address` [Value Object](https://en.wikipedia.org/wiki/Value_object). 

Closes https://github.com/software-mansion/protostar/issues/1107

PR #1170, opened 2022-11-16
Integrate `Address` value object

kasperski95:
Use `Address` and `AccountAddress` value objects consistently across the code base. Validate addresses. Display addresses consistently.

Related https://github.com/software-mansion/protostar/pull/1150


protostar/starknet_gateway/block_explorer/stark_scan_block_explorer.py
```diff
@@ -20,10 +20,8 @@ def get_name(self) -> str:
     def create_link_to_transaction(self, tx_hash: TransactionHash) -> Optional[str]:
         return f"{self._domain}/tx/0x{tx_hash:064x}"
 
-    def create_link_to_contract(
-        self, contract_address: ContractAddress
-    ) -> Optional[str]:
-        return f"{self._domain}/contract/0x{contract_address:064x}"
+    def create_link_to_contract(self, contract_address: Address) -> Optional[str]:
+        return f"{self._domain}/contract/0x{int(contract_address):064x}"
```
Arcticae, line 24: we could just do regular interpolation here

protostar/starknet_gateway/block_explorer/viewblock_block_explorer.py
```diff
@@ -17,10 +17,8 @@ def get_name(self) -> str:
     def create_link_to_transaction(self, tx_hash: TransactionHash) -> Optional[str]:
         return f"{self._prefix}/tx/0x{tx_hash:064x}{self._get_network_query_param()}"
 
-    def create_link_to_contract(
-        self, contract_address: ContractAddress
-    ) -> Optional[str]:
-        return f"{self._prefix}/contract/0x{contract_address:064x}{self._get_network_query_param()}"
+    def create_link_to_contract(self, contract_address: Address) -> Optional[str]:
+        return f"{self._prefix}/contract/0x{int(contract_address):064x}{self._get_network_query_param()}"
```
Arcticae, line 21: regular string interpolation will do the trick here

protostar/starknet_gateway/gateway_response.py
```diff
@@ -22,6 +24,6 @@ class SuccessfulInvokeResponse:
 
 @dataclass
 class SuccessfulDeployAccountResponse:
-    address: int
+    address: AccountAddress
```
Arcticae, line 27: Regular Address is fine here. We should accept lowest-possible class in the hierarchy

protostar/commands/deploy_account_command.py
```diff
@@ -45,7 +46,7 @@ def from_args(cls, args: Namespace):
 
         assert signer is not None
         return cls(
-            account_address=args.account_address,
+            account_address=AccountAddress(int(args.account_address)),
```
Arcticae, line 49: why do we coerce it to int first? Shouldn't  be "from_user_input" used here?

protostar/commands/declare_command.py
```diff
@@ -163,7 +164,7 @@ async def declare(
             response = await gateway_facade.declare(
                 compiled_contract_path=compiled_contract_path,
                 signer=signer,
-                account_address=account_address,
+                account_address=AccountAddress(int(account_address)),
```
Arcticae, line 167: unnecessary casting to int

protostar/commands/invoke_command.py
```diff
@@ -177,7 +178,7 @@ async def invoke(
             inputs=inputs,
             max_fee=max_fee if max_fee is not None else "auto",
             signer=signer,
-            account_address=account_address,
+            account_address=AccountAddress(int(account_address)),
```
Arcticae, line 181: Unnecessary casting to int

Arcticae (changes requested):
I don't really see why `AccountAddress` exists, it only makes the code harder to maintain since sometimes we expect this type and sometimes the other, while the functionality itself stays the same. Additional constructor should suffice. Moreover, i think that  `Hash` class would be appropriate since it could be reused in tx_hashes, etc. (and it has the same representation and purpose). Maybe we could just do an alias.

protostar/starknet_gateway/block_explorer/stark_scan_block_explorer.py
kasperski95: I wasn't sure about this one because two address representation and url params can change at different rate. At the other hand, these two concepts are both stable. I'll change it.

protostar/starknet_gateway/gateway_response.py
kasperski95: > We should accept lowest-possible class in the hierarchy

I don't get it tbh. We should use the most specific available types.

protostar/commands/deploy_account_command.py
kasperski95: Argument parser converts this argument to `Address` first. I could avoid casting by introducing to `account_address`  type in `protostar_arg_type.py`

protostar/commands/declare_command.py
kasperski95: resolving in favor of https://github.com/software-mansion/protostar/pull/1170#discussion_r1025372518

protostar/commands/invoke_command.py
kasperski95: https://github.com/software-mansion/protostar/pull/1170#discussion_r1025372518

protostar/starknet_gateway/block_explorer/viewblock_block_explorer.py
kasperski95: resolving in favor of https://github.com/software-mansion/protostar/pull/1170#discussion_r1025363409

kasperski95:
> I don't really see why AccountAddress exists, it only makes the code harder to maintain since sometimes we expect this type and sometimes the other, while the functionality itself stays the same. Additional constructor should suffice. 

Type narrowing. You can't pass an Address to a function that expects AccountAddress. I'll remove this type because we would have to introduce a new argument type to the users to avoid converting Address to AccountAddress.

> Moreover, i think that Hash class would be appropriate since it could be reused in tx_hashes, etc. (and it has the same representation and purpose). Maybe we could just do an alias.

I'd rather do it in a different PR. An address (and class hash) should inherit from `Hash` value object.

protostar/starknet_gateway/gateway_response.py
kasperski95: Replaced AccountAddress with Address

PR #1176, opened 2022-11-17
Warn if cairo-path does not exist

kasperski95:
Tell the user that a cairo-path doesn't exist in case of typos.


kasperski95:
https://github.com/software-mansion/protostar/pull/1180 solves this problem. Arg type `directory`raises an error if cairo path doesn't exist.

PR #1180, opened 2022-11-18
Fix parsing non-string values from configuration file

### LINKED ISSUE #1178: Non-string values from configuration file don't go through the custom parsers
`argparse` parses default values only if they are strings.

kasperski95:
Run type parsers on arguments from configuration file. Argparse runs parsers on default values only if they are strings.

Closes #1178

### DIFF
--- a/protostar/argument_parser/unparser.py
+++ b/protostar/argument_parser/unparser.py
@@ -3,6 +3,8 @@ from typing import Any
 
 def unparse_flag_or_arguments(value: Any):
     """Arguments from external sources need to be unparsed in order to be parsed by custom parsers."""
+    if value is None:
+        return None
     if isinstance(value, bool):
         return value
     return unparse_arguments(value)
--- a/protostar/argument_parser/unparser.py
+++ b/protostar/argument_parser/unparser.py
@@ -21,7 +21,7 @@ def unparse_arguments(value: Any) -> list[str]:
 
 def unparse_single_value(value: Any) -> str:
     if isinstance(value, bool):
-        return str(bool)
+        return "true" if value else "false"
     if isinstance(value, int):
         return str(value)
     if isinstance(value, str):
--- a/protostar/argument_parser/unparser.py
+++ b/protostar/argument_parser/unparser.py
@@ -1,24 +1,24 @@
-from typing import Any
+from typing import Any, Optional
 
 
-def unparse_flag_or_arguments(value: Any):
+def unparse_flag_or_arguments(value: Any) -> Optional[list[str]]:
     """Arguments from external sources need to be unparsed in order to be parsed by custom parsers."""
     if value is None:
         return None
     if isinstance(value, bool):
-        return value
+        return None
     return unparse_arguments(value)
 
 
-def unparse_arguments(value: Any) -> str:
+def unparse_arguments(value: Any) -> list[str]:
     if not isinstance(value, list):
-        return unparse_single_value(value)
+        return [unparse_single_value(value)]
 
     values = value
     unparsed_values: list[str] = []
     for val in values:
-        unparsed_values.append(unparse_arguments(val))
-    return " ".join(unparsed_values)
+        unparsed_values.append(*unparse_arguments(val))
+    return unparsed_values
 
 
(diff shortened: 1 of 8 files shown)


protostar/argument_parser/unparser.py
```diff
@@ -0,0 +1,30 @@
...
+
+    values = value
+    unparsed_values: list[str] = []
+    for val in values:
+        unparsed_values.append(*unparse_arguments(val))
+    return unparsed_values
+
+
+def unparse_single_value(value: Any) -> str:
+    assert not isinstance(value, bool)
```
mkaput, line 25: why this assertion if we raise stuff later?

protostar/argument_parser/unparser.py
```diff
@@ -0,0 +1,30 @@
...
+    if isinstance(value, int):
+        return str(value)
+    if isinstance(value, str):
+        return value
+    assert False, f"Value '{value}' couldn't be unparsed"
```
mkaput, line 30:
```suggestion
    raise ValueError(f"Value {value!r} couldn't be unparsed")
```

protostar/argument_parser/unparser.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def unparse_arguments(value: Any) -> list[str]:
+    if not isinstance(value, list):
+        return [unparse_single_value(value)]
+
+    values = value
+    unparsed_values: list[str] = []
+    for val in values:
+        unparsed_values.append(*unparse_arguments(val))
```
Arcticae, line 20: This can be replaced with list comprehension

protostar/argument_parser/unparser.py
Arcticae, line 30: That's sketchy, i'd raise an error instead of asserting false.

protostar/argument_parser/unparser.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any, Optional
+
+
+def unparse_arguments_from_external_source(value: Any) -> Optional[list[str]]:
+    """Arguments from external sources need to be unparsed in order to be parsed by custom parsers."""
+    if value is None:
+        return None
+    if isinstance(value, bool):
```
Arcticae, line 8: What's the purpose of leaving out the boolean values? I don't understand the case.

Arcticae (changes requested):

protostar/argument_parser/unparser.py
kasperski95: "Bool" represents flags. Flags are treated a bit differently because they don't accept values.
| TOML | CLI |
| --- | --- |
| `arg-name = 42` | "42" |
| `flag-name = True` | ? |

I should've used assert with a description here.

protostar/argument_parser/unparser.py
kasperski95: > Pylance: Unpack operation not allowed in comprehension

protostar/argument_parser/unparser.py
kasperski95: This assertion is meant to be used as an assertion, not as an exception.

kasperski95: https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse

protostar/argument_parser/unparser.py
kasperski95: [fix bool parser](https://github.com/software-mansion/protostar/pull/1180/commits/9bf52693fcf0aafac9f1419892668d776dd9f07b)

protostar/argument_parser/unparser.py
kasperski95: Removed assertion. Added support for parsing and unparsing for booleans despite argparse not accepting explicit values for flags.

protostar/argument_parser/unparser.py
kasperski95: Done. Added support for parsing and unparsing for booleans despite argparse not accepting explicit values for flags.

mkaput (approved):

Arcticae (approved):

PR #1182, opened 2022-11-21
Adapt migrator for UDC + fix tests

Arcticae:


tests/e2e/test_migrating/migration_signed_invokes.cairo
```diff
@@ -3,7 +3,8 @@
 @external
 func up() {
     %{
-        contract  = deploy_contract("./build/main.json")
+        declaration = declare("./build/main.json", config={"max_fee": "auto"})
+        contract = deploy_contract(declaration.class_hash, config={"max_fee": "auto"})
```
kasperski95, line 7: Shouldn't `deploy_contract` (and the new `deploy` command) run `declare` under the hood?

protostar/migrator/cheatcodes/migrator_deploy_contract_cheatcode.py
```diff
@@ -1,32 +1,59 @@
...
     # pylint: disable=keyword-arg-before-vararg
     def __call__(
         self,
-        contract: str,
+        class_hash: Union[str, int],
         constructor_args: Optional[CairoOrPythonData] = None,
+        abi_path: Optional[str] = None,
+        salt: Optional[int] = None,
         *args: Any,
         config: Optional[Any] = None,
```
kasperski95, line 58: Some arguments are passed directly and some through config. We don't need to maintain a common interface for `deploy_contract` cheatcodes and since this is a breaking change, config can be removed.

tests/e2e/test_migrating/migration_signed_invokes.cairo
Arcticae: I don't think it should, since we don't want to declare the contract each time we deploy it (for example for performance reasons, or costs).

protostar/starknet_gateway/gateway_facade.py
```diff
@@ -166,12 +166,14 @@ async def deploy_via_udc(
             contract_address=call.udc.to_addr,  # type: ignore
             entry_point_selector=call.udc.selector,  # type: ignore
             calldata=call.udc.calldata,  # type: ignore
-            max_fee=max_fee or 0,
-            nonce=await self._gateway_client.get_contract_nonce(
-                call.udc.to_addr,
-            )
-            if signer
-            else None,
+            max_fee=max_fee if isinstance(max_fee, int) else 0,
```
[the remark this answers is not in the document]
Arcticae: This is solely because i want to acknowledge that we are not missing this parameter on purpose, and that it's being used. Doesn't hurt to leave it that way, and improves readability (the reader would wonder, why leave out the `max_fee`, since we take it as an argument?).

protostar/migrator/cheatcodes/migrator_deploy_contract_cheatcode.py
kasperski95: > The thing being that we can mantain a common interface without a huge hassle, is the reason we should keep the interface.

This PR breaks the common interface (between testing and migrator `deploy_contract`). The testing environment isn't going to use UDC. 

I'd update the docs in this PR to bring the repo from one consistent state to another.

kasperski95 (approved):

protostar/migrator/cheatcodes/migrator_deploy_contract_cheatcode.py
Arcticae: Merging into a side branch, so will update the docs in the next one.

PR #1187, opened 2022-11-22
Complete UDC support

### LINKED ISSUE #1103: Support UDC as a concept in protostar
We should consider using it both internally in cheatcodes, and adding a `deploy` command which would use a UDC and accept `class_hash` as an input, and using official UDCs to deploy.

Arcticae:
closes #1103


pyproject.toml
```diff
@@ -43,7 +43,7 @@ pytest-timeout = "^2.1.0"
 pytest-xdist = "^3.0.2"
 re-assert = "^1.1.0"
 snakeviz = "^2.1.1"
-starknet-devnet = "0.3.5"
+starknet-devnet = { git = "https://github.com/Shard-Labs/starknet-devnet.git", rev = "40d5b6f" }
```
mkaput, line 46: This should be pinned to real version

pyproject.toml
```diff
@@ -20,12 +20,12 @@ packaging = "^21.3"
 pytest-datadir = "^1.4.1"
 python = ">=3.9.14, <3.10"
 requests = "^2.27.1"
+starknet-py = { git="https://github.com/software-mansion/starknet.py", rev="b029de9" }
```
mkaput, line 23: This should be pinned to real version

mkaput (changes requested):

protostar/commands/deploy_command.py
```diff
@@ -30,19 +69,27 @@ def description(self) -> str:
...
                 is_positional=True,
             ),
+            ProtostarArgument(
+                name="max-fee",
+                description=(
+                    "The maximum fee that the sender is willing to pay for the transaction. "
+                    'Provide "auto" to auto estimate the fee.'
+                ),
+                type="fee",
+            ),
```
kasperski95, line 92: You can use `MAX_FEE_ARG` from `protostar.cli.common_args`.

kasperski95 (approved):
LGTM

PR #1188, opened 2022-11-22
Remove generating `cairo-paths` 

kasperski95:
- it doesn't work as expected because package's project root is usually in a subdirectory
- `lib-path` in the configuration file is declared in the `project` section is read directly however lib-path can be passed as argument resulting in weird behavior
- this behavior is undocumented (Do we need to deprecate it first?)
- this feature is confusing, it may work for some packages, but mostly it won't


protostar/compiler/project_cairo_path_builder.py
```diff
@@ -1,38 +1,18 @@
-import os
 from pathlib import Path
 
-from protostar.configuration_file import ConfigurationFile
-
 
 class ProjectCairoPathBuilder:
```
MaksymilianDemitraszek, line 4: Can we remove this class? It seems too small to abstract it

MaksymilianDemitraszek (approved):

protostar/compiler/project_cairo_path_builder.py
kasperski95: This logic is used in various places in the application despite being "lite".

kasperski95: Resolving. I don't want to mix refactors with functionality changes if possible.

MaksymilianDemitraszek: Can you make a task to remove it?

kasperski95: https://github.com/software-mansion/protostar/issues/1207

PR #1195, opened 2022-11-22
Estimate fee for test cases — 2

kasperski95:
Taking over https://github.com/software-mansion/protostar/pull/1013


protostar/commands/test/test_command.py
```diff
@@ -152,6 +152,16 @@ def arguments(self):
                 type="bool",
                 description="Only re-run failed and broken test cases.",
             ),
+            ProtostarArgument(
```
MaksymilianDemitraszek, line 155: Do we need involve gas price? It seems not very intuitive, I think gas estimation is more than enough for relative assesment, I would presonally change it to `--estimate-gas` boolean to simplify it.

tests/integration/commands/test_command_test.py
```diff
@@ -0,0 +1,24 @@
...
+        yield protostar
+
+
+async def test_gas_estimation(
+    protostar: ProtostarFixture,
+    devnet: DevnetFixture,
+    capsys: CaptureFixture[str],
+):
+    await protostar.test(
+        targets=["::test_increase_balance"], gas_price=devnet.get_gas_price()
```
MaksymilianDemitraszek, line 21: We do we need to query `gas_price` from devnet? Cannot we just set it to 1?

protostar/commands/test/test_command.py
```diff
@@ -152,6 +152,16 @@ def arguments(self):
...
+                name="gas-price",
+                type="float",
+                description=(
+                    "Gas price per gas unit. If provided, Protostar will estimate a fee for test cases. "
```
MaksymilianDemitraszek, line 159: I would add disclamer that those estimations may be heavily inaccurate

MaksymilianDemitraszek (changes requested):

protostar/commands/test/test_command.py
kasperski95: I left the decision to the user. If the user wants to estimate gas, he can set `--gas-price 1`.

tests/integration/commands/test_command_test.py
kasperski95: It's not needed, actually. Initially, I wanted to test the estimation result against `starknet estimate-fee`.

protostar/commands/test/test_command.py
kasperski95: If estimations may be heavily inaccurate, then this feature is pointless.

protostar/commands/test/test_command.py
```diff
@@ -186,6 +192,7 @@ async def test(
         seed: Optional[int] = None,
         max_steps: Optional[int] = None,
         slowest_tests_to_report_count: int = 0,
+        gas_estimation_enabled: bool = False,
```
Arcticae, line 195: why bother with `_enabled` since it's a boolean?

Arcticae (commented):
I am a bit worried that this makes the coupling with SN library go even further. 
Since a test run is an invocation (transaction), can't we get the transaction receipt to get the info on resource usage? We wouldn't have to use the internals that way.

protostar/commands/test/test_command.py
MaksymilianDemitraszek: We cannot, it is a problem with testing framework which destroys some information

protostar/commands/test/test_command.py
MaksymilianDemitraszek: Why?

MaksymilianDemitraszek: StarkNet estimations may be inaccurrate as well, it is an estimation

MaksymilianDemitraszek (approved):
Hav you compared the estimations with those provided by devnet?

protostar/commands/test/test_command.py
kasperski95: Estimations are inaccurate by definition, hence a disclaimer isn't needed.

protostar/commands/test/test_command.py
kasperski95: To make the name reflect the type. `gas_estimation` could be interpreted as an estimation result.

protostar/commands/test/test_command.py
MaksymilianDemitraszek: I think It wouldnt hurt to avoid any misunderstandings

kasperski95: [update description](https://github.com/software-mansion/protostar/pull/1195/commits/4c2325600ff2eea98710ae1fdfdff612e73408c5)

amanusk:
I have just tried this feature on some tests, the results don't seem to represent the true costs.
Here is an example of some test results:

Run1: 
```
time=17.32s, gas=4059, steps=147619, memory_holes=32442 pedersen_builtin=21 range_check_builtin=7495
```
Run2:
```
time=31.45s, gas=4059, steps=389198, memory_holes=30887 pedersen_builtin=21 range_check_builtin=6970
```


Resources on run2 are considerably lower (especially steps) but the gas result is the same

PR #1196, opened 2022-11-23
Raise error when test parameters not found

### LINKED ISSUE #1008: Raise an error instead of showing a warning for not providing test data
In #995, if neither `example` nor `given` is used, a warning is being displayed to a user:

```
"Not providing the test parameters is deprecated and will break test cases in the future releases, "
"please use one of the following cheatcodes in the setup function in order to "
"explicitly provide data to be tested: \n- example\n- given"
```

With the following release we may want to change that to raise an error instead of just showing a warning.

`test_no_data_broken` in `fuzzing_test.py` should be expected to be broken after that.

karol-bisztyga:
fix #1008

### DIFF
--- a/protostar/testing/environments/fuzz_test_execution_environment.py
+++ b/protostar/testing/environments/fuzz_test_execution_environment.py
@@ -1,5 +1,4 @@
 import dataclasses
-import logging
 from asyncio import to_thread
 from dataclasses import dataclass
 from typing import Any, Callable, Dict, List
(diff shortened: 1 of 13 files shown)


mkaput (approved):
I think you can also remove the logic which assumes default strategy? Subject for separate PR

karol-bisztyga:
@mkaput #1198 #1199

mkaput (approved):

PR #1199, opened 2022-11-23
Remove redundant strategy logic

### LINKED ISSUE #1198: Remove redundant logic for strategies
Raised in https://github.com/software-mansion/protostar/pull/1196#pullrequestreview-1191230375

We should detect and remove all the redundant logic for strategies since we resigned from having defaults for them.

karol-bisztyga:
fix #1198

### DIFF
--- a/protostar/testing/environments/fuzz_test_execution_environment.py
+++ b/protostar/testing/environments/fuzz_test_execution_environment.py
@@ -47,6 +47,7 @@ class FuzzTestExecutionEnvironment(TestExecutionEnvironment):
         if self.state.config.profiling:
             raise ProtostarException("Fuzz tests cannot be profiled")
         self.initial_state = state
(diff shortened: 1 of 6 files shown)


karol-bisztyga:
Blocked by #1196

mkaput (approved):

karol-bisztyga:
Ok, I added some new logic. the idea is to avoid using hypothesis when we have only explicit examples and just simply call the test function in the `for` loop.

@mkaput could you please take a look once again? Thanks!

mkaput (approved):
> Ok, I added some new logic. the idea is to avoid using hypothesis when we have only explicit examples and just simply call the test function in the `for` loop.
> 
> @mkaput could you please take a look once again? Thanks!

This wouldn't play well with future plans like examples database, **but**  because of the plans to write new test runner we can accept this I think.

PR #1212, opened 2022-11-25
Multicall use case

kasperski95:
Add multicall core functionality that can be exposed to the user in the next PR.

Related #1000


protostar/starknet_gateway/account_manager.py
```diff
@@ -0,0 +1,62 @@
...
+
+from .multicall import SignedMulticallTransaction, MulticallAccountManagerProtocol
+
+
+class AccountManager(MulticallAccountManagerProtocol):
+    def __init__(
+        self,
+        private_key: int,
+        address: Address,
+        signer: BaseSigner,
```
Arcticae, line 20: I saw that those three are also used in DevnetAccount - maybe it's worth to introduce this account structure somehow to our production code as well?

protostar/starknet_gateway/account_manager.py
```diff
@@ -0,0 +1,62 @@
...
+        gateway_client = GatewayClient(gateway_url)
+        self._account_client = AccountClient(
+            address=int(address),
+            client=gateway_client,
+            signer=signer,
+            key_pair=KeyPair.from_private_key(private_key),
+            supported_tx_version=1,
+        )
+
+    def get_account_address(self):
```
Arcticae, line 36: we could have a public property named `account_address`, set in the constructor instead.

protostar/starknet_gateway/multicall/multicall_protocols.py
```diff
@@ -0,0 +1,31 @@
...
+from protostar.starknet import Address
+
+from .multicall_structs import (
+    SignedMulticallTransaction,
+    MulticallClientResponse,
+    UnsignedMulticallTransaction,
+)
+
+
+class MulticallAccountManagerProtocol(Protocol):
```
Arcticae, line 14: Isn't that an ABC, not a protocol?

protostar/starknet_gateway/multicall/multicall_structs.py
```diff
@@ -0,0 +1,62 @@
...
+class DeployCall(CallBase):
+    class_hash: int
+    name: Optional[DeployCallName] = None
+
+
+@dataclass
+class ResolvedCall:
+    address: Address
+    selector: Selector
+    calldata: list[int]
```
Arcticae, line 32: Isn't the structure basically same as the `Call` from starknet.py?

protostar/starknet_gateway/account_manager.py
```diff
@@ -0,0 +1,62 @@
...
+                SNCall(
+                    to_addr=int(call.address),
+                    selector=int(call.selector),
+                    calldata=call.calldata,
+                )
+                for call in unsigned_transaction.calls
+            ],
+            max_fee=self._max_fee if isinstance(self._max_fee, int) else None,
+            auto_estimate=self._max_fee == "auto",
+            version=self._account_client.supported_tx_version,
```
Arcticae, line 53: we don't support v0, so if account only supports v0, user should upgrade his account.

protostar/starknet_gateway/account_manager.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+        assert tx.nonce is not None
```
Arcticae, line 55: Why this assertion?

protostar/starknet_gateway/multicall/call_resolver.py
```diff
@@ -0,0 +1,82 @@
...
+    CallBase,
+    InvokeCall,
+    DeployCall,
+    DeployCallName,
+    MulticallInputCalldata,
+    ResolvedCall,
+)
+
+
+class CallResolver:
```
Arcticae, line 19: The resolving idea seems nice, but could we make it more generic? So that a user could define some `$variable` in a file and i.e. pass it as a parameter to the script? We could raise on a missing parameter value. Of course this can be in a separate PR but i'd think a bit about the syntax here - a name on itself might be confusing to read. People are familiar with this kind of variable interpolation, so perhaps a deploy could have a "bind_to": "$address1" instead of "name" and then the concept is becoming more universal than just addresses. WDYT?

Arcticae (approved):
A few Qs and nitpicks, but the code looks fine to me.

protostar/starknet_gateway/account_manager.py
kasperski95: I think `AccountManager` is sufficient. `DevnetAccount` is a data structure which works well for tests, but production code needs an object.

protostar/starknet_gateway/account_manager.py
kasperski95: 1) I'd have to define an abstract property in the `MulticallAccountManagerProtocol` what creates awkward code. The use case class uses `AccountManager` indirectly through an interface to break a cyclic dependency.
2) I try to avoid creating hybrid objects. 

> Hybrids are classes that acts both as Data Structures (having public accessors and mutators)  and Objects (functions that do significant things)

protostar/starknet_gateway/multicall/multicall_structs.py
kasperski95: It's similar. This data structure uses value objects. Use case layer shouldn't be polluted by external library.

protostar/starknet_gateway/account_manager.py
kasperski95: NVM. I added this data structure to reduce boilerplate in tests. `DevnetAccount` will inherit from `Account` in a different PR because there's a mismatch between types of private keys and I don't want to do the integration in this PR.

[use account data structure](https://github.com/software-mansion/protostar/pull/1212/commits/1ea413230fc29fb7b8afb258c8d36be3c9bcb34e)

protostar/starknet_gateway/account_manager.py
kasperski95: [use optional nonce](https://github.com/software-mansion/protostar/pull/1212/commits/de3600cdd809e490f5a6cca7c13b75ff19e2db6f)

protostar/starknet_gateway/account_manager.py
kasperski95: Added account version validation.

[use account data structure](https://github.com/software-mansion/protostar/pull/1212/commits/1ea413230fc29fb7b8afb258c8d36be3c9bcb34e)

Arcticae (approved):

PR #1218, opened 2022-11-28
Remove unsigned transactions support

Arcticae:
Please forgive the size, the changes are not that complicated 🙄


tests/integration/gateway/gateway_facade_test.py
```diff
@@ -152,41 +206,59 @@ async def test_call_to_with_positional_incorrect_args(
...
+        result = await gateway_facade.deploy_via_udc(
             class_hash=compiled_contract_without_constructor_class_hash,
             abi=contract_abi,
             inputs={"UNKNOWN_INPUT": 42},
+            account_address=devnet_account.address,
+            signer=devnet_account.signer,
+            max_fee="auto",
+            wait_for_acceptance=True,
         )
+        print(result)
```
kasperski95, line 242: Print can be removed

tests/integration/protostar_fixture.py
```diff
@@ -257,6 +258,7 @@ async def invoke(
         function_name: str,
         inputs: Optional[CairoOrPythonData],
         gateway_url: str,
+        resolve_proxies: bool = False,
```
kasperski95, line 261: It looks like `resolve_proxies` isn't used by the command.

kasperski95 (approved):

tests/integration/protostar_fixture.py
Arcticae: right 👍

PR #1236, opened 2022-12-05
Allow preparing contract adresses deterministically

kasperski95:
Make `prepare` cheatcode accept a salt parameter to allow deterministic contract addresses.

### DIFF
--- a/protostar/testing/cheatcodes/prepare_cheatcode.py
+++ b/protostar/testing/cheatcodes/prepare_cheatcode.py
@@ -39,6 +39,7 @@ class PrepareCheatcode(Cheatcode):
         self,
         declared: DeclaredContract,
         constructor_calldata: Optional[CairoOrPythonData] = None,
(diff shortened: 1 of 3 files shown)


MaksymilianDemitraszek:
@kasperski95 Can you explain the use case of this feature 🤔 ? I believe the point of `prepare` was to deterministically calculate the contract address and hide `salt` behind the hood.

kasperski95:
@MaksymilianDemitraszek 

> I need the contract to verify a signature on some data that includes the contract address itself.
The signer is a bit trickier to re-implement, so I was hoping to create the signature elsewhere, and just provide the signed data in the test.
But if I get the contract deployed to a different address everytime I run protostar test, I cannot pre-prepare the signatre.

`prepare` hasn't been deterministic because salt was to increment every time it was used, and the deployer address changed with every run.

Arcticae (approved):

PR #1227, opened 2022-12-05
Multicall docs

kasperski95:
Closes #1000

PR #1242, opened 2022-12-06
Add docs for json flag

### LINKED ISSUE #1210: Add a section in documentation for `json` outputs of commands
In order to fill the void created by coming removal of migrations, we should probably hint the users to use the new flow since we're heading towards json commands output (and we'll be adding some interesting features to it in the future).

karol-bisztyga:
fix #1210

### DIFF
--- a/website/docs/tutorials/08-interacting-with-starknet/scripting.md
+++ b/website/docs/tutorials/08-interacting-with-starknet/scripting.md
@@ -3,7 +3,7 @@ In order to automate a process of your desire that includes protostar operations
 
 This tutorial shows a simple example of how to do such a thing using scripting in bash.
 
-We are going to write a script that builds, tests, declares, deploys and, in the end, calls the contract. We are also going to make use of the protostar's structured output so we can use previous' commands outputs in the following ones.
+We are going to write a script that builds, tests, declares, deploys, and, in the end, calls the contract. We are also going to make use of the protostar's structured output so we can use previous' commands outputs in the following ones.
 
 ### File Structure
 
@@ -19,18 +19,18 @@ First, let's create a basic protostar file structure. You can generate it by cal
 
 #### The contract
 
-`protostar init` automatically fills generated files with a sample content. We're not going to change them, because the example contract is sufficient for us and it is not important for this tutorial what it does exactly.
+`protostar init` automatically fills generated files with sample content. We're not going to change them, because the example contract is sufficient for us and it is not important for this tutorial what it does exactly.
 
 #### protostar.toml file
 
 You can read about how to compose the protostar configuration file [here](../04-configuration-file.md). The point is to keep protostar commands clean and simple and leave such things as the network configuration away from them.
 
-In this file, you should declare following sections:
+In this file, you should declare the following sections:
 - `declare`
 - `deploy`
 - `call`
 
-In these sections, you can specify properties like `network`, `account-address` or `private-key-path`.
+In these sections, you can specify properties like `network`, `account-address`, or `private-key-path`.
 
 ### Bash script
 
@@ -67,7 +67,7 @@ Normally, we would start with something like this:
 protostar declare ./build/main.json
 ```
 
-But `deploy` command needs the contract's class hash that comes from the `declare`'s command output. Therefore we need to get this output in a standardized way. That's when the `--json` flag comes into play.
+But the `deploy` command needs the contract's class hash that comes from the `declare`'s command output. Therefore we need to get this output in a standardized way. That's when the `--json` flag comes into play.
 
 By doing:
 
@@ -131,4 +131,4 @@ protostar call --contract-address $CONTRACT_ADDRESS --function get_balance
 
 Of course, we could exclude pulling a specific field from the output to a separate bash function but that's not important.
 
-Using `--json` flag is also good if you prefer the more compressed output from protostar commands.
+Using the `--json` flag is also good if you prefer the more compressed output from protostar commands.
--- a/website/docs/tutorials/08-interacting-with-starknet/scripting.md
+++ b/website/docs/tutorials/08-interacting-with-starknet/scripting.md
@@ -38,7 +38,7 @@ In these sections, you can specify properties like `network`, `account-address`,
 
 Let's start with something like this:
 
-```shell
+```shell title="automate_protostar_operations.sh"
 #!/bin/bash
 
 set -e
@@ -50,7 +50,7 @@ The `set -e` instruction tells the interpreter to exit the script immediately if
 
 #### Make sure the contract is correct
 
-```shell
+```shell title="automate_protostar_operations.sh"
 protostar build
 protostar test
 ```
@@ -89,7 +89,7 @@ Now, we can parse the json and pull all the desired information from it easily a
 
 We could do something like this in our bash script:
 
-```shell
+```shell title="automate_protostar_operations.sh"
 OUTPUT=$(protostar declare ./build/main.json --json)
 CLASS_HASH=$(python -c "import sys, json; print(json.loads(sys.argv[1])['class_hash'])" $OUTPUT)
 protostar deploy $CLASS_HASH --inputs 100
@@ -97,7 +97,7 @@ protostar deploy $CLASS_HASH --inputs 100
 
 You can use any alternative to python that will parse the json for you. This is how it would work with [jq](https://stedolan.github.io/jq/):
 
-```
+``` title="automate_protostar_operations.sh"
 OUTPUT=$(protostar declare ./build/main.json --json)
 CLASS_HASH=$(echo $OUT | jq -r ".class_hash")
 protostar deploy $CLASS_HASH --inputs 100
@@ -109,7 +109,7 @@ Now, let's say we want to call our contract. In this case, we need the contract 
 
 We are basically going to do the same thing as previously to pass the contract address from `deploy` to `call`:
 
-```shell
+```shell title="automate_protostar_operations.sh"
 OUT=$(protostar deploy $CLASS_HASH --inputs 100 --json)
 CONTRACT_ADDRESS=$(python -c "import sys, json; print(json.loads(sys.argv[1])['contract_address'])" $OUT)
 protostar call --contract-address $CONTRACT_ADDRESS --function get_balance
@@ -117,7 +117,7 @@ protostar call --contract-address $CONTRACT_ADDRESS --function get_balance
 
 In the end, our bash script looks like this:
 
-```shell
+```shell title="automate_protostar_operations.sh"
 #!/bin/bash
 
 set -e
@@ -143,7 +143,7 @@ The main advantage of using the JSON format is that the users may automate proto
 
 We can achieve the same result as above writing the equivalent python script:
 
-```python
+```python title="automate_protostar_operations.py"
 import subprocess, json
 
 def run_command(cmd):
--- a/website/docs/tutorials/08-interacting-with-starknet/scripting.md
+++ b/website/docs/tutorials/08-interacting-with-starknet/scripting.md
@@ -81,6 +81,10 @@ we get an output like this:
 {"class_hash":"0x07a70656b5612a2f87bd98af477c0be5fa2113d13fe1069e55ad326a3e6f4fe6","transaction_hash":"0x01f6a2c391d1bd0a51322ba73037ada20e0b30da8232bb86028f813a0d4c1fdb"} 
 ```
 
+:::note
+The `--json` output is, in fact, formatted in [NDJSON](https://github.com/ndjson/ndjson-spec), but here we operate on commands that return single outputs, therefore we can treat them as JSONs.
+:::
+
 Now, we can parse the json and pull all the desired information from it easily as json is a format that is widely supported.
 
 We could do something like this in our bash script:
--- a/website/docs/tutorials/08-interacting-with-starknet/scripting.md
+++ b/website/docs/tutorials/08-interacting-with-starknet/scripting.md
@@ -3,102 +3,36 @@ In order to automate a process of your desire that includes protostar operations
 
 This tutorial shows a simple example of how to do such a thing using scripting in bash.
 
+We are going to write a script that builds, tests, declares, deploys and, in the end, calls the contract. We are also going to make use of the protostar's structured output so we can use previous' commands outputs in the following ones.
+
 ### File Structure
 
-Let's create the following file structure:
+First, let's create a basic protostar file structure. You can generate it by calling `protostar init`. It looks like this:
 
 ```
 - src
-  - basic_contract.cairo
+  - main.cairo
 - tests
-  - test_basic_contract.cairo
+  - test_main.cairo
 - protostar.toml
 ```
 
-#### Basic contract
+#### The contract
 
-```cairo
-%lang starknet
-
-from starkware.cairo.common.cairo_builtins import HashBuiltin
-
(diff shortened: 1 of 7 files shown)


website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
...
+CLASS_HASH=$(python -c "import sys, json; print(json.loads(sys.argv[1])['class_hash'])" $DECLARE_OUTPUT)
+
+DEPLOY_OUTPUT=$(protostar deploy $CLASS_HASH --inputs 100 --json)
+CONTRACT_ADDRESS=$(python -c "import sys, json; print(json.loads(sys.argv[1])['contract_address'])" $DEPLOY_OUTPUT)
+protostar call --contract-address $CONTRACT_ADDRESS --function get_balance
+```
+
+Of course, we could exclude pulling a specific field from the output to a separate bash function but that's not important.
+
+Using `--json` flag is also good if you prefer the more compressed output from protostar commands.
```
karol-bisztyga, line 194: If this file is ok here, it might be a good idea to prepare a repository and paste a link here

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
...
+DECLARE_OUTPUT=$(protostar declare ./build/main.json --json)
+CLASS_HASH=$(python -c "import sys, json; print(json.loads(sys.argv[1])['class_hash'])" $DECLARE_OUTPUT)
+
+DEPLOY_OUTPUT=$(protostar deploy $CLASS_HASH --inputs 100 --json)
+CONTRACT_ADDRESS=$(python -c "import sys, json; print(json.loads(sys.argv[1])['contract_address'])" $DEPLOY_OUTPUT)
+protostar call --contract-address $CONTRACT_ADDRESS --function get_balance
+```
+
+Of course, we could exclude pulling a specific field from the output to a separate bash function but that's not important.
+
```
Arcticae, line 193: I would also recommend using `jq` if someone doesn't have `python` , as an alternative.

Arcticae (changes requested):
I would wait with this for multicall to be ready (which is pretty soon i  believe) and coordinate the documentation with that feature, since we might want to show a script with a multicall usage (common deployment scenario).

I think you should be OK to look at https://github.com/software-mansion/protostar/pull/1227 and incorporate some of the info into this page. (please consult with Arek)

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
+# Scripting
+In order to automate a process of your desire that includes protostar operations, you may want to assemble some protostar commands and put them in a single script.
+
+This tutorial shows a simple example of how to do such a thing using scripting in bash.
+
+### File Structure
+
+Let's create the following file structure:
```
MaksymilianDemitraszek, line 8:
```suggestion
Let's assume the following file structure:
```
Maybe would be better to just you standard template created by protostar init? User may want to do it alongside the tutorial

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
...
+}
+
+@view
+func get_balance{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (res: felt) {
+    let (res) = balance.read();
+    return (res,);
+}
+```
+
+#### Test for the contract
```
MaksymilianDemitraszek, line 54: You would avoid this section if you used init-generated template

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
...
+
+```shell
+OUT=$(protostar deploy $CLASS_HASH --inputs 100 --json)
+CONTRACT_ADDRESS=$(python -c "import sys, json; print(json.loads(sys.argv[1])['contract_address'])" $OUT)
+protostar call --contract-address $CONTRACT_ADDRESS --function get_balance
+```
+
+In the end, our bash script looks like this:
+
+```shell
```
MaksymilianDemitraszek, line 176: I would write an example script in python i believe it will be more useful for users

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
...
+
+    BasicContract.increase_balance(contract_address=contract_address, amount=50);
+    let (res,) = BasicContract.get_balance(contract_address=contract_address);
+    assert res = 150;
+
+    return ();
+}
+```
+
+#### protostar.toml file
```
MaksymilianDemitraszek, line 95: Can we remove this section, it is not relevant imo.

MaksymilianDemitraszek (changes requested):
Great job! 
Can we iterate to script in python though? I would like to empahise that users don't need to use bash.

website/docs/tutorials/08-interacting-with-starknet/scripting.md
MaksymilianDemitraszek: I think the fact you can use bash is pretty obvious for somebody that knows bash. 

> Most of the machines do have bash

Python as well. I would prefer to direct user towards python as it might be not obvious to the user that it is an option. We don't want a steep learning curve for protostar.

MaksymilianDemitraszek: Discussed on slack, we should use python as it is going to be simpler to read and reuse

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
+# Scripting
+In order to automate a process of your desire that includes protostar operations, you may want to assemble some protostar commands and put them in a single script.
+
+This tutorial shows a simple example of how to do such a thing using scripting in bash.
```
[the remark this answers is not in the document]
karol-bisztyga: Yes, good idea!

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: Great idea of using the `protostar init`, haven't thought of that, thanks!

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: I think it is relevant. Eventually, we want the commands in the script to be as simple as possible, we don't want something like this:
```
protostar declare ./build/main.json --gateway-url http://127.0.0.1:5050/ --account-address 0x0668c707Af664985EDfB3Ea616a0189660C2C54ADCa787B870B55DC314b91d94 --private-key-path priv.key --max-fee auto --chain-id 1536727068981429685321
```
Instead, I believe it is better and cleaner to have something like this:
```
protostar declare ./build/main.json
```
I think it is worth mentioning why all commands are so thin.

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
...
+
+    return ();
+}
+```
+
+#### protostar.toml file
+
+You can read about how to compose the protostar configuration file [here](../04-configuration-file.md). The point is to keep protostar commands clean and simple and leave such things as the network configuration away from them.
+
+### Bash script
```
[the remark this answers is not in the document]
karol-bisztyga: That's true, but I'd leave the header, it mark that the section of the script begins

karol-bisztyga:
> Great job! Can we iterate to script in python though? I would like to empahise that users don't need to use bash.

I'd like to discuss this a bit. Actually, I do not agree here. Essentially, this tutorial refers to the CLI, right? This means the commands are invoked in the command line, which is basically bash for the Linux/mac (ok, may be something else but I bet people who use something else are already familiar with bash). Also, there is not that much of the language itself in this script. I'm not using any advanced features and even the basic ones like loops or conditions. At most, there are a few commands and passing inputs around, it's nothing fancy. I consider it totally readable for people to be honest. If someone reads this tutorial, that means they can operate on the shell, and if they do, they'll be able to understand what's going on.

I'm also not sure how you imagine a python script here. Wouldn't be it just invoking the CLI commands through `subprocess` or something? How that script would look like?

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: Good idea to provide an alternative

website/docs/tutorials/08-interacting-with-starknet/scripting.md
MaksymilianDemitraszek: Ok I agree

website/docs/tutorials/08-interacting-with-starknet/scripting.md
MaksymilianDemitraszek: We can write about it but i would stick with python, it gives user a nice idea and makes it easier to start.

karol-bisztyga:
Discussed the python script offline, I added it at the end and kept the step-by-step tutorial in bash.

MaksymilianDemitraszek:
- [ ] Add information about `ndjson`

MaksymilianDemitraszek:
@karol-bisztyga Is this one ready for re-review?

karol-bisztyga:
I think so, yes

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,168 @@
+# Scripting
+In order to automate a process that includes protostar operations, you may want to build a script in language of your choice.
+
+This tutorial shows a simple example of how to do such a thing using scripting in bash.
+
+We are going to write a script that builds, tests, declares, deploys, and, in the end, calls the contract. We are also going to make use of the protostar's structured output so we can use previous' commands outputs in the following ones.
+
+### File Structure
+
+First, let's create a basic protostar file structure. You can generate it by calling `protostar init`. It looks like this:
```
MaksymilianDemitraszek, line 10: Can we add link to protostar init?

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+we get an output like this:
+
+```json
+{"class_hash":"0x07a70656b5612a2f87bd98af477c0be5fa2113d13fe1069e55ad326a3e6f4fe6","transaction_hash":"0x01f6a2c391d1bd0a51322ba73037ada20e0b30da8232bb86028f813a0d4c1fdb"} 
+```
+
+:::note
+The `--json` output is, in fact, formatted in [NDJSON](https://github.com/ndjson/ndjson-spec), but here we operate on commands that return single outputs, therefore we can treat them as JSONs.
```
MaksymilianDemitraszek, line 85:
```suggestion
The `--json` output is, in fact, formatted in [NDJSON](https://github.com/ndjson/ndjson-spec), but here we operate on commands that return single message, therefore we can treat them as JSONs.
```

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,168 @@
...
+CLASS_HASH=$(python -c "import sys, json; print(json.loads(sys.argv[1])['class_hash'])" $DECLARE_OUTPUT)
+
+DEPLOY_OUTPUT=$(protostar deploy $CLASS_HASH --inputs 100 --json)
+CONTRACT_ADDRESS=$(python -c "import sys, json; print(json.loads(sys.argv[1])['contract_address'])" $DEPLOY_OUTPUT)
+protostar call --contract-address $CONTRACT_ADDRESS --function get_balance
+```
+
+Of course, we could exclude pulling a specific field from the output to a separate bash function but that's not important.
+
+Using the `--json` flag is also good if you prefer the more compressed output from protostar commands.
```
MaksymilianDemitraszek, line 138:
```suggestion
```
I am not sure if this always the case

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Python script
+
+The main advantage of using the JSON format is that the users may automate protostar operations easily no matter what languages and technologies they use.
```
MaksymilianDemitraszek, line 142:
```suggestion
The main advantage of using the JSON format is that the user may automate protostar operations easily no matter what languages and technologies they use.
```

MaksymilianDemitraszek (approved):

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: I think we can do something between:

```
Using the `--json` flag may also be good if you prefer the more compressed output from protostar commands.
```

website/docs/tutorials/08-interacting-with-starknet/scripting.md
```diff
@@ -0,0 +1,194 @@
...
+- src
+  - basic_contract.cairo
+- tests
+  - test_basic_contract.cairo
+- protostar.toml
+```
+
+#### Basic contract
+
+```cairo
```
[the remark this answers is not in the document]
karol-bisztyga: [add titles](https://github.com/software-mansion/protostar/pull/1242/commits/99d1299b7a7ed94bef24bbf2ff1d61bcdc21f9fc)

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: [address feedback](https://github.com/software-mansion/protostar/pull/1242/commits/977c4c633d782c864696e16ec4f28783134b7531)

Please, let me know if you're okay with this so we can merge this, thx!

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: [address feedback](https://github.com/software-mansion/protostar/pull/1242/commits/977c4c633d782c864696e16ec4f28783134b7531)

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: [address feedback](https://github.com/software-mansion/protostar/pull/1242/commits/977c4c633d782c864696e16ec4f28783134b7531)

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: [address feedback](https://github.com/software-mansion/protostar/pull/1242/commits/977c4c633d782c864696e16ec4f28783134b7531)

website/docs/tutorials/08-interacting-with-starknet/scripting.md
karol-bisztyga: Since you're out until Jan 2nd, I'm going to resolve this, I don't think it is a big deal and you've already approved this. In case you strongly think we should change that, we can throw a tiny PR later.

Issue #1245, opened 2022-12-07
Invalid gas estimation


kasperski95:
I have just tried this feature on some tests, the results don't seem to represent the true costs.
Here is an example of some test results:

Run1: 
```
time=17.32s, gas=4059, steps=147619, memory_holes=32442 pedersen_builtin=21 range_check_builtin=7495
```
Run2:
```
time=31.45s, gas=4059, steps=389198, memory_holes=30887 pedersen_builtin=21 range_check_builtin=6970
```

Resources on run2 are considerably higher (especially steps) but the gas result is the same

_Originally posted by @amanusk in https://github.com/software-mansion/protostar/issues/1195#issuecomment-1340683586_

amanusk:
Tested this codebase

https://github.com/HerodotusDev/herodotus-eth-starknet/tree/v2

This is the command (latest protostar build)
```
 /home/amanusk/Code/Cairo/protostar/dist/protostar/protostar test --estimate-gas ./tests/contracts/test_l1_facts_registry.cairo::test_prove_account
```

PR #1246, opened 2022-12-07
Hotfix — hide gas estimation

kasperski95:
Hide gas estimation functionality to keep the mainline healthy. The fix will be in a different PR.

Related #1245


kasperski95:
Closing, found the bug. Fixing in a different PR.

PR #1249, opened 2022-12-07
Fix gas estimation

### LINKED ISSUE #1245: Invalid gas estimation

kasperski95:
Fixes #1245.


tests/integration/commands/test_command_test.py
```diff
@@ -18,4 +17,14 @@ async def test_gas_estimation(
 ):
     await protostar.test(targets=["::test_increase_balance"], estimate_gas=True)
 
-    assert "gas=4059" in capsys.readouterr().out
+    assert "gas=4356" in capsys.readouterr().out
+
+
+async def test_gas_estimation_for_identity_function(
```
MaksymilianDemitraszek, line 23:
``` suggestion
async def test_gas_estimation_for_empty_function(
```

MaksymilianDemitraszek (approved):
Can we make minor release including this?

kasperski95:
@MaksymilianDemitraszek Do the release then. Btw. multicall could be included in the release, if the following PR wasn't blocked: https://github.com/software-mansion/protostar/pull/1224.

PR #1254, opened 2022-12-08
Structured (json) output for the test command

### LINKED ISSUE #1172: Implement structured output for `test` command
Follow up of #1114, specifically for `protostar test` command.

karol-bisztyga:
fix #1172 

todo:

- [ ] docs


protostar/testing/test_scheduler.py
```diff
@@ -33,10 +73,12 @@ def run(
         exit_first: bool,
         testing_seed: Seed,
         max_steps: Optional[int],
-        project_root_path_str: str,
+        project_root_path: Path,
```
karol-bisztyga, line 76: looks weird that it was passed as a `str` before

protostar/testing/test_scheduler.py
```diff
@@ -2,16 +2,56 @@
...
+def make_path_relative_if_possible(test_result: TestResult, path: Path) -> TestResult:
+    try:
+        test_result = dataclasses.replace(
+            test_result,
+            file_path=test_result.file_path.resolve().relative_to(path.resolve()),
+        )
+    except ValueError:
+        # We do this to preserve the functionality of running tests that are outside of the project
+        pass
+    return test_result
```
karol-bisztyga, line 34: not sure if this place is good for this function, alternatively I can move this to a separate file like `test_utils.py` or something. Bottom line is, there was a circular dependency error.

protostar/commands/test/test_command.py
```diff
@@ -38,6 +40,58 @@
...
+    summary: TestingSummary
+    test_collector_result: TestCollector.Result
+
+    def format_human(self, fmt: LogColorProvider) -> str:
+        assert False, "Tests should use live logging for the human-readable output"
+
+    def format_dict(self) -> dict:
+        test_cases = 0
+        for test_suite in self.test_collector_result.test_suites:
+            test_cases += len(test_suite.test_cases)
```
karol-bisztyga, line 54: I wanted to use `sum` + comprehensive list but the linter complained. I guess this is because the comprehensive list allocates an extra list in the memory.

protostar/commands/test/test_command.py
```diff
@@ -38,6 +40,58 @@
...
+        failed_tests_paths = {str(item.file_path) for item in self.summary.failed}
+        passed_test_suites = failed_test_suites = 0
+        for test_suite in self.test_collector_result.test_suites:
+            if str(test_suite.test_path) in failed_tests_paths:
+                failed_test_suites += 1
+            else:
+                passed_test_suites += 1
+
+        return {
+            "collected_tests": {
```
kasperski95, line 70: I think `TestCollectingResult` deserves the own `Message` class.

kasperski95 (dismissed):

protostar/commands/test/test_command.py
```diff
@@ -38,6 +40,58 @@
...
+                "test_cases": test_cases,
+                "duration": self.test_collector_result.duration,
+            },
+            "summary": {
+                "test_suites": {
+                    "total": failed_test_suites + passed_test_suites,
+                    "failed": failed_test_suites,
+                    "passed": passed_test_suites,
+                },
+                "tests": {
```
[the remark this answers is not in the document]
karol-bisztyga: Ok, but then I'd use `test_suite_counts`

protostar/commands/test/test_command.py
```diff
@@ -38,6 +40,58 @@
...
+                ),
+                "test_suites": len(self.test_collector_result.test_suites),
+                "test_cases": test_cases,
+                "duration": self.test_collector_result.duration,
+            },
+            "summary": {
+                "test_suites": {
+                    "total": failed_test_suites + passed_test_suites,
+                    "failed": failed_test_suites,
+                    "passed": passed_test_suites,
```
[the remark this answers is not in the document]
karol-bisztyga: I will add info about the skipped ones, but as of broken, there's already an information about them in the collector result.

protostar/commands/test/test_command.py
karol-bisztyga: I'm not sure what's the gain here. Also, should we create a separate class for the summary as well? This looks pretty short and I'm not sure it's going to be reused.

protostar/commands/test/test_command.py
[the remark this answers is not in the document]
karol-bisztyga: [address feedback](https://github.com/software-mansion/protostar/pull/1254/commits/376aab9515abdfc82a9dc7d61378975c0e4773a2)

protostar/commands/test/test_command.py
```diff
@@ -38,6 +40,58 @@
...
+                },
+                "tests": {
+                    "total": failed_tests + passed_tests,
+                    "failed": failed_tests,
+                    "passed": passed_tests,
+                },
+                "seed": self.summary.testing_seed,
+                "execution_time": f"{sum(execution_times):0.2f}",
```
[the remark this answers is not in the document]
karol-bisztyga: [address feedback](https://github.com/software-mansion/protostar/pull/1254/commits/376aab9515abdfc82a9dc7d61378975c0e4773a2)

protostar/commands/test/test_command.py
```diff
@@ -38,6 +40,58 @@
...
+        for test_suite in self.test_collector_result.test_suites:
+            if str(test_suite.test_path) in failed_tests_paths:
+                failed_test_suites += 1
+            else:
+                passed_test_suites += 1
+
+        return {
+            "collected_tests": {
+                "broken_test_suites": len(
+                    self.test_collector_result.broken_test_suites
```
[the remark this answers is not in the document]
karol-bisztyga: [address feedback](https://github.com/software-mansion/protostar/pull/1254/commits/376aab9515abdfc82a9dc7d61378975c0e4773a2)

protostar/commands/test/test_command.py
kasperski95: Protostar shows collected tests information before running tests, unlike the testing summary. The JSON argument shouldn't change messages order. It's confusing for users and developers.

Protostar might be converted to a plugin of the CargoLikeTool and JSON format could be used to communicate with Protostar, so messages shouldn't be delayed if it can be avoided.

> should we create a separate class for the summary as wel

This data class can be converted to represent the summary.

protostar/testing/testing_summary.py
```diff
@@ -19,6 +19,12 @@
 from .testing_seed import Seed
 
 
+def calculate_skipped(
+    total_count: int, broken_count: int, failed_count: int, passed_count: int
+):
+    return total_count - (broken_count + failed_count + passed_count)
```
kasperski95, line 25: `skipped_test_cases_count` and `skipped_test_suites_count` could be a `@property` of `TestingSummary`.

protostar/commands/test/test_command.py
```diff
@@ -174,8 +247,16 @@ async def run(self, args: Namespace) -> TestingSummary:
...
+            json_format=json_format,
         )
         cache.write_failed_tests_to_cache(summary)
+        if json_format:
+            if self._messenger_factory is None:
+                raise ProtostarException(
+                    "Messanger factory has to be provided in order to format the output"
+                )
+            write = self._messenger_factory.json()
+            write(SuccessfulTestMessage(summary, test_collector_result))
```
kasperski95, line 259: Why do you raise an error to the user? Why is the `MessangerOptional` optional?

protostar/testing/test_scheduler.py
```diff
@@ -33,10 +77,12 @@ def run(
         exit_first: bool,
         testing_seed: Seed,
         max_steps: Optional[int],
-        project_root_path_str: str,
+        project_root_path: Path,
         cwd: Path,
         active_profile_name: Optional[str],
         gas_estimation_enabled: bool,
+        testing_summary: TestingSummary,
+        json_format: bool = False,
```
kasperski95, line 85: `TestScheduler` shouldn't know the format (and logging in general). TestScheduler and LiveLogger need to be refactored before adding JSON support.

karol-bisztyga:
Please do not review this ATM

protostar/commands/test/test_command.py
karol-bisztyga: Discussed offline

protostar/commands/test/test_command.py
karol-bisztyga: Not related anymore

protostar/commands/test/test_result_structured_formatter.py
```diff
@@ -0,0 +1,193 @@
+from typing import Callable
```
karol-bisztyga, line 1: todo tests for this file

protostar/commands/test/testing_live_logger.py
```diff
@@ -34,6 +38,55 @@ def __init__(
...
+        }
+        passed_test_suites = failed_test_suites = 0
+        for test_suite in test_collector_result.test_suites:
+            if str(test_suite.test_path) in failed_tests_paths:
+                failed_test_suites += 1
+            else:
+                passed_test_suites += 1
+
+        return {
+            "type": "test_summary",
```
karol-bisztyga, line 63: we can standardize newly introduced structures with dataclasses

karol-bisztyga (commented):
I think this can be reviewed

protostar/testing/test_scheduler.py
karol-bisztyga: Can you specify what refactoring you mean? Is there a task for this? Are you sure that refactoring cannot be applied after this PR is merged?

protostar/testing/testing_summary.py
karol-bisztyga: They weren't before these changes, I'd address it in another PR.

protostar/commands/test/testing_live_logger.py
karol-bisztyga: What I mean by this: dicts that are turned into jsons and printed out, in the scope of this PR, can be wrapped into something like this:

```
@dataclass
class StandarizedOutput:
  type: Literal["test_summary", "test_collector_result", "test_case_result"]
```
Then, we could have some children for any repetitive fields, like this:
```
@dataclass
class StandarizedTestCaseResult(StandarizedOutput):
  type = "test_case_result"
  status: Literal[...]
  test_suite_path: str
  ...
```

etc, what do you think?

protostar/commands/test/test_result_structured_formatter.py
```diff
@@ -0,0 +1,193 @@
...
+            test_case_name=passed_test_case_result.test_case_name,
+            fuzz_runs_count=None,
+        )
+    )
+
+
+def _format_failed_test_case_result(
+    failed_test_case_result: FailedTestCaseResult,
+) -> str:
+    result: PythonData = {
```
kasperski95, line 64: I don't see a conceptual connection between formatting result and `PythonData`. The result of formatting should be just a dict that can be represented as JSON.

protostar/commands/test/test_result_structured_formatter.py
kasperski95, line 1: Why don't you use `StructuredMessage`?

protostar/testing/testing_summary.py
kasperski95: Why not in this PR?

kasperski95 (commented):
`StructuredMessage` and `Messenger` aren't used.

protostar/commands/test/test_result_structured_formatter.py
karol-bisztyga: This is correct, I used it because it's essentially `Dict[str, Any]` which seemed to fit the purpose. I can just rename it for this case, no problem (I'm not sure if there's a way to type a JSON-serializable dict).

protostar/commands/test/test_command.py
```diff
@@ -163,6 +168,7 @@ def arguments(self):
 
     async def run(self, args: Namespace) -> TestingSummary:
         cache = TestCommandCache(CacheIO(self._project_root_path))
+        structured_format = args.json if hasattr(args, "json") else False
```
MaksymilianDemitraszek, line 171: Why do we need to use `hasattr` here?

protostar/commands/test/test_command.py
```diff
@@ -245,7 +254,24 @@ async def test(
                 "Only one test case can be profiled at the time. Please specify path to a single test case."
             )
 
-        self._log_test_collector_result(test_collector_result)
+        if structured_format:
+            test_cases = 0
+            for test_suite in test_collector_result.test_suites:
+                test_cases += len(test_suite.test_cases)
```
MaksymilianDemitraszek, line 260: Please add `_count` prefix, test_cases suggests list with `TestCase` type.

[[Nitpick]] Can we also move out counting from this function? It clutters the code. I would add `test_cases_count` property to TestSuite and then write it like this.

`test_cases_count = sum(suite.test_cases_count for suite in test_collector_result.test_suites)`

protostar/commands/test/test_command.py
```diff
@@ -245,7 +254,24 @@ async def test(
...
+            collected_tests = {
+                "type": "test_collector_result",
+                "broken_test_suites_count": len(
+                    test_collector_result.broken_test_suites
+                ),
+                "test_suites": len(test_collector_result.test_suites),
```
MaksymilianDemitraszek, line 266: `test_cases_count`
`test_suites_count`

protostar/commands/test/test_result_structured_formatter.py
```diff
@@ -0,0 +1,193 @@
...
+        "test_suite_path": str(unexpected_exception_test_suite_result.file_path),
+        "traceback": unexpected_exception_test_suite_result.traceback,
+        "exception": str(unexpected_exception_test_suite_result.exception),
+        "protostar_message": UNEXPECTED_PROTOSTAR_ERROR_MSG,
+    }
+
+    return json.dumps(result)
+
+
+def get_formatted_execution_time(execution_time: float) -> float:
```
MaksymilianDemitraszek, line 192: What is the purpose of this function? If rounding to 2dec places then you can use 
`"{:.2f}".format(execution_time)` instead.

protostar/testing/test_scheduler.py
```diff
@@ -2,16 +2,60 @@
...
+            test_result,
+            file_path=test_result.file_path.resolve().relative_to(path.resolve()),
+        )
+    except ValueError:
+        # We do this to preserve the functionality of running tests that are outside of the project
+        pass
+    return test_result
+
+
+def _update_summary(
```
MaksymilianDemitraszek, line 37: Is this used?

protostar/testing/testing_summary.py
MaksymilianDemitraszek: I would do this in this PR, it is small, related, change and good suggestion

protostar/commands/test/testing_live_logger.py
MaksymilianDemitraszek: Is there a value of creating such structures? Those seem redundant with structures we have already

protostar/commands/test/test_command.py
karol-bisztyga: we don't, I will remove this

karol-bisztyga: we can use `vars(args).get("json")`

protostar/commands/test/test_command.py
karol-bisztyga: you mean suffix, right? :D 

What you proposed doesn't really change anything. If I change my code to a one-liner like that, we will have:
```
test_cases_count = [sum(len(test_suite.test_cases) for test_suite in test_collector_result.test_suites)]
```
So we'd end up just adding the field `test_cases_count ` to `TestSuite` which would basically be `len(test_cases)`, right?

protostar/commands/test/test_result_structured_formatter.py
karol-bisztyga: I know, I first did this with the `format` function but back then I wanted to have floats there, now we can use string right away I think.

protostar/commands/test/test_command.py
karol-bisztyga: Also, lint keeps complaining: "consider using a generator" for this (I think I probably mentioned this in this PR), I don't think a simple for loop here is such a big deal overall.

protostar/testing/test_scheduler.py
karol-bisztyga: Apparently no, I must've forgotten to remove it, sorry

protostar/commands/test/testing_live_logger.py
karol-bisztyga: It was just a suggestion. Right now we have 3 types of these outputs:
- collector result
- test case result
- test summary

Only test case results are standardized, the rest are kept in a similar manner and are parts of the testing output but are defined implicitly. I don't think this would be an unnecessary boilerplate but I don't feel strongly about adding them. I don't mind, just saying it feels like they should be wrapped.

protostar/commands/test/test_result_structured_formatter.py
karol-bisztyga: [wrap in the StructuredMessage](https://github.com/software-mansion/protostar/pull/1254/commits/e2c3d62e4693b96fe89949030c488c378772caf2)

protostar/commands/test/test_command.py
karol-bisztyga: naming is resolved in [format + names + remove redundant function](https://github.com/software-mansion/protostar/pull/1254/commits/f62b551533db9762d7ce549c5fa2823cc8965a14)

Please, let me know if you're okay with leaving the for loop

protostar/commands/test/test_result_structured_formatter.py
karol-bisztyga: [format + names + remove redundant function](https://github.com/software-mansion/protostar/pull/1254/commits/f62b551533db9762d7ce549c5fa2823cc8965a14)

protostar/testing/test_scheduler.py
karol-bisztyga: [format + names + remove redundant function](https://github.com/software-mansion/protostar/pull/1254/commits/f62b551533db9762d7ce549c5fa2823cc8965a14)

protostar/testing/testing_summary.py
karol-bisztyga: I don't think this is a super small, tiny change.

The problem here is that you want to make `skipped_tests` a part of the testing summary but we need a test collector result to calculate the skipped tests as well. That means that we can do one of these things:
- always calculate the skipped tests count when creating a test summary (which bonds the collector result existence with the testing summary creation in one scope)
- sometimes calculate the skipped tests count immediately when creating a testing summary and sometimes do this lazily
- always calculate the skipped tests count lazily (when needed). That may result in introducing some optionals/conditions, etc.

This PR is already pretty big and I'm afraid if we keep pumping up changes like that into it, it's going to turn unmanageable. Looks like it's about time to start decomposing work here. I didn't look that closely at the code, so I'm basically driven by my intuition here, nevertheless, I do not see any disadvantages in opening another PR for this. At most, we're going to have a tiny PR and just merge it fast.

I know it's 2 vs 1 but I have a pretty strong opinion that we should deal with this separately. If you insist, I can try to resolve this here.

protostar/commands/test/test_command.py
MaksymilianDemitraszek: >Please, let me know if you're okay with leaving the for loop
Ofc I am, It is marked as nitpick 😉

MaksymilianDemitraszek: FYI
```
 test_cases_count = sum(len(test_suite.test_cases) for test_suite in test_collector_result.test_suites)
```
You do not need square brackets, then lint will not complain and sum is executed lazily. I dont heave heavy preference between your and this solution

karol-bisztyga: Right, this works, thanks!

[use one-liner](https://github.com/software-mansion/protostar/pull/1254/commits/2c2ed597f95ecca8013b75fc9366028b065a49df)

protostar/commands/test/test_result_structured_formatter.py
```diff
@@ -0,0 +1,224 @@
...
+    steps: Optional[str] = None
+    memory_holes: Optional[str] = None
+    builtins: Optional[list[JsonData]] = None
+
+    reason: Optional[str] = None
+
+    type: str = "test_case_result"
+
+    def format_human(self, fmt: LogColorProvider) -> str:
+        assert False, "Tests should use live logging for the human-readable output"
```
kasperski95, line 48: This method should be implemented. Instead of creating one `UberTestCaseResultMessage` I'd create a message for each testing result (or formatter function). `LiveLogger` should use messages. I'd create a new directory `protostar/commands/test/messages` and keep each message in a separate file.

protostar/testing/test_scheduler.py
kasperski95: Nvm. It can be postponed. I thought `LiveLogger` is in the protostar/testing module and that would cause cyclic dependency problems.

protostar/testing/testing_summary.py
kasperski95: `TestingSummary` has `skipped` property that can be used to calculate skipped test cases count. To display skipped test suites count, I'd create a method in testing summary that accepts total test suites count because it contains most of the required information.

> Information expert (also expert or the expert principle) is a principle used to determine where to delegate responsibilities such as methods, computed fields, and so on.
Using the principle of information expert, a general approach to assigning responsibilities is to look at a given responsibility, determine the information needed to fulfill it, and then determine where that information is stored.
This will lead to placing the responsibility on the class with the most information required to fulfill it.
https://en.wikipedia.org/wiki/GRASP_(object-oriented_design)

protostar/commands/test/test_result_structured_formatter.py
```diff
@@ -0,0 +1,224 @@
...
+)
+from protostar.io import StructuredMessage, LogColorProvider
+from protostar.testing import OutputName
+
+JsonData = dict[str, Any]
+
+
+@dataclass
+# pylint: disable=too-many-instance-attributes
+class TestCaseResult(StructuredMessage):
```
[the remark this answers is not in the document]
karol-bisztyga: sure

protostar/commands/test/test_result_structured_formatter.py
karol-bisztyga: I understand this comment as two comments actually that are unrelated to each other:

1.
> This method should be implemented.

As I understand, you referred to `format_human`. Why do you think it should be implemented? This part of the functionality stands as the opposite of the human-readable output and should print only JSON, right? Can you specify what output would be expected here and why should we implement this?

2. 

> Instead of creating one `UberTestCaseResultMessage` I'd create a message for each testing result (or formatter function). `LiveLogger` should use messages. I'd create a new directory `protostar/commands/test/messages` and keep each message in a separate file.

This is something totally different and I agree that we can create a hierarchy here and split the message class into multiple smaller message classes, one for each type of test result.

kasperski95: > As I understand, you referred to format_human. Why do you think it should be implemented?

It's part of the Marek's design. That's why `StructuredMessage` forces you to implement `format_human` and this code fights against the design constraints, which results in:
- code duplication
  - `format_test_result_structured`
  - `json.dumps` in `TestingLiveLogger` is already implemented in `JsonMessenger` which I've just noticed is configured differently than `json.dumps` in `TestingLiveLogger`. 
- lack of consistency with the way how the `StructuredMessage` is used by other commands.

> This part of the functionality stands as the opposite of the human-readable output and should print only JSON, right? Can you specify what output would be expected here and why should we implement this?

`format_human` should return the same thing what `format_test_result` returns.

karol-bisztyga: Please check out the newest changes

protostar/commands/test/test_command.py
```diff
@@ -245,7 +261,24 @@ async def test(
...
+                    len(test_suite.test_cases)
+                    for test_suite in test_collector_result.test_suites
+                ),
+                "duration_in_seconds": get_formatted_execution_time_structured(
+                    test_collector_result.duration
+                ),
+            }
+            print(json.dumps(collected_tests))
+        else:
+            self._log_test_collector_result(test_collector_result)
```
kasperski95, line 281: All `_log_*` methods in `TestCommand` should be replaced with messages.

protostar/commands/test/messages/test_case_result_message.py
```diff
@@ -0,0 +1,65 @@
...
+
+@dataclass
+class TestCaseResultMessage(StructuredMessage):
+    @abstractmethod
+    def format_human(self, fmt: LogColorProvider) -> str:
+        assert False
+
+    @abstractmethod
+    def format_dict(self) -> dict:
+        assert False
```
kasperski95, line 25: It looks like this class is no longer necessary.

protostar/commands/test/test_result_formatter.py
```diff
@@ -75,248 +66,53 @@ def _format_passed_test_case_result(
...
-    first_line_elements: List[str] = []
-    first_line_elements.append(f"[{log_color_provider.colorize('GREEN', 'PASS')}]")
-    formatted_file_path = _get_formatted_file_path(
-        passed_fuzz_test_case_result.file_path
-    )
-    first_line_elements.append(
-        f"{formatted_file_path} {passed_fuzz_test_case_result.test_case_name}"
+) -> PassedFuzzTestCaseResultMessage:
+    return PassedFuzzTestCaseResultMessage(
+        passed_fuzz_test_case_result=passed_fuzz_test_case_result
```
kasperski95, line 83: These functions could be removed in favor of creating messages directly in `format_test_result`.

protostar/commands/test/testing_live_logger.py
```diff
@@ -47,8 +99,9 @@ def log(
         self,
         shared_tests_state: SharedTestsState,
         test_collector_result: "TestCollector.Result",
+        structured_format: bool,
+        messanger: Optional[Messenger] = None,
```
kasperski95, line 103:
```suggestion
        messenger: Optional[Messenger] = None,
```

kasperski95, line 103:
```suggestion
        messanger: Messenger,
```
Messenger should always be required.

protostar/commands/test/testing_live_logger.py
```diff
@@ -93,9 +154,18 @@ def log(
...
-                    progress_bar.clear()
-                    self.log_testing_summary(test_collector_result)
+                    if structured_format:
+                        print(
+                            json.dumps(
+                                self.get_structured_testing_summary(
+                                    test_collector_result
+                                )
+                            )
+                        )
```
kasperski95, line 164: `print + jsons.dumps` shouldn't be used anywhere in this PR. That's responsibility of `JsonMessenger` which also ensures the output is ndjson friendly.

protostar/commands/test/messages/test_case_result_message.py
karol-bisztyga: Ah, right, I added it because it would specify some fields that would be required in the children's classes but had to resign from that in the meantime, thanks for catching it!

protostar/commands/test/messages/testing_summary_message.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+@dataclass
+class TestingSummaryResultMessage(StructuredMessage):
+    test_collector_result: TestCollector.Result
+    testing_summary: TestingSummary
+    slowest_tests_to_report_count: int
+
+    _bold_begin = "\033[1m"
+    _bold_end = "\033[0m"
```
kasperski95, line 153: Why not use `fmt.bold`?

protostar/commands/test/test_command.py
```diff
@@ -162,6 +158,12 @@ def arguments(self):
         ]
 
     async def run(self, args: Namespace) -> TestingSummary:
+        structured_format = False
+        if vars(args).get("json"):
+            structured_format = args.json
+        else:
+            args.json = None
```
kasperski95, line 165: This code could be removed. You can use `isinstance(HumanMessenger)` to use the progress bar in the live logger.

kasperski95 (commented):
Looks ok.

- IMO progress bar shouldn't be used when `--json` flag is provided.
- `TypeError: Object of type BreakingReportedException is not JSON serializable`

karol-bisztyga:
Sorry, I had to force-push because of the pylint version incompatibility between the branch and the CI

protostar/commands/test/messages/testing_summary_message.py
karol-bisztyga: [use fmt.bold](https://github.com/software-mansion/protostar/pull/1254/commits/16641b396eb7a339c8849ab928b053e4fd4207f1)

protostar/commands/test/test_command.py
karol-bisztyga: [use isinstance HumanMessenger instead of additional variable](https://github.com/software-mansion/protostar/pull/1254/commits/7b99db00b0fa7cb38b6323eab3a4c92525116de4)

kasperski95 (commented):
I'll take a look once tests pass.

protostar/commands/test/testing_live_logger.py
```diff
@@ -34,21 +35,49 @@ def __init__(
...
+                else:
+                    tests_left_n -= 1
+        finally:
+            messenger(
+                TestingSummaryResultMessage(
+                    test_collector_result=test_collector_result,
+                    testing_summary=self.testing_summary,
+                    slowest_tests_to_report_count=self.slowest_tests_to_report_count,
+                )
+            )
```
kasperski95, line 73: Ideally, we shouldn't duplicate the queue consumer. This can be refactored in a different PR.

kasperski95 (approved):

protostar/commands/test/messages/broken_test_case_result_message.py
```diff
@@ -0,0 +1,74 @@
...
+                captured_stdout=self.broken_test_case_result.captured_stdout,
+                log_color_provider=fmt,
+            )
+        )
+
+        return "".join(result)
+
+    def format_dict(self) -> dict:
+        return {
+            "type": "test_case_result",
```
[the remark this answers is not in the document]
karol-bisztyga: I'm not sure this is a good idea. I think it's good to have a mutual type for all of these messages, it helps distinguish them from test collector results and test summaries.
I know we could have a mutual part of the type like `test_case_result_failed_test_case`, `test_case_result_skipped_test_case`, etc. But at the same time, we've got the `status` field which would basically duplicate the information on the specific type of message.

We could leave the type as is and make `status`es more specific, so instead of having `status: failed`, we'd have `status: failed_test_case`, etc. What do you think?

protostar/commands/test/messages/__init__.py
```diff
@@ -0,0 +1,15 @@
+from .utils import (
```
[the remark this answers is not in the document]
karol-bisztyga: [rename utils => formatters](https://github.com/software-mansion/protostar/pull/1254/commits/fc631890c3a3d609b6120f1c610cc840fd372da3)

protostar/commands/test/messages/broken_test_case_result_message.py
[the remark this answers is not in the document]
kasperski95: >  think it's good to have a mutual type for all of these messages, it helps distinguish them from test collector results and test summaries.

This problem can be solved by common suffix.

For example:
`passed_test_case_result`
`failed_test_case_result`
`fuzz_failed_test_case_result`
`test_summary`

To find all test cases result you can run:
```py
[msg for msg in msgs if msg.type.ends_with("test_case_result")]
```

>We could leave the type as is and make statuses more specific, so instead of having status: failed, we'd have status: failed_test_case, etc. What do you think?

Protostar also returns fuzzing test results, and it might be helpful to have an easy way to distinguish such results from others. IMO `type` property is more fitting for storing that information.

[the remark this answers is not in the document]
karol-bisztyga: Yes, I understand, as I said:

> I know we could have a mutual part of the type like `test_case_result_failed_test_case`, `test_case_result_skipped_test_case`, etc.

So I know you can operate on the string to pull the info you're interested in.

I have yet another solution in mind. The case here is that you're thinking of the test type and I'm thinking about the message type. I think it'd be profitable to keep both pieces of information in the result. How about this then: We can keep the current type and rename it to the `message_type` or `result_type` and add another field, let's say `test_type` which would be `passed_test_case_result`/`failed_test_case_result`, etc. That way we could distinguish the results of test cases/test summary/test collector between each other and also distinguish the test case results between each other. I think that in this case, we could remove the `status` field. 
I think this solution is not expensive and it's going to be easier to use - I think that the possibility of separating different messages easily in the test output may be very handy.
What do you think?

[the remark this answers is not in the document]
kasperski95: IMO, result types are a subset of message types, but "test_type" should be ok.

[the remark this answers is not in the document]
karol-bisztyga: [type => message_type, status => test_type](https://github.com/software-mansion/protostar/pull/1254/commits/e850cb777dc38a105486a4aaf1101905dc23e6b0)

protostar/commands/test/testing_live_logger.py
karol-bisztyga: Why did you resolve this conversation without providing a reason for it? What happened? Did you create an issue for it? Did you already post a PR? Did you change your mind about it?

protostar/testing/testing_summary.py
karol-bisztyga: > `TestingSummary` has `skipped` property that can be used to calculate skipped test cases count.

I disagree. I looked at this and I think we have to rely strictly on the test collector result. If we use the `-x` flag, the tests are going to be omitted and therefore not added explicitly to the `skipped` property. We have to calculate the difference between cases collected by the collector and cases that have been executed in the current run. 

> To display skipped test suites count, I'd create a method in testing summary that accepts total test suites count because it contains most of the required information.

I added methods for obtaining the amount of both skipped test cases and skipped test suites.

I put up a separate PR for this as you didn't really respond to my arguments regarding splitting the work on this - #1315. I think we should merge this one and move on with any discussions about these skipped tests to a separate place - the new PR. The current one is very big and's been hanging for a while.

karol-bisztyga: Resolving, for further discussion, please post comments to #1315

kasperski95 (approved):

protostar/commands/test/testing_live_logger.py
karol-bisztyga: OK. #1359, resolving.

protostar/commands/test/messages/passed_fuzz_test_case_result_message.py
```diff
@@ -0,0 +1,113 @@
...
+            to_join: list[str] = [first_line, second_line]
+            if stdout_elements:
+                to_join.append("".join(stdout_elements))
+            return "\n".join(to_join)
+
+        return first_line
+
+    def format_dict(self) -> dict:
+        result = {
+            "message_type": "test_case_result",
```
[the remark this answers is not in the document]
karol-bisztyga: I wouldn't remove the `message_type` field. I think we discussed this already in this PR (if necessary, I can find the discussion in resolved comments). The bottom line is:
- `message_type` distinguishes test outputs from each other, you can have `test_case_result`, `test_collector_result`, and `testing_summary`.
- `test_type` is a field only for `test_case_result` and it can be `broken_test_suite`, `failed_test_case`, etc.

I'm going to juz add `type=test`. I think having a few identifying fields like these is not expensive and can help a lot.

protostar/commands/test/messages/formatters.py
```diff
@@ -0,0 +1,49 @@
...
+    for name, value in captured_stdout.items():
+        if value:
+            result.append(
+                f"[{format_output_name(name)}]:\n{log_color_provider.colorize('GRAY', value)}\n"
+            )
+
+    return result
+
+
+def get_formatted_metadata(metadata: ExceptionMetadata) -> str:
```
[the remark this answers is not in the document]
karol-bisztyga: [get formatter => format](https://github.com/software-mansion/protostar/pull/1254/commits/48f1d927c6eef1bc841415a09f7939f5d38edab5)

protostar/commands/test/messages/passed_fuzz_test_case_result_message.py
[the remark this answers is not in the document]
karol-bisztyga: [add type field](https://github.com/software-mansion/protostar/pull/1254/commits/dd26b46021fb5d27fb62177c4a7a492c067ea6c5)

PR #1279, opened 2022-12-14
Prevent using felt pointer for dict

### LINKED ISSUE #1266: `felt*` validation
> I'm looking to deploy a proxy contract in my tests, however my implementation class is taking a felt* in its initializer.
So it leads to have a felt* (the initializer input) inside a felt* (proxy calldata input).
In protostar I cannot have a list inside a list to describe my inputs, I got the following msg:
calldata[2] should be int
I also tried to flatten my list, but the expected number of inputs do not match anymore.
Is it allowed to initialize an implementation through a proxy initialization where both have a felt* argument?

https://discord.com/channels/793094838509764618/964116575387021413/1049775735788798024

karol-bisztyga:
The point is to prevent people from using `felt*` for passing `dict`s like this:
```
%{ ids.proxy_contract_address = deploy_contract("./ctr.cairo", {"d": {"a": 1, "b": 3}}).contract_address %}
// ctr.cairo
func constructor{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(d_len: felt, d: felt*) {
    %{
        from starkware.cairo.lang.vm.relocatable import RelocatableValue
        rel = RelocatableValue.to_felt_or_relocatable(ids.d)
        values = [memory[rel+i] for i in range(0, ids.d_len)]
        print("length", ids.d_len)
        print("values", values) # that's going to give you only the keys and will cause a data loss
    %}
    return ();
}
```

fix #1266

### DIFF
--- a/tests/integration/data_transformer/data_transformer_test.py
+++ b/tests/integration/data_transformer/data_transformer_test.py
@@ -31,6 +31,24 @@ from tests.integration.data_transformer.contracts import (
     "contract,function_name,mode,data",
     [
         [
+            DATA_TRANSFORMER_FELT_CONSTRUCTOR_CONTRACT,
+            "constructor",
+            "inputs",
+            {"arg": 0xD00D00},
+        ],
+        [
+            DATA_TRANSFORMER_UINT256_CONTRACT,
+            "input_uint256",
+            "inputs",
+            {"arg": 1 << 237},
+        ],
+        [
+            DATA_TRANSFORMER_UINT256_CONTRACT,
+            "input_uint256",
+            "inputs",
+            {"arg": {"high": 101, "low": 102}},
+        ],
+        [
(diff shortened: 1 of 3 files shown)


tests/integration/data_transformer/data_transformer_test.py
```diff
@@ -23,30 +23,13 @@
...
     DATA_TRANSFORMER_STRUCTS_CONTRACT,
     DATA_TRANSFORMER_TUPLE_CONTRACT,
+    DATA_TRANSFORMER_LIST_CONTRACT,
 )
 
 
 @pytest.mark.parametrize(
     "contract,function_name,mode,data",
     [
-        [
```
karol-bisztyga, line 32: not sure why they are removed, may be a mistake, I will check this.

kasperski95 (approved):

PR #1292, opened 2022-12-19
Lint everything

kasperski95:
Check linting for every Python file in the project. Disabled unspecified-encoding rule. Currently, the linting checks only the files in the Protostar directory.

---

> Missing timeout argument for method ‘requests.get’ can cause your program to hang indefinitely

Maybe this is why CI kept freezing?

PR #1295, opened 2022-12-20
Improve git hooks

karol-bisztyga:
I noticed that for pre-commit hooks, all files are processed.
I think we should process only the staged files since they're the ones that are going to be in the commit.
I think we shouldn't check the files that are not going to be in the commit.


kasperski95 (approved):

pre-commit
```diff
@@ -1,5 +1,23 @@
 #!/bin/sh
 
-poetry run poe type_check
-poetry run poe format_check
-poetry run poe lint
+set -e
+
+STAGED_FILES=$(git diff --name-only --cached)
+STAGED_FILES_FOR_FORMAT=$(poetry run poe show_format_check_path | tail -n 1 | xargs git diff --name-only --staged)
+STAGED_FILES_FOR_LINT=$(poetry run poe show_lint_path | tail -n 1 | xargs git diff --name-only --staged)
```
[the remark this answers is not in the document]
karol-bisztyga: Yes, this is correct, I'm going to rebase to that PR and try to figure it out.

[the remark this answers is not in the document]
karol-bisztyga: Resolving, newest changes take care of this

scripts/run_format_check.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+find protostar -name '*.py' | xargs black --check
```
karol-bisztyga, line 3: I couldn't find a way of doing this without a separate script, `pyproject.toml` seems to be a bit retarded when it comes to running bash instructions.

scripts/run_lint.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+find . -name '*.py' | xargs pylint
```
karol-bisztyga, line 3: I couldn't find a way of doing this without a separate script, `pyproject.toml` seems to be a bit retarded when it comes to running bash instructions.

Also, with the changes added in #1292, the CLI output was very polluted by poetry which printed out all affected files (they make a pretty big list).

pre-commit
karol-bisztyga, line 7: oops, this needs fixing :/

karol-bisztyga, line 7: oops, this needs fixing

karol-bisztyga:
I saw some bug, lemme fix that

karol-bisztyga:
I think it can be reviewed now

scripts/get_code_check_paths.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+set -e
+
+if [ "$#" != 1 ]; then
+  echo "Invalid number of arguments, expected 1: lint/format";
+  exit 1;
+fi;
+
+if [ "$1" == "lint" ]; then
+  find . -name '*.py'
```
kasperski95, line 10: I've just run it locally, and the linter complained about files in the virtual environment (.venv). Maybe we can duplicate source of paths in favor of simplicity?

karol-bisztyga: Nice catch. First, I thought we could use paths specified in gitignore to exclude them from the search but it looked too hacky and weird. I think specifying the paths explicitly is simpler and clearer.

karol-bisztyga: [explicitly specify paths to look for python files](https://github.com/software-mansion/protostar/pull/1295/commits/8e6123f43cbf50d807047a164327afc6256b6713)

kasperski95 (approved):

PR #1296, opened 2022-12-20
Fix undeployed account error message

### LINKED ISSUE #1294: Raise custom `Account not deployed` error instead of `Requested contract address is not deployed`
The error is confusing, because users think "contract" refers to the contract being deployed.

https://discord.com/channels/793094838509764618/964116575387021413/1054504335272312852

karol-bisztyga:
fix #1294

### DIFF
--- a/protostar/starknet_gateway/gateway_facade.py
+++ b/protostar/starknet_gateway/gateway_facade.py
@@ -164,6 +164,15 @@ class GatewayFacade(MulticallClientProtocol):
         except TransactionFailedError as ex:
             raise TransactionException(str(ex)) from ex
         except ClientError as ex:
+            account_address_hex = hex(int(account_address))
+            if (
+                ex.code == "500"
+                and f"Requested contract address {account_address_hex} is not deployed"
+                in ex.message
+            ):
+                raise TransactionException(
+                    f"Account {account_address_hex} is not deployed"
+                ) from ex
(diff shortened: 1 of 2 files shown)


protostar/starknet_gateway/gateway_facade.py
```diff
@@ -164,6 +164,15 @@ async def deploy_via_udc(
         except TransactionFailedError as ex:
             raise TransactionException(str(ex)) from ex
         except ClientError as ex:
+            account_address_hex = hex(int(account_address))
```
kasperski95, line 167: You can use `account_address` directly in a string without creating hex representation. `account_address` is an `Account` value object, which has overwritten `__str__` method.

kasperski95 (approved):

protostar/starknet_gateway/gateway_facade.py
karol-bisztyga: The problem is that this `__str__` method prints the hex with all the unnecessary zeros so instead of `0xf50bc325` you have `0x00000000000000000000000000000000000000000000000000000000f50bc325`. It's hard to do pattern matching in such a case.

Anyway, `account_address` was passed as an int although the typing says it's `Address`. Weird that type checker didn't see that.

karol-bisztyga: Please let me know if this can be resolved.

PR #1298, opened 2022-12-20
Split `ProtostarFixture` file

kasperski95:
Make the file smaller.


MaksymilianDemitraszek:
I would restrain from such refactors ATM it may interfere with stuff @Arcticae is working on.

kasperski95:
This change should be unrelated to Cairo test runner.

tests/integration/_conftest/spying_gateway_facade_factory.py
```diff
@@ -0,0 +1,35 @@
...
+from starknet_py.net.gateway_client import GatewayClient
+
+from protostar.compiler.compiled_contract_reader import CompiledContractReader
+from protostar.starknet_gateway import GatewayFacade, GatewayFacadeFactory
+
+from .gateway_client_tx_interceptor import GatewayClientTxInterceptor
+from .transaction_registry import TransactionRegistry
+
+
+class SpyingGatewayFacadeFactory(GatewayFacadeFactory):
```
karol-bisztyga, line 13: Just out of curiosity: where'd you come up with such a name? Why "spying"?

karol-bisztyga (commented):
I think it looks ok, but I'd wait for one more review

tests/integration/_conftest/spying_gateway_facade_factory.py
kasperski95: Because it uses `GatewayClientTxInterceptor` instead of `GatewayClient`. (Maybe `SpyingGatewayClient` would be a better name).  That client puts every transaction into `TransactionRegistry` which can be later used by tests to analyze what is actually sent.

I renamed it from `TestFriendyGatewayFacadeFactory`, because `spy` seems a better word than `test friendly`.

> Spies are stubs that also record some information based on how they were called.
https://martinfowler.com/bliki/TestDouble.html

kasperski95:
@Arcticae @MaksymilianDemitraszek ping

tests/integration/_conftest/transaction_registry.py
```diff
@@ -0,0 +1,16 @@
+from dataclasses import dataclass, field
+
+from starknet_py.net.client_models import InvokeFunction, StarknetTransaction
+from starknet_py.net.models.transaction import DeployAccount
+
+
+@dataclass
+class TransactionRegistry:
```
Arcticae, line 8: Why bother with 2 separate classes doing basically the same thing? (capturing the txns)

tests/integration/_conftest/gateway_client_tx_interceptor.py
```diff
@@ -0,0 +1,26 @@
...
+from starknet_py.net.gateway_client import GatewayClient
+from starknet_py.net.client_models import StarknetTransaction
+from starknet_py.net.networks import Network
+
+from .transaction_registry import TransactionRegistry
+
+
+class GatewayClientTxInterceptor(GatewayClient):
+    def __init__(
+        self, net: Network, transaction_registry: Optional[TransactionRegistry] = None
```
Arcticae, line 12: why make the tx registry optional?

Arcticae (changes requested):
I don't see why we should keep 2 classes which one of them only keeps the txns sorted by the type. I'd rather have a getter method which returns transactions of a certain type.

tests/integration/_conftest/transaction_registry.py
kasperski95: `GatewayClient` is instantiated multiple times (each network command creates a `GatewayClient`). `TransactionRegistry` stores transactions from all gateway clients instantiated within a single test. 

These classes already exist in the code base. This PR splits the `protostar_fixture.py` file.

tests/integration/_conftest/gateway_client_tx_interceptor.py
kasperski95: [remove optional from transaction registry](https://github.com/software-mansion/protostar/pull/1298/commits/311602d4daa3d52eb9d0d0477fe65a6ddb6930e9)

Arcticae (approved):

PR #1301, opened 2022-12-22
Update docusaurus.config.js

sjchmiela:
While sharing from SFWebViewController it shows wrong URL. Funnily enough, not if you share from Safari.

6E90EDB6-94C5-4FB9-BCBB-FC4FC33153E8


kasperski95:
Thanks for the fix.

PR #1302, opened 2022-12-22
Promote OpenZeppelin contract wizard

kasperski95:
Show link to the OZ contract wizard. It should help newcomers to kick-start the project.

### DIFF
--- a/protostar/commands/init/init_command.py
+++ b/protostar/commands/init/init_command.py
@@ -1,5 +1,6 @@
 from argparse import Namespace
 from glob import glob
+from logging import getLogger
 from typing import Optional
 
 from protostar.cli import ProtostarArgument, ProtostarCommand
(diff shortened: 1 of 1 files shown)


Arcticae (changes requested):
I don't understand why we need to do this, promoting 3rd party tools is not responsibility of protostar

kasperski95:
@Arcticae Implicit responsibility of Protostar is minimizing resources (human time) needed to create and maintain contracts on StarkNet. This link cost us nothing and helps push new projects in a right direction.

protostar/commands/init/init_command.py
```diff
@@ -73,12 +75,16 @@ def init(
...
+                    "as a project instead of creating a new project?"
                 )
 
         if should_adapt_existing_project:
             self._adapted_project_creator.run()
         else:
+            logger.info(
+                "Consider using OpenZeppelin Contracts Wizard to kickstart your project.\n"
+                "https://wizard.openzeppelin.com/cairo"
+            )
```
MaksymilianDemitraszek, line 87: We definiately shouldn't include such stuff in logs. We can consider adding it to the docs but I am not sure about this either

MaksymilianDemitraszek (changes requested):

kasperski95:
Outvoted

PR #1304, opened 2022-12-22
Argument tokenizer in ProtostarFixture

kasperski95:
Use tokenizer in `ProtostarFixture` to make this fixture simpler. We shouldn't use Namespace directly because there could be problems with parsing (e.g. list of named arguments provided to `--inputs` argument, which I plan to test/fix in another PR).

### DIFF
--- a/tests/integration/_conftest/tokenizer.py
+++ b/tests/integration/_conftest/tokenizer.py
@@ -1,7 +1,7 @@
 from typing import Any
 
 
-def untokenize(positional_args: list[Any], named_args: dict[str, Any]):
+def tokenize(positional_args: list[Any], named_args: dict[str, Any]):
     tokens: list[str] = []
     for value in positional_args:
         if value is not None:
--- a/tests/integration/_conftest/tokenizer.py
+++ b/tests/integration/_conftest/tokenizer.py
@@ -2,10 +2,7 @@ from typing import Any
 
 
 def tokenize(positional_args: list[Any], named_args: dict[str, Any]):
-    tokens: list[str] = []
-    for value in positional_args:
-        if value is not None:
-            tokens.append(str(value))
+    tokens = [str(item) for item in positional_args if item is not None]
     for name, value in named_args.items():
         if value is None:
             continue
@@ -15,8 +12,7 @@ def tokenize(positional_args: list[Any], named_args: dict[str, Any]):
             continue
         tokens.append(f"--{name}")
         if isinstance(value, list):
-            for val in value:
-                tokens.append(str(val))
+            tokens += [str(item) for item in value]
             continue
         tokens.append(str(value))
     return tokens
--- a/tests/integration/_conftest/tokenizer.py
+++ b/tests/integration/_conftest/tokenizer.py
@@ -6,9 +6,10 @@ def tokenize(positional_args: list[Any], named_args: dict[str, Any]):
     for name, value in named_args.items():
         if value is None:
             continue
-        if isinstance(value, bool):
-            if value:
-                tokens.append(f"--{name}")
+        if value is False:
+            continue
+        if value is True:
+            tokens.append(f"--{name}")
             continue
         tokens.append(f"--{name}")
         if isinstance(value, list):
--- a/tests/integration/_conftest/__init__.py
+++ b/tests/integration/_conftest/__init__.py
@@ -1,0 +1,1 @@
+from .untokenizer import untokenize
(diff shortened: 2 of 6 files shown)


tests/integration/_conftest/tokenizer.py
```diff
@@ -0,0 +1,22 @@
+from typing import Any
+
+
+def tokenize(positional_args: list[Any], named_args: dict[str, Any]):
```
karol-bisztyga, line 4: Why not `args` & `kwargs`?

tests/integration/_conftest/tokenizer.py
```diff
@@ -0,0 +1,22 @@
...
+    tokens: list[str] = []
+    for value in positional_args:
+        if value is not None:
+            tokens.append(str(value))
```
karol-bisztyga, line 8:
```suggestion
    tokens: list[str] = [item for item in positional_args if item is not None]
```

tests/integration/_conftest/tokenizer.py
```diff
@@ -0,0 +1,22 @@
...
+    for name, value in named_args.items():
+        if value is None:
+            continue
+        if isinstance(value, bool):
+            if value:
+                tokens.append(f"--{name}")
+            continue
```
karol-bisztyga, line 15:
```suggestion
        if value == False:
            continue
```

tests/integration/_conftest/tokenizer.py
```diff
@@ -0,0 +1,22 @@
...
+        tokens.append(f"--{name}")
+        if isinstance(value, list):
+            for val in value:
+                tokens.append(str(val))
```
karol-bisztyga, line 19:
```suggestion
            tokens += [str(item) for item in value]
```

tests/integration/_conftest/tokenizer.py
kasperski95: It won't work because if the value is True, the code below will append key and value instead of key only. Argparse expects no values for flags.

tests/integration/_conftest/tokenizer.py
kasperski95: `ProtostarArgument` names use dashes, hence they can't be represented as kwargs in Python.

tests/integration/_conftest/tokenizer.py
kasperski95: [use list comprehensions](https://github.com/software-mansion/protostar/pull/1304/commits/8b6860dc3c00ab1f8eae7ebcbd65d5d02fa84e85)

tests/integration/_conftest/tokenizer.py
kasperski95: [use list comprehensions](https://github.com/software-mansion/protostar/pull/1304/commits/8b6860dc3c00ab1f8eae7ebcbd65d5d02fa84e85)

tests/integration/_conftest/tokenizer.py
karol-bisztyga: Ah, right. Nevertheless, I think you can still skip the `isinstance` thing and just check strictly for `True` to avoid an implicit casting:
```
if value == True:
  tokens.append(f"--{name}")
  continue
```

kasperski95: You need an explicit check for "False" then.

I changed it, as you propose, to avoid nested conditions.

tests/integration/_conftest/tokenizer.py
karol-bisztyga: You can use dict:

```
def f(*args, **kwargs):
  print('args', type(args))
  for a in args:
    print(type(a), '/', a)
  print('kwargs', type(kwargs))
  for k, v in kwargs.items():
    print(type(k), type(v), '', k, '=>', v)

f(1, 2, 3, **{'a-a': 4, 'b': 5})
```

kasperski95: If somebody starts using kwargs in this way:
```
f(foo=1, bar=2, baz=3, ...)
```
And later he needs to add kwarg with dash in its name, he needs to rewrite all arguments. Hence, it's better to require dict.

This shouldn't be a problem with args.

karol-bisztyga: Fair point

karol-bisztyga (approved):

MaksymilianDemitraszek (approved):
Seems a bit of a overkill but we can merge it

PR #1321, opened 2023-01-02
Fix pre commit hook

karol-bisztyga:
I encountered some problems with the hooks and this is the fix. More details in single comments.

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -10,10 +10,14 @@ if [ -z "${STAGED_FILES}" ]; then
   exit 0;
 fi
 
-poetry run poe type_check "${STAGED_FILES}"
+poetry run poe type_check $(echo "${STAGED_FILES}")
+
+STAGED_FILES_SELECTED_ARR=($STAGED_FILES_SELECTED)
 
 if [ -n "${STAGED_FILES_SELECTED}" ]; then
   echo "${STAGED_FILES_SELECTED}" | xargs poetry run poe format_check_selected
-  echo "${STAGED_FILES_SELECTED}" | xargs poetry run poe lint_selected
+  for FILE in "${STAGED_FILES_SELECTED_ARR[@]}"; do
+    echo "${FILE}" | xargs poetry run poe lint_selected
+  done;
 fi
 


pre-commit
```diff
@@ -10,10 +10,14 @@ if [ -z "${STAGED_FILES}" ]; then
...
+poetry run poe type_check $(echo "${STAGED_FILES}")
+
+STAGED_FILES_SELECTED_ARR=($STAGED_FILES_SELECTED)
 
 if [ -n "${STAGED_FILES_SELECTED}" ]; then
   echo "${STAGED_FILES_SELECTED}" | xargs poetry run poe format_check_selected
-  echo "${STAGED_FILES_SELECTED}" | xargs poetry run poe lint_selected
+  for FILE in "${STAGED_FILES_SELECTED_ARR[@]}"; do
+    echo "${FILE}" | xargs poetry run poe lint_selected
+  done;
```
karol-bisztyga, line 21: Unfortunately, bash+pylint have been really mean and I decided that taking care of every single line separately is the only reasonable way to go. Pylint doesn't seem to accept a list of files.

pre-commit
```diff
@@ -10,10 +10,14 @@ if [ -z "${STAGED_FILES}" ]; then
   exit 0;
 fi
 
-poetry run poe type_check "${STAGED_FILES}"
+poetry run poe type_check $(echo "${STAGED_FILES}")
+
+STAGED_FILES_SELECTED_ARR=($STAGED_FILES_SELECTED)
```
karol-bisztyga, line 15: This should be defined in the condition below

pre-commit
```diff
@@ -10,10 +10,14 @@ if [ -z "${STAGED_FILES}" ]; then
   exit 0;
 fi
 
-poetry run poe type_check "${STAGED_FILES}"
+poetry run poe type_check $(echo "${STAGED_FILES}")
```
karol-bisztyga, line 13: This didn't work properly as well, I didn't notice the error message in the cluttered output. It was saying something like this:
```
File 'path1
path2
path3' not found
```
So it cannot handle the `\n`-separated file list, it has to be space-split (which is done by using `echo`)

MaksymilianDemitraszek (approved):

PR #1325, opened 2023-01-03
fix(install): support ash shell

remiroyc:
Docker Alpine images come with ash as the default shell instead of bash.

The support of the ash shell would allow using docker containers such as ```python:3.9-alpine``` allowing a quick setup for the Cairo env.


Radinyn (approved):

Radinyn:
Thanks for you contribution!

PR #1326, opened 2023-01-04
Fixed exit first flag

Radinyn:


protostar/testing/test_results.py
```diff
@@ -161,3 +161,12 @@ class BrokenTestSuiteResult(TestResult):
 @dataclass(frozen=True)
 class UnexpectedBrokenTestSuiteResult(BrokenTestSuiteResult):
     traceback: Optional[str]
+
+
+NONNEGATIVE_RESULTS = (
```
karol-bisztyga, line 166: Why do you use a caps-locked name for this? Also, I'd include "test", maybe something like this: `NonnegativeTestResults`?

protostar/commands/test/testing_live_logger.py
```diff
@@ -81,9 +82,10 @@ def log(
                         if (
                             self.exit_first
                             and shared_tests_state.any_failed_or_broken()
+                            and not isinstance(test_result, NONNEGATIVE_RESULTS)
                         ):
                             tests_left_n = 0
-                            return
+                            continue
```
karol-bisztyga, line 88: Is this change correct? Shouldn't we break when we encounter a broken/failed test result?

protostar/commands/test/testing_live_logger.py
```diff
@@ -81,9 +82,10 @@ def log(
                         if (
                             self.exit_first
                             and shared_tests_state.any_failed_or_broken()
+                            and not isinstance(test_result, NONNEGATIVE_RESULTS)
```
karol-bisztyga, line 85: Is this necessary? Isn't this already taken care of in [protostar/testing/test_shared_tests_state.py](https://github.com/software-mansion/protostar/pull/1326/files#diff-ced3a88b3f75b799756c02bea11369baf57085419b57fe9ceade5bd6d6060229)'s `put_result`?

protostar/commands/test/testing_live_logger.py
MaksymilianDemitraszek: I believe we set test_left to 0 so in this case those would be equivalent? break may slightly better explain the intention but imo but this is a [[Nitpick]]

protostar/commands/test/testing_live_logger.py
MaksymilianDemitraszek: I believe this is a reason of the issue? Because variable may be set to true but the result may be on the queue? is it @Radinyn ?

My question here is, why we need the any_failed_or_broken variable in the first place? Wouldn't be better to keep the control in the parent process here and just terminate other processes when first fail is reported? Every piece of shared state removed is a benefit for the codebase 🏗️

Radinyn: Last time we were working at this code we couldn't find a way to do this without delegating some more logic to the logger, I think that removing this shared flag isn't as important as preventing this.

protostar/testing/test_shared_tests_state.py
```diff
@@ -21,7 +21,7 @@ def get_result(self) -> TestResult:
         return self._shared_queue.get(block=True, timeout=20000)
 
     def put_result(self, item: TestResult) -> None:
-        if not isinstance(item, PassedTestCaseResult):
+        if not isinstance(item, NONNEGATIVE_RESULTS):
```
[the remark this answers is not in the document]
Radinyn: There are also some unexpected error test cases and I would like for them to be considered failed as well. I think it's a matter of naming NONNEGATIVE_RESULTS better.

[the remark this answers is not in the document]
MaksymilianDemitraszek: Wouldn't it be more readable to:
```
 isinstance(test_result, FAILED_RESULT)
```
idk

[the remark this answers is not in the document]
Radinyn: @kasperski95 came up with "AcceptableResult"

protostar/testing/test_results.py
```diff
@@ -9,6 +9,13 @@
 from .test_output_recorder import OutputName
 
 
+@dataclass(frozen=True)
+class AcceptableResult:
```
karol-bisztyga, line 13: Nit: Maybe more clear would be `AcceptableTestCaseResult`? Just `AcceptableResult` does not seem to indicate what result we mean.

protostar/commands/test/testing_live_logger.py
```diff
@@ -124,16 +121,22 @@ def log_human(
                             progress_bar.update(1)
                             tests_left_n -= 1
                 finally:
-                    progress_bar.write("")
-                    progress_bar.clear()
+                    progress_bar.close()
```
karol-bisztyga, line 124: So `progress_bar.close()` == `progress_bar.write("")` + `progress_bar.clear()`? Or is the behavior different?
Asking out of curiosity.

karol-bisztyga (approved):

protostar/testing/test_results.py
Radinyn: There are also setup cases, test suites and some unexpected errors, that's why `AcceptableTestCaseResult` isn't general enough.

protostar/commands/test/testing_live_logger.py
Radinyn: This way we guarantee that the progress bar actually disappears without filling completely, that wasn't the case before.

Issue #1327, opened 2023-01-04
Enable passing arrays with named args


karol-bisztyga:
When invoking a `call` method, we can do something like this:
```
protostar call --contract-address 0x123 --function sum_threee_numbers --inputs 1 2 3
```
as well as something like this:
```
protostar call --contract-address 0x123 --function sum_threee_numbers --inputs a=1 b=2 c=3
```
But when it comes to arrays, the named arguments cannot be used. You can do something like this:
```
protostar call --contract-address 0x123 --function sum_array --inputs 3 4 1 5
```
where the first arg is the length of the arrays and the rest of the args are the item of the array.

Something like this, however, will not work:
```
protostar call --contract-address 0x123 --function sum_array --inputs arr_len=3 arr=4 2 6
```
The goal is to enable such behavior in protostar.

PR #1332, opened 2023-01-04
Allow passing arrays via named args from the CLI

karol-bisztyga:
In response to #1327.

This is an early POC.

The idea is to always convert the inputs to a list. If they're passed in the named manner (like `a=1 b=2 c=3`), then, relying on the order of the inputs from the abi, they're converted to the list `1 2 3` (so if you pass `a=1 c=3 b=2` you'll still get `1 2 3`, because in the abi you'll have a correct order of args).

I realize this code's not perfect and there's some refactoring ahead but I'm putting this up to just present the idea for this.

So, to pass an array via named args, you'd do something like this:
```
protostar call --contract-address 0x123 --function sum_array --inputs arr_len=3 arr=4 5 4
```

### DIFF
--- a/protostar/argument_parser/abi_obtainer.py
+++ b/protostar/argument_parser/abi_obtainer.py
@@ -1,0 +1,17 @@
+from typing import Any
+
+from protostar.cli import NetworkCommandUtil
+from protostar.starknet_gateway import (
+    AbiResolver,
+    DataTransformerPolicy,
+)
+
+
+async def abi_from_args(args: Any):
+    network_command_util = NetworkCommandUtil(args)
+    gateway_client = network_command_util.get_gateway_client()
+    abi_resolver = AbiResolver(client=gateway_client)
+    data_transformer_policy = DataTransformerPolicy(abi_resolver=abi_resolver)
+    return await data_transformer_policy.resolve_abi_or_fail(
+        address=args.contract_address
+    )
--- a/protostar/argument_parser/argument_parser_facade.py
+++ b/protostar/argument_parser/argument_parser_facade.py
@@ -27,20 +27,29 @@ ArgTypeNameT_contra = TypeVar(
 
 
 def parse_collection_arg(arg: list[Union[int, dict[str, Any]]]):
-    arg_type: Optional[type] = None
+    dict_present = False
     result = arg
     for input_item in result:
-        if arg_type is None:
-            arg_type = type(input_item)
-            continue
-        if not isinstance(input_item, arg_type):
-            raise InconsistentInputTypesException()
+        if isinstance(input_item, dict):
+            dict_present = True
+            break
+    if not isinstance(arg[0], dict) and dict_present:
+        raise InconsistentInputTypesException()
 
-    if arg_type == dict:
+    if isinstance(arg[0], dict):
         parsed = {}
+        last_key = None
         for arg_item in result:
-            assert isinstance(arg_item, dict)
-            parsed.update(dict(arg_item))
+            if isinstance(arg_item, dict):
+                keys = list(arg_item.keys())
+                assert len(keys) == 1
+                last_key = keys[0]
+                parsed.update(dict(arg_item))
+            else:
+                if isinstance(parsed[last_key], list):
+                    parsed[last_key].append(arg_item)
+                else:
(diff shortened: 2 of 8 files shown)


protostar/cli/network_command_util.py
```diff
@@ -93,6 +93,6 @@ def _normalize_chain_id(
             return StarknetChainId(arg)
         except ValueError as ex:
             raise ProtostarException(
-                "Invalid chain id value.\n"
+                f"Invalid chain id value: {arg}.\n"
```
karol-bisztyga, line 96: A bit unrelated but I decided we could maybe squash it in here. No big deal anyhow, I think it's good to show what the invalid value is, it makes debugging easier/faster.

protostar/argument_parser/argument_parser_facade.py
```diff
@@ -27,20 +27,32 @@
...
     for input_item in result:
-        if arg_type is None:
-            arg_type = type(input_item)
-            continue
-        if not isinstance(input_item, arg_type):
-            raise InconsistentInputTypesException()
+        if isinstance(input_item, dict):
+            dict_present = True
+            break
+    if not isinstance(arg[0], dict) and dict_present:
```
karol-bisztyga, line 39: So if we have a named arg present (represented by a `dict`), the args in general have to start with a named arg. Then, we can have just regular args (that would be interpreted as a list) but they cannot start with a list.

protostar/start.py
```diff
@@ -37,8 +38,28 @@ def parse_args(parser: ArgumentParserFacade) -> Any:
...
 
 
-def run_protostar(protostar_cli: ProtostarCLI, parser: ArgumentParserFacade, args: Any):
+def run_protostar(
+    protostar_cli: ProtostarCLI,
+    parser: ArgumentParserFacade,
+    args: Any,
+):
     try:
+        if hasattr(args, "inputs") and isinstance(args.inputs, dict):
```
karol-bisztyga, line 47: replace `hasattr`

protostar/start.py
```diff
@@ -37,8 +38,28 @@ def parse_args(parser: ArgumentParserFacade) -> Any:
...
+                    break
+            assert target_abi_item
+            inputs_as_list = []
+            for input_item in target_abi_item["inputs"]:
+                new_item = args.inputs[input_item["name"]]
+                if isinstance(new_item, list):
+                    inputs_as_list += new_item
+                else:
+                    inputs_as_list.append(new_item)
+            args.inputs = inputs_as_list
```
karol-bisztyga, line 62: move this to a separate function

karol-bisztyga (commented):
- [ ] docs + examples
- [x] tests

protostar/argument_parser/abi_obtainer.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+async def abi_from_args(args: Any):
+    network_command_util = NetworkCommandUtil(args)
+    gateway_client = network_command_util.get_gateway_client()
+    abi_resolver = AbiResolver(client=gateway_client)
+    data_transformer_policy = DataTransformerPolicy(abi_resolver=abi_resolver)
+    return await data_transformer_policy.resolve_abi_or_fail(
+        address=args.contract_address
+    )
```
kasperski95, line 17: `ArgumentParser` component should be unaware of Protostar. It should be actually moved out of the Protostar directory. `protostar.cli` would be a better place for this function.

protostar/start.py
kasperski95, line 62: This code definitely shouldn't be in the `start.py`.

kasperski95 (commented):
The behavior looks ok, but the problem will be with the design. This code should be mainly in `protostar.cli` or in `protostar.commands`.

protostar/start.py
karol-bisztyga: > I realize this code's not perfect and there's some refactoring ahead but I'm putting this up to just present the idea for this.

karol-bisztyga:
I think we can close this one. The code here is a bit edgy and the feature is not urgent as it can be easily done with regular args. I'm going to leave the issue as is so we can always reopen this one or use some parts of it.

PR #1328, opened 2023-01-04
deploy_command: tx hash log changed to hex

benceharomi:
resolves #1324

PR #1333, opened 2023-01-05
Print transaction hash in hex

karol-bisztyga:


karol-bisztyga:
already done in #1328

PR #1355, opened 2023-01-09
Refactor Cairo cheatcodes

kasperski95:


Arcticae:
It's making impossible to access cheaters in the cheatcodes, which we'll want to be a main interface of interaction with the test state

kasperski95:
> It's making impossible to access cheaters in the cheatcodes 

WDYM? Cheatcodes have access to the cheaters through the `ForkableStarknet`.

> we'll want to be a main interface of interaction with the test state

I don't see why `CairoCheatcode` should be responsible for that.

Arcticae:
Then let's pass over the cheaters to each cheatcode instance instead. It's rather a cosmetic change though i suppose.

MaksymilianDemitraszek:
I think is irrevelant at the current stage. We have so many open questions about how to execute the integration that  discussing such minor details is a waste of time.

kasperski95:
Outvoted.
IMO a design flaw rather than cosmetic change. The problem grows with every new CairoCheatcode.

Issue #1361, opened 2023-01-10
Add `call` cheatcodes tests through proxy


MaksymilianDemitraszek:
Historically there were issues including this scenario
https://www.cairo-lang.org/docs/hello_starknet/default_entrypoint.html

Examples of such test are in
`tests/integration/cheatcodes/mock_call`

PR #1363, opened 2023-01-11
Add expect call cairo cheatcode

karol-bisztyga:
TODO tests

<s>
First, it will be probably better to complete #1361 

- [ ] Expect call through proxy
- [ ] Expect call in a remote contract (with the depth of call 2) -> expect call on contract C and test calls B which calls C
- [ ] expect call with selector instead of str
</s>

The tasks above have been delegated to #1383

### DIFF
--- a/protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
+++ b/protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
@@ -15,6 +15,12 @@ from protostar.cairo_testing.cheatcode_factories.cairo_test_cheatcode_factory im
 )
 from protostar.cairo import HintLocalsDict
 
+from protostar.cheatable_starknet.cheatable_cached_state import CheatableCachedState
+from protostar.cheatable_starknet.cheaters.expects import ExpectsCairoCheater
+from protostar.protostar_exception import ProtostarException
+from protostar.starknet.cheater import CheaterException
+from protostar.starknet import ReportedException
+
 from .cairo_execution_environment import CairoExecutionEnvironment
 
 
@@ -32,8 +38,18 @@ class CairoTestExecutionEnvironment(CairoExecutionEnvironment):
 
     async def execute(self, function_name: str) -> Any:
         with self.state.output_recorder.redirect("test"):
-            await self.execute_test_case(function_name)
-            return TestExecutionResult(execution_resources=None)
+            result = TestExecutionResult(
+                execution_resources=await self.execute_test_case(function_name)
+            )
+            try:
+                cheatable_state = self.state.starknet.state.state
+                assert isinstance(cheatable_state, CheatableCachedState)
+                ExpectsCairoCheater.assert_no_expected_calls(
+                    cheatable_state.expected_contract_calls
+                )
+            except (ProtostarException, CheaterException) as e:
+                raise ReportedException from e
+            return result
 
     # TODO #1303: Estimate gas if self.state.config.gas_estimation_enabled
     async def execute_test_case(
--- a/protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
+++ b/protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
@@ -18,7 +18,7 @@ from protostar.cairo import HintLocalsDict
 from protostar.cheatable_starknet.cheatable_cached_state import CheatableCachedState
 from protostar.cheatable_starknet.cheaters.expects import ExpectsCairoCheater
 from protostar.protostar_exception import ProtostarException
-from protostar.starknet.cheater import CheaterException
+from protostar.cheatable_starknet.cheaters.expects import ExpectsCheaterException
 from protostar.starknet import ReportedException
 
 from .cairo_execution_environment import CairoExecutionEnvironment
@@ -47,7 +47,7 @@ class CairoTestExecutionEnvironment(CairoExecutionEnvironment):
                 ExpectsCairoCheater.assert_no_expected_calls(
                     cheatable_state.expected_contract_calls
                 )
-            except (ProtostarException, CheaterException) as e:
+            except (ProtostarException, ExpectsCheaterException) as e:
                 raise ReportedException from e
             return result
 
--- a/protostar/cheatable_starknet/cheatcodes/expect_call_cairo_cheatcode.py
+++ b/protostar/cheatable_starknet/cheatcodes/expect_call_cairo_cheatcode.py
@@ -1,0 +1,48 @@
+import asyncio
+from typing import Callable, Optional
+
+from protostar.starknet import CheatcodeException
+from protostar.starknet import RawAddress, Address
+from protostar.starknet.cheater import CheaterException
+from protostar.starknet.data_transformer import CairoOrPythonData
+from protostar.cheatable_starknet.cheatcodes.cairo_cheatcode import CairoCheatcode
+
+
+class ExpectCallCairoCheatcode(CairoCheatcode):
+    @property
+    def name(self) -> str:
+        return "expect_call"
+
+    def _build(
+        self,
+    ) -> Callable[[RawAddress, str, Optional[CairoOrPythonData]], Callable]:
+        return self.expect_call
+
+    def expect_call(
+        self,
+        contract_address: RawAddress,
+        function_name: str,
+        calldata: Optional[CairoOrPythonData] = None,
+    ) -> Callable:
+        return asyncio.run(
+            self._expect_call(
+                contract_address=Address.from_user_input(contract_address),
+                function_name=function_name,
+                calldata=calldata,
+            )
+        )
+
+    async def _expect_call(
+        self,
+        contract_address: Address,
+        function_name: str,
+        calldata: Optional[CairoOrPythonData] = None,
+    ) -> Callable:
+        try:
+            return await self.cheaters.expects.expect_call(
+                contract_address=contract_address,
+                function_name=function_name,
+                calldata=calldata,
+            )
+        except CheaterException as exc:
+            raise CheatcodeException(self, exc.message) from exc
--- a/protostar/cheatable_starknet/cheatcodes/expect_call_cairo_cheatcode.py
+++ b/protostar/cheatable_starknet/cheatcodes/expect_call_cairo_cheatcode.py
@@ -3,9 +3,9 @@ from typing import Callable, Optional
 
 from protostar.starknet import CheatcodeException
 from protostar.starknet import RawAddress, Address
-from protostar.starknet.cheater import CheaterException
 from protostar.starknet.data_transformer import CairoOrPythonData
 from protostar.cheatable_starknet.cheatcodes.cairo_cheatcode import CairoCheatcode
+from protostar.cheatable_starknet.cheaters.expects import ExpectsCheaterException
 
 
 class ExpectCallCairoCheatcode(CairoCheatcode):
@@ -44,5 +44,5 @@ class ExpectCallCairoCheatcode(CairoCheatcode):
                 function_name=function_name,
                 calldata=calldata,
             )
-        except CheaterException as exc:
+        except ExpectsCheaterException as exc:
             raise CheatcodeException(self, exc.message) from exc
--- a/protostar/cheatable_starknet/cheaters/contracts.py
+++ b/protostar/cheatable_starknet/cheaters/contracts.py
@@ -1,4 +1,3 @@
-import collections
 import copy
 from pathlib import Path
 from typing import List, Optional, TYPE_CHECKING
@@ -34,12 +33,12 @@ from protostar.cheatable_starknet.cheatable_entry_point import (
     CheatableExecuteEntryPoint,
 )
 from protostar.starknet.types import ClassHashType
+from protostar.starknet.cheater import CheaterException
 from protostar.starknet.address import Address
 from protostar.starknet.data_transformer import (
-    DataTransformerException,
     CairoOrPythonData,
     CairoData,
-    from_python_transformer,
+    transform_calldata_to_cairo_data,
 )
 
 if TYPE_CHECKING:
@@ -48,10 +47,8 @@ if TYPE_CHECKING:
     from . import CairoCheaters
 
 
-class ContractsCheaterException(Exception):
-    def __init__(self, message: str):
-        super().__init__(message)
-        self.message = message
+class ContractsCheaterException(CheaterException):
+    pass
 
 
 class ConstructorInputTransformationException(ContractsCheaterException):
@@ -65,44 +62,6 @@ class ConstructorInvocationException(ContractsCheaterException):
 class ContractsCairoCheater:
     def __init__(self, cheatable_state: "CheatableCachedState"):
         self.cheatable_state = cheatable_state
-
-    async def _transform_calldata_to_cairo_data_by_addr(
-        self,
-        contract_address: Address,
-        function_name: str,
-        calldata: Optional[CairoOrPythonData] = None,
-    ) -> CairoData:
-        contract_address_int = int(contract_address)
-        class_hash = await self.cheatable_state.get_class_hash_at(contract_address_int)
-        return await self._transform_calldata_to_cairo_data(
-            class_hash=from_bytes(class_hash, "big"),
-            function_name=function_name,
-            calldata=calldata,
-        )
-
-    async def _transform_calldata_to_cairo_data(
-        self,
-        class_hash: int,
-        function_name: str,
-        calldata: Optional[CairoOrPythonData] = None,
-    ) -> CairoData:
-        if isinstance(calldata, collections.Mapping):
-            contract_class = await self.cheatable_state.get_contract_class(
-                class_hash=to_bytes(class_hash, 32, "big")
-            )
-            assert contract_class.abi, f"No abi found for the contract at {class_hash}"
-
-            transformer = from_python_transformer(
-                contract_class.abi, function_name, "inputs"
-            )
-            try:
-                return transformer(calldata)
-            except DataTransformerException as dt_exc:
-                raise ConstructorInputTransformationException(
-                    f"There was an error while parsing the arguments for the function {function_name}:\n"
-                    + f"{dt_exc.message}",
-                ) from dt_exc
-        return calldata or []
 
     async def declare_contract(
         self,
@@ -182,8 +141,10 @@ class ContractsCairoCheater:
         return DeployedContract(contract_address=prepared.contract_address)
 
     async def invoke_constructor(self, prepared: PreparedContract):
-        await self._transform_calldata_to_cairo_data(
-            class_hash=prepared.class_hash,
+        await transform_calldata_to_cairo_data(
+            contract_class=await self.cheatable_state.get_contract_class_by_address(
+                contract_address=prepared.contract_address
+            ),
             function_name="constructor",
             calldata=prepared.constructor_calldata,
         )
@@ -218,8 +179,10 @@ class ContractsCairoCheater:
         constructor_calldata: CairoOrPythonData,
         salt: int,
     ) -> PreparedContract:
-        constructor_calldata = await self._transform_calldata_to_cairo_data(
-            class_hash=declared.class_hash,
+        constructor_calldata = await transform_calldata_to_cairo_data(
+            contract_class=await self.cheatable_state.get_contract_class(
+                class_hash=to_bytes(declared.class_hash)
+            ),
             function_name="constructor",
             calldata=constructor_calldata,
         )
@@ -250,8 +213,10 @@ class ContractsCairoCheater:
         calldata: Optional[CairoOrPythonData] = None,
     ) -> CairoData:
         contract_address_int = int(contract_address)
-        cairo_calldata = await self._transform_calldata_to_cairo_data_by_addr(
-            contract_address=contract_address,
+        cairo_calldata = await transform_calldata_to_cairo_data(
+            contract_class=await self.cheatable_state.get_contract_class_by_address(
+                contract_address=contract_address_int
+            ),
             function_name=function_name,
             calldata=calldata,
         )
@@ -276,8 +241,10 @@ class ContractsCairoCheater:
     ):
         contract_address_int = int(contract_address)
 
-        cairo_calldata = await self._transform_calldata_to_cairo_data_by_addr(
-            contract_address=contract_address,
+        cairo_calldata = await transform_calldata_to_cairo_data(
+            contract_class=await self.cheatable_state.get_contract_class_by_address(
+                contract_address=contract_address_int
+            ),
             function_name=function_name,
             calldata=calldata,
         )
--- a/protostar/cheatable_starknet/cheaters/contracts.py
+++ b/protostar/cheatable_starknet/cheaters/contracts.py
@@ -39,6 +39,7 @@ from protostar.starknet.data_transformer import (
     CairoOrPythonData,
     CairoData,
     transform_calldata_to_cairo_data,
+    DataTransformerException,
 )
 
 if TYPE_CHECKING:
@@ -141,13 +142,16 @@ class ContractsCairoCheater:
         return DeployedContract(contract_address=prepared.contract_address)
 
     async def invoke_constructor(self, prepared: PreparedContract):
-        await transform_calldata_to_cairo_data(
-            contract_class=await self.cheatable_state.get_contract_class_by_address(
-                contract_address=prepared.contract_address
-            ),
-            function_name="constructor",
-            calldata=prepared.constructor_calldata,
-        )
+        try:
+            await transform_calldata_to_cairo_data(
+                contract_class=await self.cheatable_state.get_contract_class_by_address(
+                    contract_address=prepared.contract_address
+                ),
+                function_name="constructor",
+                calldata=prepared.constructor_calldata,
+            )
+        except DataTransformerException as e:
+            raise ContractsCheaterException(e.message) from e
         await self.execute_constructor_entry_point(
             class_hash_bytes=to_bytes(prepared.class_hash),
             constructor_calldata=prepared.constructor_calldata,
@@ -179,13 +183,16 @@ class ContractsCairoCheater:
         constructor_calldata: CairoOrPythonData,
         salt: int,
     ) -> PreparedContract:
-        constructor_calldata = await transform_calldata_to_cairo_data(
-            contract_class=await self.cheatable_state.get_contract_class(
-                class_hash=to_bytes(declared.class_hash)
-            ),
-            function_name="constructor",
-            calldata=constructor_calldata,
-        )
+        try:
+            constructor_calldata = await transform_calldata_to_cairo_data(
+                contract_class=await self.cheatable_state.get_contract_class(
+                    class_hash=to_bytes(declared.class_hash)
+                ),
+                function_name="constructor",
+                calldata=constructor_calldata,
+            )
+        except DataTransformerException as e:
+            raise ContractsCheaterException(e.message) from e
 
         contract_address = calculate_contract_address_from_hash(
             salt=salt,
@@ -213,13 +220,16 @@ class ContractsCairoCheater:
         calldata: Optional[CairoOrPythonData] = None,
     ) -> CairoData:
         contract_address_int = int(contract_address)
-        cairo_calldata = await transform_calldata_to_cairo_data(
-            contract_class=await self.cheatable_state.get_contract_class_by_address(
-                contract_address=contract_address_int
-            ),
-            function_name=function_name,
-            calldata=calldata,
-        )
+        try:
+            cairo_calldata = await transform_calldata_to_cairo_data(
+                contract_class=await self.cheatable_state.get_contract_class_by_address(
+                    contract_address=contract_address_int
+                ),
+                function_name=function_name,
+                calldata=calldata,
+            )
+        except DataTransformerException as e:
+            raise ContractsCheaterException(e.message) from e
         entry_point = CheatableExecuteEntryPoint.create_with_cheaters(
             contract_address=contract_address_int,
             calldata=cairo_calldata,
@@ -240,14 +250,16 @@ class ContractsCairoCheater:
         calldata: Optional[CairoOrPythonData] = None,
     ):
         contract_address_int = int(contract_address)
-
-        cairo_calldata = await transform_calldata_to_cairo_data(
-            contract_class=await self.cheatable_state.get_contract_class_by_address(
-                contract_address=contract_address_int
-            ),
-            function_name=function_name,
-            calldata=calldata,
-        )
+        try:
+            cairo_calldata = await transform_calldata_to_cairo_data(
+                contract_class=await self.cheatable_state.get_contract_class_by_address(
+                    contract_address=contract_address_int
+                ),
+                function_name=function_name,
+                calldata=calldata,
+            )
+        except DataTransformerException as e:
+            raise ContractsCheaterException(e.message) from e
         entry_point = CheatableExecuteEntryPoint.create_with_cheaters(
             contract_address=contract_address_int,
             calldata=cairo_calldata,
--- a/protostar/cheatable_starknet/cheaters/expects.py
+++ b/protostar/cheatable_starknet/cheaters/expects.py
@@ -3,6 +3,7 @@ from typing import Optional, Callable, TYPE_CHECKING
 
 from starkware.starknet.public.abi import get_selector_from_name
 
+from protostar.starknet import CheatcodeException
 from protostar.starknet.cheater import CheaterException
 from protostar.starknet.address import Address
 from protostar.starknet.data_transformer import (
@@ -15,6 +16,10 @@ from protostar.starknet.types import SelectorType
 
 if TYPE_CHECKING:
     from protostar.cheatable_starknet.cheatable_cached_state import CheatableCachedState
+
+
+class ExpectsCheaterException(CheaterException):
+    pass
 
 
 class ExpectsCairoCheater:
@@ -45,10 +50,13 @@ class ExpectsCairoCheater:
         )
 
         def stop_callback():
-            ExpectsCairoCheater.assert_no_expected_calls(
-                expected_calls=self.cheatable_state.expected_contract_calls,
-                fixed_address=contract_address,
-            )
+            try:
+                ExpectsCairoCheater.assert_no_expected_calls(
+                    expected_calls=self.cheatable_state.expected_contract_calls,
+                    fixed_address=contract_address,
+                )
+            except ExpectsCheaterException as e:
+                raise CheatcodeException("expect_call", e.message) from e
 
         return stop_callback
 
@@ -101,4 +109,4 @@ class ExpectsCairoCheater:
             for details_item in details:
                 msg += f"{os.linesep}    - function name: {details_item[0]}, calldata: {details_item[1]}"
         if any_addresses_present:
-            raise CheaterException(msg)
+            raise ExpectsCheaterException(msg)
--- a/protostar/cheatable_starknet/cheaters/expects.py
+++ b/protostar/cheatable_starknet/cheaters/expects.py
@@ -1,0 +1,104 @@
+import os
+from typing import Optional, Callable, TYPE_CHECKING
+
+from starkware.starknet.public.abi import get_selector_from_name
+
+from protostar.starknet.cheater import CheaterException
+from protostar.starknet.address import Address
+from protostar.starknet.data_transformer import (
+    CairoOrPythonData,
+    transform_calldata_to_cairo_data,
+    CairoData,
+)
+from protostar.starknet.types import SelectorType
+
+
+if TYPE_CHECKING:
+    from protostar.cheatable_starknet.cheatable_cached_state import CheatableCachedState
+
+
+class ExpectsCairoCheater:
+    def __init__(self, cheatable_state: "CheatableCachedState"):
+        self.cheatable_state = cheatable_state
+
+    async def expect_call(
+        self,
+        contract_address: Address,
+        function_name: str,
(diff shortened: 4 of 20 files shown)


protostar/starknet/cheatable_cached_state.py
```diff
@@ -15,10 +15,12 @@
 
 from protostar.starknet.cheaters import BlockInfoCheater, Cheaters, ContractsCheater
 from protostar.starknet.types import ClassHashType, SelectorType
-from protostar.starknet.data_transformer import CairoOrPythonData
+from protostar.starknet.data_transformer import CairoOrPythonData, CairoData
 
 from .address import Address
 
+ExpectedCallsData = list[tuple[str, CairoOrPythonData]]
```
karol-bisztyga, line 22: probably redundant type

protostar/starknet/cheatable_cached_state.py
```diff
@@ -174,6 +177,32 @@ def unregister_expected_call(
             if not data_for_address:
                 del self.expected_contract_calls[contract_address]
 
+    def register_expected_call_cairo(
```
karol-bisztyga, line 180: For now, I decided to keep both versions of functions for `expect_call` - the original ones and the new ones for the new arch.

protostar/starknet/cheaters/utils.py
```diff
@@ -0,0 +1,24 @@
+import os
+from typing import Optional
+
+from protostar.protostar_exception import ProtostarException
+from protostar.starknet.address import Address
+from protostar.starknet.data_transformer import CairoOrPythonData
+
+
+def assert_no_expected_calls(
```
karol-bisztyga, line 9: Not sure if this place is right for this function. I tried a couple of different places but I was getting a lot of complaints about circular dependencies.

protostar/testing/cairo_test_runner.py
```diff
@@ -199,9 +201,11 @@ async def _invoke_test_case(
             program=program,
             state=state,
         )
-        return await StandardTestCaseRunner(
+        runner = StandardTestCaseRunner(
```
karol-bisztyga, line 204: changes in this file should be reverted 🥶

karol-bisztyga: [revert redundant changes](https://github.com/software-mansion/protostar/pull/1363/commits/20ff0c6ba52a828ef2d747e147158aa04c94db4c)

protostar/starknet/cheatable_cached_state.py
karol-bisztyga: [revert redundant changes](https://github.com/software-mansion/protostar/pull/1363/commits/20ff0c6ba52a828ef2d747e147158aa04c94db4c)

protostar/starknet/cheatable_cached_state.py
karol-bisztyga: **TODO** Worth re-checking if there's a way to use the same structure (and registering functions) for both

tests/integration/pure_cairo_vm/cheatcodes/expect_call/expect_call_cairo_cheatcode_test.py
```diff
@@ -0,0 +1,47 @@
...
+TEST_PATH = Path(__file__).parent
+
+
+@pytest.fixture(autouse=True, name="protostar")
+def protostar_fixture(create_protostar_project: CreateProtostarProjectFixture):
+    with create_protostar_project() as protostar:
+        yield protostar
+
+
+async def test_call_pipeline(
```
karol-bisztyga, line 22: fix name

protostar/starknet/cheaters/utils.py
Arcticae: oh we don't like utils modules around here :P let's name this file something else.

karol-bisztyga: Honestly, the name is not a problem. It can be anything, really. If the fact that this function is in a separate file and it lies in this directory is okay then we're good to go I think.

tests/integration/pure_cairo_vm/cheatcodes/expect_call/expect_call_test.cairo
```diff
@@ -0,0 +1,64 @@
+from starkware.cairo.common.math import assert_not_zero
```
Arcticae, line 1: Can we add a test for proxy calling the contract? This should also be registered as a contract call.

Arcticae (changes requested):

protostar/starknet/cheaters/contracts.py
```diff
@@ -288,3 +297,29 @@ async def call(
...
+            calldata=cairo_calldata,
+        )
+
+        def stop_callback():
+            assert_no_expected_calls(
+                expected_calls=self.cheatable_state.expected_contract_calls_cairo,
+                fixed_address=contract_address,
+            )
+
+        return stop_callback
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: Maybe we should have a separate `MockingCheater`

protostar/starknet/cheatable_cached_state.py
```diff
@@ -45,6 +45,9 @@ def __init__(
         self.expected_contract_calls: dict[
             Address, list[tuple[SelectorType, CairoOrPythonData]]
         ] = {}
+        self.expected_contract_calls_cairo: dict[
+            Address, list[tuple[str, CairoOrPythonData]]
+        ] = {}
```
MaksymilianDemitraszek, line 50: Why do we need two maps with the same purpose 🤔

tests/integration/pure_cairo_vm/cheatcodes/expect_call/expect_call_test.cairo
```diff
@@ -0,0 +1,64 @@
...
+
+func test_expect_call_success(){
```
MaksymilianDemitraszek, line 3: Please add tests:
- [] Expect call through proxy
- [] Expect call in a remote contract (with the depth of call 2) -> expect call on contract C and test calls B which calls C
- [] expect call with selector instrad of str

protostar/testing/test_case_runners/test_case_runner.py
```diff
@@ -34,11 +37,19 @@ def __init__(
         self._output_recorder = output_recorder
         self._stopwatch = stopwatch
 
-    async def run(self) -> TestCaseResult:
+    async def run(self, state: TestExecutionState) -> TestCaseResult:
         try:
             with self._stopwatch.lap(self._test_case.test_fn_name):
                 execution_result = await self._run_test_case()
 
+            # check expected calls
```
MaksymilianDemitraszek, line 45: Unnecessary comment

protostar/starknet/cheaters/utils.py
MaksymilianDemitraszek: I don't think it should be located in this module, this is not a general util function but bussines logic which is reused in two places. I would create `ExpectCheaters`, put logic there and include this function there, WDYT?

protostar/starknet/cheaters/contracts.py
```diff
@@ -288,3 +297,29 @@ async def call(
             general_config=StarknetGeneralConfig(),
         )
         return result.retdata
+
+    async def expect_call(
```
MaksymilianDemitraszek, line 301: I would put it in `ExpectCheaters`, contract cheaters are for legal starknet operations :police:

MaksymilianDemitraszek (changes requested):
Please include more tests

protostar/starknet/cheaters/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    expected_calls: dict[Address, list[tuple[str, CairoOrPythonData]]],
+    fixed_address: Optional[Address] = None,
+) -> None:
+    if expected_calls:
+        msg = "expected calls not fulfilled:"
+        any_addresses_present = False
+        for address, details in expected_calls.items():
+            if fixed_address is not None and address != fixed_address:
+                continue
+            any_addresses_present = True
```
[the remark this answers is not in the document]
karol-bisztyga: I think it would be needed anyway, we would need to know if we had at least one iteration in the loop, because if we didn't we'd not want to raise an error.

What you were thinking about would be correct if we had the condition `if expected_calls:` in the beginning, which you pointed out as redundant. So if we remove it, we need the flag.

protostar/starknet/cheaters/utils.py
```diff
@@ -0,0 +1,24 @@
...
+def assert_no_expected_calls(
+    expected_calls: dict[Address, list[tuple[str, CairoOrPythonData]]],
+    fixed_address: Optional[Address] = None,
+) -> None:
+    if expected_calls:
+        msg = "expected calls not fulfilled:"
+        any_addresses_present = False
+        for address, details in expected_calls.items():
+            if fixed_address is not None and address != fixed_address:
+                continue
```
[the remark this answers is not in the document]
karol-bisztyga: We could do that, but I think it's more readable now, plus comprehensions allocate extra memory. I would leave it as is.

protostar/starknet/cheatable_cached_state.py
```diff
@@ -174,6 +177,32 @@ def unregister_expected_call(
             if not data_for_address:
                 del self.expected_contract_calls[contract_address]
 
+    def register_expected_call_cairo(
+        self, contract_address: Address, function_name: str, calldata: CairoData
+    ):
+        if self.expected_contract_calls_cairo.get(contract_address):
```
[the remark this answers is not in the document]
karol-bisztyga: What do you mean? I don't see any weird behavior here but maybe I'm missing something?

`self.expected_contract_calls_cairo.get(contract_address)` returns just `None` and leaves the `dict` unchanged (for `defaultdict` as well).

protostar/starknet/cheatable_cached_state.py
```diff
@@ -174,6 +177,32 @@ def unregister_expected_call(
...
+        self, contract_address: Address, function_name: str, calldata: CairoData
+    ):
+        data_for_address = self.expected_contract_calls_cairo.get(contract_address)
+        if not data_for_address:
+            return
+        for index, (stored_function_name, stored_calldata) in enumerate(
+            data_for_address
+        ):
+            if stored_function_name == function_name and stored_calldata == calldata:
+                del data_for_address[index]
```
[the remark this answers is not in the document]
karol-bisztyga: Well, ok. I came up with something like this:
```
self.expected_contract_calls_cairo[contract_address] = [
    (stored_function_name, stored_calldata)
    for (stored_function_name, stored_calldata) in self.expected_contract_calls_cairo[contract_address]
    if stored_function_name != function_name or stored_calldata != calldata
]
```
Instead of this
```
for index, (stored_function_name, stored_calldata) in enumerate(
    data_for_address
):
    if stored_function_name == function_name and stored_calldata == calldata:
        del data_for_address[index]
```

I don't know, for me, it's far worse to read but I don't mind that much, also it adds a memory overhead (additional alloc instead of unlink? As far as I understand) but we don't operate on big data sets so that's fine I guess.

I will follow your advice but honestly, I'm not convinced this is better.

tests/integration/pure_cairo_vm/cheatcodes/expect_call/expect_call_test.cairo
karol-bisztyga: I think this is blocked by https://github.com/software-mansion/protostar/issues/1361

I'm going to add a TODO to this PR with the proxy tests

protostar/starknet/cheatable_cached_state.py
karol-bisztyga: Ye, I wrote above that they should be merged into just one

protostar/starknet/cheaters/utils.py
```diff
@@ -0,0 +1,24 @@
...
+from protostar.protostar_exception import ProtostarException
+from protostar.starknet.address import Address
+from protostar.starknet.data_transformer import CairoOrPythonData
+
+
+def assert_no_expected_calls(
+    expected_calls: dict[Address, list[tuple[str, CairoOrPythonData]]],
+    fixed_address: Optional[Address] = None,
+) -> None:
+    if expected_calls:
```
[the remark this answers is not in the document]
karol-bisztyga: [address various feedback comments](https://github.com/software-mansion/protostar/pull/1363/commits/9b066b37a326344e44dbcc224db465ed14eddb4e)

protostar/starknet/cheaters/utils.py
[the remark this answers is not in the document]
karol-bisztyga: Please let me know if you feel strongly or we can resolve this.

protostar/starknet/cheaters/utils.py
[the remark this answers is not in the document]
karol-bisztyga: [address various feedback comments](https://github.com/software-mansion/protostar/pull/1363/commits/9b066b37a326344e44dbcc224db465ed14eddb4e)

protostar/starknet/cheatable_cached_state.py
```diff
@@ -174,6 +177,32 @@ def unregister_expected_call(
...
+            self.expected_contract_calls_cairo[contract_address] = [
+                (function_name, calldata)
+            ]
+
+    def unregister_expected_call_cairo(
+        self, contract_address: Address, function_name: str, calldata: CairoData
+    ):
+        data_for_address = self.expected_contract_calls_cairo.get(contract_address)
+        if not data_for_address:
+            return
```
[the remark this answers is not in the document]
karol-bisztyga: [address various feedback comments](https://github.com/software-mansion/protostar/pull/1363/commits/9b066b37a326344e44dbcc224db465ed14eddb4e)

protostar/starknet/cheatable_cached_state.py
```diff
@@ -174,6 +177,32 @@ def unregister_expected_call(
...
+        for index, (stored_function_name, stored_calldata) in enumerate(
+            data_for_address
+        ):
```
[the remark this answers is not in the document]
karol-bisztyga: [address various feedback comments](https://github.com/software-mansion/protostar/pull/1363/commits/9b066b37a326344e44dbcc224db465ed14eddb4e)

protostar/starknet/cheatable_cached_state.py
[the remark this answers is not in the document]
karol-bisztyga: [address various feedback comments](https://github.com/software-mansion/protostar/pull/1363/commits/9b066b37a326344e44dbcc224db465ed14eddb4e)

tests/integration/pure_cairo_vm/cheatcodes/expect_call/expect_call_test.cairo
karol-bisztyga: Added as TODO in the main PR comment, resolving this one

tests/integration/pure_cairo_vm/cheatcodes/expect_call/expect_call_test.cairo
karol-bisztyga: Adding as todos in the main PR comment

protostar/testing/test_case_runners/test_case_runner.py
karol-bisztyga: [address various feedback comments](https://github.com/software-mansion/protostar/pull/1363/commits/9b066b37a326344e44dbcc224db465ed14eddb4e)

tests/integration/pure_cairo_vm/cheatcodes/expect_call/expect_call_cairo_cheatcode_test.py
karol-bisztyga: [rename](https://github.com/software-mansion/protostar/pull/1363/commits/4fa80177f786bb5b121e6ba257e25b070a524e33)

protostar/starknet/cheatable_cached_state.py
karol-bisztyga: [unify expected call data](https://github.com/software-mansion/protostar/pull/1363/commits/dce29d380746999059899f807331f7f29dc849d5)

protostar/starknet/cheaters/contracts.py
karol-bisztyga: [add 'expects' cheater, modify cheaters hierarchy](https://github.com/software-mansion/protostar/pull/1363/commits/650efd7813074d4654e979c9a21a14da3dffad95)

protostar/starknet/cheaters/contracts.py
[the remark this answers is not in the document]
karol-bisztyga: [add 'expects' cheater, modify cheaters hierarchy](https://github.com/software-mansion/protostar/pull/1363/commits/650efd7813074d4654e979c9a21a14da3dffad95)

protostar/starknet/cheaters/utils.py
karol-bisztyga: [add 'expects' cheater, modify cheaters hierarchy](https://github.com/software-mansion/protostar/pull/1363/commits/650efd7813074d4654e979c9a21a14da3dffad95)

karol-bisztyga:
Ok, I think you guys can take a look. Please, pay special attention to [add 'expects' cheater, modify cheaters hierarchy](https://github.com/software-mansion/protostar/pull/1363/commits/650efd7813074d4654e979c9a21a14da3dffad95) because these changes are the most radical and important I think.

Thanks for the reviews so far ✌️

protostar/starknet/cheaters/stateful.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, cheatable_state: "CheatableCachedState"):
+        self.cheatable_state = cheatable_state
+
+    def copy(self) -> Self:
+        return copy.copy(self)
+
+    def apply(self, parent: Self) -> None:
+        pass
+
+    async def _transform_calldata_to_cairo_data_by_addr(
```
MaksymilianDemitraszek, line 33: TBH I think this class is a bit messy, 
`_transform_calldata_to_cairo_data_by_addr`
`_transform_calldata_to_cairo_data`
are just a bit random logic dump touching different parts of application.
I still think we should add
`get_contract_class_by_address` to cheatable state or separate function. If we need some transformation logic abstraction we should add it to the transformer.

protostar/testing/cheatcodes/expect_call_cheatcode.py
```diff
@@ -32,25 +25,25 @@ def build(self) -> Callable[[int, str, list[int]], Callable]:
         return self.expect_call
 
     def expect_call(
-        self, raw_address: RawAddress, fn_name: str, calldata: list[int]
+        self, raw_address: RawAddress, function_name: str, calldata: list[int]
```
MaksymilianDemitraszek, line 28: Those changes seem out of scope of this PR

protostar/testing/cairo_cheatcodes/expect_call_cairo_cheatcode.py
```diff
@@ -0,0 +1,48 @@
...
+        function_name: str,
+        calldata: Optional[CairoOrPythonData] = None,
+    ) -> Callable:
+        try:
+            return await self.cheaters.expects.expect_call(
+                contract_address=contract_address,
+                function_name=function_name,
+                calldata=calldata,
+            )
+        except ContractsCheaterException as exc:
```
MaksymilianDemitraszek, line 47: Is it a correct expection to catch?

MaksymilianDemitraszek (approved):
I am ok with doing tests in a separate PR but please create task/s for that so we don't forget

protostar/testing/cairo_cheatcodes/expect_call_cairo_cheatcode.py
karol-bisztyga: Hmm, right, we should catch `CheaterException`, thanks!

protostar/starknet/cheaters/stateful.py
karol-bisztyga: Do you feel strongly about this? Or can we address this separately?

I don't find this messy, this function ("overloaded") is pretty generic and should be in such a mutual place - it is used by both, contract and expects cheaters.

karol-bisztyga: I also don't fully understand what you said :/ Are there 2 separate parts of your comments or are they related? How adding `get_contract_class_by_address` is connected to the fact that `_transform_calldata_to_cairo_data` is where it is now?

protostar/testing/cairo_cheatcodes/expect_call_cairo_cheatcode.py
karol-bisztyga: [fix error type](https://github.com/software-mansion/protostar/pull/1363/commits/e95b228dec4bb0e767ae21e28e9712961a7cd2ba)

protostar/testing/cheatcodes/expect_call_cheatcode.py
MaksymilianDemitraszek: I am not sure if we even have to unify those intefeaces, It could have been adressed separately, but if you feel it is not worth it you can leave it

karol-bisztyga: It's better to do it, otherwise, we'd end up with two different structures/sets of methods for the same purpose. For the reference, you can check out [protostar/starknet/cheatable_cached_state.py](https://github.com/software-mansion/protostar/pull/1363/commits/dce29d380746999059899f807331f7f29dc849d5#diff-8d74f2401d78312e7801aa75441a7f0268fc107e46778325d4354d83908bd104): `self.expected_contract_calls` and `self.expected_contract_calls_cairo`

protostar/starknet/cheaters/stateful.py
karol-bisztyga: As I said, I'm not sure if I understood well what you meant. Could you create an issue? I'd do it but I'm afraid I'm not going to attach all the information in the proper way.

karol-bisztyga: Ok, explained offline, [move transform_calldata_to_cairo_data](https://github.com/software-mansion/protostar/pull/1363/commits/1ce5ae1de8a3f05f51ed99b1e898aed3d9bd50fd)

protostar/starknet/cheatable_cached_state.py
```diff
@@ -156,23 +165,46 @@ def get_abi_from_contract_address(self, contract_address: int) -> AbiType:
...
+        if not self.expected_contract_calls[contract_address]:
+            del self.expected_contract_calls[contract_address]
+
+    async def get_contract_class_by_address(
+        self, contract_address: Union[Address, int]
+    ) -> ContractClass:
+        if isinstance(contract_address, Address):
+            contract_address = int(contract_address)
+        class_hash = await self.get_class_hash_at(contract_address)
+        return await self.get_contract_class(class_hash=class_hash)
```
Arcticae, line 207: Let's move this to the cheater itself. The cheater will have access to the state anyway

protostar/testing/test_case_runners/test_case_runner.py
```diff
@@ -34,11 +38,18 @@ def __init__(
...
             with self._stopwatch.lap(self._test_case.test_fn_name):
                 execution_result = await self._run_test_case()
 
+            try:
+                assert_no_expected_calls(
+                    state.starknet.cheatable_state.cheatable_state.expected_contract_calls
+                )
+            except (ProtostarException, CheaterException) as e:
+                raise ReportedException from e
+
```
Arcticae, line 52: I think this should be moved elsewhere, `test_case_runner` seems like a bad place to include cheating logic

protostar/starknet/cheatable_cached_state.py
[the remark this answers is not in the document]
karol-bisztyga: Ok, I misunderstood what you meant by saying "consider" here. You're right, but I discovered that to make it work, we have to implement a condition in the state class' `_apply` method. We have to do this:
```
        if parent.expected_contract_calls or self.expected_contract_calls:  # <--- adding this if
            parent.expected_contract_calls = {
                **parent.expected_contract_calls,
                **self.expected_contract_calls,
            }
```
Otherwise, we're going to have `{}` instead of a `defaultdict` object, and eventually, a `KeyError` will be raised here. So either way we have to perform a check.

I think it is better to do it here, it's just safer. If for some reason, `expected_contract_calls` is modified somehow and we get a `{}` here, we'll get a `KeyError`. This simple check here makes sure it never happens. What you proposed is a nice trick but I think it's not worth implementing in this case considering the risk.

protostar/starknet/cheatable_cached_state.py
karol-bisztyga: This is also used by the cheatable syscall handler and the old arch implementation of the cheat code. If we move this to the cheater it will become exclusive to it and the old cheat code will no longer work :/

Arcticae: New arch will have a separate state class, so we can leverage that by leaving the old one as-is, and leaving new implementation cleaner.

protostar/testing/test_case_runners/test_case_runner.py
Arcticae: Yep, that seems like a good place for this logic

Arcticae (approved):
Apart from this one i think it's pretty much good to go. 
https://github.com/software-mansion/protostar/pull/1363#discussion_r1073530254

Please consider merging after rebase on #1368 merge to master

protostar/starknet/cheatable_cached_state.py
karol-bisztyga: With the new changes, it is moved to the cheater now.

protostar/cheatable_starknet/cheaters/contracts.py
```diff
@@ -253,14 +218,28 @@ async def call(
         contract_address: Address,
         function_name: str,
         cheaters: "CairoCheaters",
+        unregister_expected_call: Callable,
```
karol-bisztyga, line 221: I don't really like passing it like this, @Arcticae is there a way I can access a cheater from another cheater? Would it make sense to create such links? Or passing stuff like this is alright?
Right now, I access this from the level of the cheat code, via the cheatable state which has a list of cheaters, and, only then, I pass it here.

protostar/cheatable_starknet/cheaters/contracts.py
```diff
@@ -253,14 +218,28 @@ async def call(
...
         calldata: Optional[CairoOrPythonData] = None,
     ) -> CairoData:
         contract_address_int = int(contract_address)
-        cairo_calldata = await self._transform_calldata_to_cairo_data_by_addr(
-            contract_address=contract_address,
+        cairo_calldata = await transform_calldata_to_cairo_data(
+            contract_class=await self.cheatable_state.get_contract_class(
+                class_hash=await self.cheatable_state.get_class_hash_at(
+                    contract_address=contract_address_int
+                )
```
karol-bisztyga, line 229: this is repeated too many times, I'd definitely create a method like `get_contract_class_by_address`

karol-bisztyga: [add get_contract_class_by_address](https://github.com/software-mansion/protostar/pull/1363/commits/856826bf8d1207436a370d8aa5ea9dae60670a62)

protostar/cheatable_starknet/cheaters/contracts.py
Arcticae: I think the only thing you need to instantiate a cheater is cheatable state, so we don't really have to rely on passing them around. I'd be careful though, since we really don't want to interlink their logic, it will create spaghetti

Arcticae: In this case unregistering should happen in handling the cheatable syscall, right? No need to unregister in the contracts cheater itself

protostar/cairo_testing/cairo_test_execution_environment.py
```diff
@@ -36,9 +40,17 @@ def __init__(self, program: Program, state: CairoTestExecutionState):
 
     async def execute(self, function_name: str) -> TestExecutionResult:
         with self.state.output_recorder.redirect("test"):
-            return TestExecutionResult(
+            result = TestExecutionResult(
                 execution_resources=await self.execute_test_case(function_name)
             )
+            try:
+                if isinstance(self.state.starknet.state.state, CheatableCachedState):
```
Arcticae, line 47: please extract this to a variable

protostar/cheatable_starknet/cheaters/contracts.py
```diff
@@ -224,8 +185,10 @@ async def prepare(
         constructor_calldata: CairoOrPythonData,
         salt: int,
     ) -> PreparedContract:
-        constructor_calldata = await self._transform_calldata_to_cairo_data(
-            class_hash=declared.class_hash,
+        constructor_calldata = await transform_calldata_to_cairo_data(
+            contract_class=await self.cheatable_state.get_contract_class(
+                class_hash=to_bytes(declared.class_hash, 32, "big")
```
Arcticae, line 190: Use `starkware`'s byte transform here

protostar/cheatable_starknet/cheaters/expects.py
```diff
@@ -0,0 +1,104 @@
...
+    any_addresses_present = False
+    for address, details in expected_calls.items():
+        if fixed_address is not None and address != fixed_address:
+            continue
+        any_addresses_present = True
+        msg += f"{os.linesep}  - for address {address}:"
+        for details_item in details:
+            msg += f"{os.linesep}    - function name: {details_item[0]}, calldata: {details_item[1]}"
+    if any_addresses_present:
+        raise CheaterException(msg)
```
Arcticae, line 34: We raise a `CheaterException` here, let's make this a `staticmethod` of `ExpectsCairoCheater`.

protostar/cheatable_starknet/cheaters/contracts.py
karol-bisztyga: There are two scenarios when we unregister:
- `stop_expect` is called
- end of the scope is reached

So I think we kind of need this in both, a call's cheater and in the syscall handler

protostar/cheatable_starknet/cheaters/contracts.py
karol-bisztyga: [address feeback](https://github.com/software-mansion/protostar/pull/1363/commits/18139ec67977c1e8a2a28880088e9ba1d613f633)

protostar/cheatable_starknet/cheaters/expects.py
karol-bisztyga: [address feeback](https://github.com/software-mansion/protostar/pull/1363/commits/18139ec67977c1e8a2a28880088e9ba1d613f633)

protostar/cairo_testing/cairo_test_execution_environment.py
```diff
@@ -44,9 +44,10 @@ async def execute(self, function_name: str) -> TestExecutionResult:
...
             )
             try:
-                if isinstance(self.state.starknet.state.state, CheatableCachedState):
-                    assert_no_expected_calls(
-                        self.state.starknet.state.state.expected_contract_calls
+                maybe_cheatable_state = self.state.starknet.state.state
+                if isinstance(maybe_cheatable_state, CheatableCachedState):
+                    ExpectsCairoCheater.assert_no_expected_calls(
+                        maybe_cheatable_state.expected_contract_calls
                     )
```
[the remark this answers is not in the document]
karol-bisztyga: [replace check with assert](https://github.com/software-mansion/protostar/pull/1363/commits/1b78e7882e183a2a3dc7a4cba5c6da66576f5eef)

MaksymilianDemitraszek (approved):

protostar/cheatable_starknet/cheatcodes/expect_call_cairo_cheatcode.py
```diff
@@ -0,0 +1,48 @@
...
+        function_name: str,
+        calldata: Optional[CairoOrPythonData] = None,
+    ) -> Callable:
+        try:
+            return await self.cheaters.expects.expect_call(
+                contract_address=contract_address,
+                function_name=function_name,
+                calldata=calldata,
+            )
+        except CheaterException as exc:
```
MaksymilianDemitraszek, line 47: Can we not catch CheaterException here, but specific errors raised from cheaters? It is definition of user-interface and the decision should be defined in Cheatcode logic

MaksymilianDemitraszek (changes requested):

protostar/cheatable_starknet/cheatcodes/expect_call_cairo_cheatcode.py
karol-bisztyga: Sorry, but I don't understand what's wrong with catching the parent class error here. From some places, we're going to receive just `CheaterException`, for example from `transform_calldata_to_cairo_data`.

> the decision should be defined in Cheatcode logic

what decision? I understand that we just get an error here and we rethrow so the error message is accessible higher in the stack.

If you'd prefer to make changes to the error-handling mechanism maybe that should be placed in a separate PR.

MaksymilianDemitraszek: It can be done in separate PR if you prefer to do so, but it would be simpler to include it here. I provided more extensive argument behind this/explanation on slack, I probably should let you know about it here, sorry 😞

karol-bisztyga: Okay, I think we should wait until #1404 is merged

protostar/cheatable_starknet/cheaters/expects.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+
+        def stop_callback():
+            try:
+                ExpectsCairoCheater.assert_no_expected_calls(
+                    expected_calls=self.cheatable_state.expected_contract_calls,
+                    fixed_address=contract_address,
+                )
+            except ExpectsCheaterException as e:
+                raise CheatcodeException("expect_call", e.message) from e
```
karol-bisztyga, line 59: Is this ok? `stop_callback` is going to be called from hints:
```
stop_expect = expect_call(addr, "increase_balance", [50])
stop_expect()
```

protostar/starknet/cheatable_syscall_handler.py
```diff
@@ -142,9 +142,11 @@ def _call_contract(
             raise NotImplementedError(f"Unsupported call type {syscall_name}.")
 
         # region Modified Starknet code.
-        contract_calldata = (int(str(request.function_selector)), calldata)
+        selector = int(str(request.function_selector))
         self.cheatable_state.unregister_expected_call(
-            contract_address=Address(contract_address), calldata=contract_calldata
+            contract_address=Address(contract_address),
+            function_selector=selector,
+            calldata=calldata,
```
Arcticae, line 149: This is also an old implementation

protostar/starknet/cheater.py
```diff
@@ -3,6 +3,12 @@
 from typing_extensions import Self
 
 
+class CheaterException(Exception):
+    def __init__(self, message: str):
+        super().__init__(message)
+        self.message = message
```
Arcticae, line 9: There is no need here for the `__init__` method. If you inherit from `Exception` you can do `str(CheaterException("message"))` and it will yield `"message"`

protostar/starknet/data_transformer.py
```diff
@@ -92,3 +95,24 @@ def transform(data: CairoData) -> PythonData:
...
     return transform
+
+
+async def transform_calldata_to_cairo_data(
+    contract_class: ContractClass,
+    function_name: str,
+    calldata: Optional[CairoOrPythonData] = None,
+) -> CairoData:
+    if isinstance(calldata, collections.Mapping):
+        assert contract_class.abi, "Abi required but not found"
```
Arcticae, line 106: I feel like this should also be `DataTransformerException` but i'm not completely sure of all the cases on this one

protostar/testing/cheatcodes/expect_call_cheatcode.py
```diff
@@ -8,13 +8,6 @@
 from protostar.starknet import RawAddress, Address
```
Arcticae, line 8: This is an old implementation and should be left unmodified

Arcticae (changes requested):

protostar/cheatable_starknet/cheaters/expects.py
Arcticae: I'm not completely sure if we still will be able to return callbacks in cairo 1.0. Probably we'll use an id-based solution (like `setTimeout`/`clearTimeout` does). Base class of the cheatcode states that it will be wrapped in a `ExecutionResult` anyways right now

protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
```diff
@@ -32,8 +38,18 @@ def __init__(
...
+            )
+            try:
+                cheatable_state = self.state.starknet.state.state
+                assert isinstance(cheatable_state, CheatableCachedState)
+                ExpectsCairoCheater.assert_no_expected_calls(
+                    cheatable_state.expected_contract_calls
+                )
+            except (ProtostarException, ExpectsCheaterException) as e:
+                raise ReportedException from e
+            return result
```
Arcticae, line 52: I think this part could be done with a test finish hook, like in #1412 (`ExpectEventsController`)

karol-bisztyga: Yes, I saw that too 👍

Arcticae:
@karol-bisztyga When reimplementing, please remember to address the TODO at `protostar/cheatable_starknet/cheatables/cheatable_syscall_handler.py:126`

karol-bisztyga:
closing in favor of https://github.com/software-mansion/protostar/pull/1481

PR #1368, opened 2023-01-11
Swap for cheatable entrypoints in `invoke` and `call` cairo cheatcodes

Arcticae:
- [x] Make it work with `invoke` as well


tests/integration/pure_cairo_vm/cheatcodes/warp/warp_test.cairo
```diff
@@ -5,10 +5,37 @@ func test_warp(){
...
+    assert_not_zero(deployed_contract_address);
+
+    // Set the timestamp to rolled value
+    %{ invoke(ids.deployed_contract_address, "block_timestamp_setter") %}
+
+    // Retrieve stored value
+    local stored_block_timestamp;
+    %{ ids.stored_block_timestamp = call(ids.deployed_contract_address, "stored_block_timestamp_getter")[0] %}
+    assert stored_block_timestamp = 123;
     return ();
```
kasperski95, line 40: Tests in this file could in one big hint.

kasperski95 (commented):
Why can't `CheatableCachedState` be reused? Duplicating Cheatables wouldn't be necessary.

protostar/starknet/cheaters_of.py
```diff
@@ -0,0 +1,41 @@
...
+
+from services.everest.business_logic.state_api import StateProxy
+from starkware.starknet.business_logic.fact_state.state import CarriedState
+from starkware.starknet.business_logic.state.state import CachedState, StateSyncifier
+
+if TYPE_CHECKING:
+    from protostar.starknet.cheaters import Cheaters
+
+
+def cheaters_of(state: StateProxy) -> "Cheaters":
```
MaksymilianDemitraszek, line 11: We remove cheaters from cached state so i believe this function can be removed

protostar/starknet/cairo_cheatcode_syscall_handler.py
```diff
@@ -0,0 +1,204 @@
...
+    def __init__(self, message: str):
+        self.message = message
+        super().__init__(message)
+
+
+class CheatableCairoSysCallHandler(BusinessLogicSysCallHandler):
+    def __init__(self, state: SyncState, **kwargs: Any):
+        # This field must be set before entering super constructor,
+        # because it calls the setter for the `block_info` property.
+        self.cheaters = cheaters_of(state)
```
MaksymilianDemitraszek, line 33: `CheatableCairoSysCallHandler` handler shouldn't have access to cheaters

MaksymilianDemitraszek (changes requested):
Can we move copy and apply logic from cheaters to the CheatableCachedState and remove cheaters from CheatableCachedState? We can do this in this PR or in a separate PR

protostar/commands/test/test_command.py
```diff
@@ -26,7 +26,7 @@
     TestScheduler,
     determine_testing_seed,
 )
-from protostar.testing.cairo_test_runner import CairoTestRunner
+from protostar.testing.new_arch.cairo_test_runner import CairoTestRunner
```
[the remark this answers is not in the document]
Arcticae: then we'll name the module something else, or flatten the structure

protostar/starknet/cheatable_cached_state.py
```diff
@@ -175,7 +174,16 @@ def unregister_expected_call(
                 del self.expected_contract_calls[contract_address]
 
 
-def cheaters_of(state: StateProxy) -> Cheaters:
+class CheatableStateException(Exception):
```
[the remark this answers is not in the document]
Arcticae: It's the way it was implemented previously, no change here

protostar/starknet/new_arch/cheatable_cairo_cached_state.py
```diff
@@ -0,0 +1,176 @@
...
+from typing_extensions import Self
+
+from protostar.starknet.address import Address
+from protostar.starknet.new_arch.cheaters import BlockInfoCairoCheater
+from protostar.starknet.types import ClassHashType, SelectorType
+from protostar.starknet.data_transformer import CairoOrPythonData
+
+
+# pylint: disable=too-many-instance-attributes
+class CheatableCairoCachedState(CachedState):
```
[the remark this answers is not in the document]
Arcticae: > the name does not reflect that

The namings is consistent with all of the components which are relevant to cairo testing flow. It was called `CheatableCachedState` previously, so no major changes introduced in naming. 

> I would create a component called `protostar.cheatable_starknet_fixture`

It's not really a fixture as well, since it's extension of `state`. 

> Similarly, I'd create a new component `protostar.cairo_testing` instead of `protostar.testing.new_arch.`

The idea i had in putting those new counterparts in `new_arch`, is that when we remove old flow completely, we can just copy things over one catalogue higher and ditch the `new_arch` prefix. I think creating completely new structure just for sake of refactoring at this point may result in confusion and will make the rewrite less workable.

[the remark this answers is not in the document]
kasperski95: > The namings is consistent with all of the components which are relevant to cairo testing flow. It was called CheatableCachedState previously, so no major changes introduced in naming.

The naming is consistent, although it's confusing. `CheatableCairoCachedState` on its own could be interpreted in two invalid ways:
- `CheatableCairo` `CachedState` — Protostar cheats on StarkNet, not Cairo.
- `Cheatable` `CairoCachedState` — It's StarkNet's state, not Cairo's.

`Cairo` prefix makes sense within `protostar.testing` component, but it doesn't in `protostar.starknet`. This could in a different PR.

> It's not really a fixture as well, since it's extension of state.

In this case, I meant renaming/relocating the directory `new_arch`, not the class name.

> The idea i had in putting those new counterparts in new_arch, is that when we remove old flow completely, we can just copy things over one catalogue higher and ditch the new_arch prefix. I think creating completely new structure just for sake of refactoring at this point may result in confusion and will make the rewrite less workable.

Ditching `new_arch` will also create a confusion. Refactoring often introduces confusion, but it's temporary. This PR changes the structure anyway.

---

After looking at this PR again, I'd move these components in the following way:
`protostar.starknet.new_arch` -> `protostar.cheatable_starknet`
`protostar.testing.new_arch` -> `protostar.cheatable_starknet_fixture`

`protostar.starknet` and `protostar.testing` are already quite large, and this is a good opportunity to move stuff out of them.

protostar/starknet/new_arch/cheatable_cairo_entry_point.py
```diff
@@ -0,0 +1,281 @@
...
+            entry_point_type=EntryPointType.EXTERNAL,
+            caller_address=0,
+        )
+
+    @classmethod
+    def factory(cls, cheaters: "CairoCheaters") -> type[ExecuteEntryPoint]:
+        def factory_function(*args: Any, **kwargs: Any) -> ExecuteEntryPoint:
+            return cls(*args, cheaters=cheaters, **kwargs)
+
+        return cast(type[ExecuteEntryPoint], factory_function)
```
kasperski95, line 281: Don't we usually keep class methods at the of the class?

protostar/starknet/new_arch/cheaters/contracts.py
```diff
@@ -63,35 +68,9 @@ class ConstructorInvocationException(ContractsCheaterException):
     pass
 
 
-class ContractsCheater(Cheater):
```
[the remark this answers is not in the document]
Arcticae: I guess it would be okay, but only thing it would include would be a constructor. Don't know if that's worth the complexity ATM

protostar/starknet/new_arch/cheatable_cairo_cached_state.py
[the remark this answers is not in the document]
Arcticae: > After looking at this PR again, I'd move these components in the following way:
`protostar.starknet.new_arch` -> `protostar.cheatable_starknet`
`protostar.testing.new_arch` -> `protostar.cheatable_starknet_fixture`

Agree with the first, but the second one includes execution environment, test runner, etc. 
I would rather go with the first proposition and name it `cairo_testing`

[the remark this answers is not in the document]
Arcticae: As for the name of `CheatableCairoCachedState` i would gladly accept another name if you have ideas. `StarknetCheatableCachedState` is not acceptable because `StarknetState` is not an instance of `CachedState` - those are two different concepts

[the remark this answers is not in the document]
kasperski95: > I would rather go with the first proposition and name it cairo_testing

Ok, let it be `protostar.cairo_testing`. I suggested `protostar.cheatable_starknet_fixture` because `protostar.cairo_testing` should be unaware of StarkNet.

> As for the name of CheatableCairoCachedState i would gladly accept another name if you have ideas. StarknetCheatableCachedState is not acceptable because StarknetState is not an instance of CachedState - those are two different concepts

I'd keep `CheatableCachedState` as long as the name is unique within the component. Old classes are going to be removed anyway. If the name must be globally unique, I'd prefix old `Cheatable` classes with `Deprecated` e.g. `DeprecatedCheatableCachedState`.

protostar/commands/test/test_command.py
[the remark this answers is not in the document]
Arcticae: this was moved

protostar/starknet/new_arch/cheatable_cairo_cached_state.py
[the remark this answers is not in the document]
Arcticae: Done as discussed

protostar/starknet/new_arch/cheatable_cairo_entry_point.py
Arcticae: done

protostar/starknet/cairo_cheatcode_syscall_handler.py
Arcticae: done

protostar/cairo_testing/cairo_test_cheatcode_factory.py
```diff
@@ -0,0 +1,82 @@
...
+)
+from protostar.cheatable_starknet.cheatcodes.call_cairo_cheatcode import (
+    CallCairoCheatcode,
+)
+
+
+class CairoTestCheatcodeFactory:
+    def __init__(
+        self,
+        state: CairoTestExecutionState,
```
kasperski95, line 45:
```suggestion
        cheatable_cached_state: CheatableCachedState,
```

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC
```
kasperski95, line 1: I think `CairoCheatcodes` should be in `cairo_testing`. `CairoCheatcodes` are basically views in the Cairo testing context.

tests/integration/pure_cairo_vm/cheatcodes/warp/warp_test.cairo
```diff
@@ -5,10 +5,37 @@ func test_warp(){
...
+    local timestamp;
+    %{
+        ids.timestamp = call(ids.deployed_contract_address, "timestamp_getter")[0]
+    %}
+
+    assert timestamp = 123;
+    return ();
+}
+
+func test_warp_with_invoke(){
```
MaksymilianDemitraszek, line 23: Can we test it for depth 2? warp contract B, invoke contract A, A calls B, B asserts it's warped

Can be done in a separate PR

tests/integration/pure_cairo_vm/cheatcodes/roll/roll_test.cairo
```diff
@@ -5,10 +5,38 @@ func test_roll(){
...
+
+    return ();
+}
+
+func test_roll_with_invoke(){
+    alloc_locals;
+    local deployed_contract_address;
+
+    %{
+        ids.deployed_contract_address = deploy_contract("./src/main.cairo").contract_address
```
MaksymilianDemitraszek, line 29: Can we test it for depth 2? roll contract B, invoke contract A, A calls B, B asserts it's warped

Can be done in a separate PR

kasperski95 (approved):

protostar/cairo_testing/cairo_test_execution_state.py
```diff
@@ -0,0 +1,81 @@
...
+)
+from protostar.testing.stopwatch import Stopwatch
+from protostar.testing.test_config import TestConfig
+from protostar.testing.test_context import TestContext
+from protostar.testing.test_output_recorder import OutputRecorder
+
+
+@dataclass
+class CairoTestExecutionState:
+    starknet: Starknet
```
MaksymilianDemitraszek, line 30: I think it would be beneficial to use CheatableCahedState here, but it is not something to address in this PR

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
Arcticae: They have more in common with `CheatableState` than cairo testing itself, i'm not entirely convinced on this one.

Arcticae: Also they can modify the state so they're not views per-se

MaksymilianDemitraszek (approved):

protostar/cairo_testing/cairo_test_cheatcode_factory.py
MaksymilianDemitraszek: +1, but I think it should be adressed in a separate PR, alongside with that we should stop using `Starknet` at all, from variouse preformance reasons

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: > I am not sure, cairo_testing is responsible for running test, and CairoCheatcodes are the exposed view layer of cheatable starknet, they have nothing to do with test execution environment

> They have more in common with CheatableState than cairo testing itself, i'm not entirely convinced on this one.
> Also they can modify the state so they're not views per-se

I'd say they are somewhere between `CairoTesting` and `CheatableStarknet`. They adapt input and output data from user to cheaters in this case, and cheaters modify the state. In the future, similar adapters will be needed by the Scarb plugin. Scarb plugin shouldn't depend on a component that includes CairoTesting's adapters.

It's not super important, though. Resolving.

PR #1373, opened 2023-01-13
Allow using custom ABI in `protostar call`

### LINKED ISSUE #1216: Provide a way to bypass ABI contract fetching from API
Generally speaking ABI doesn't have to be compliant with the bytecode present on the blockchain.
We should provide a way for user to pass the contracts' ABI to the commands in such situations. 

This is not high priority, but a nice-to-have.

kasperski95:
Add `--abi` argument to `protostar call`.

Related #1216


protostar/compiler/compiled_contract_reader.py
```diff
@@ -16,13 +16,13 @@ def load_abi_from_contract_path(self, contract_path: Path) -> Optional[AbiType]:
         abi_path = self._get_abi_path_from_contract_path(contract_path)
         if not abi_path.exists():
             return None
-        return self._load_abi(abi_path)
+        return self.load_abi(abi_path)
```
karol-bisztyga, line 19: Shouldn't we first check for the explicitly specified abi and only if there's none, fall back to the "default" one?

protostar/starknet/abi.py
```diff
@@ -28,10 +30,13 @@ def get_function_parameters(contract_abi: AbiType, name: str) -> Dict[str, Cairo
...
-    for item in contract_abi:
-        if item["name"] == name:
-            return item
-    raise AbiItemNotFoundException(f"Couldn't find '{name}' ABI")
+    try:
+        for item in contract_abi:
+            if item["name"] == name:
+                return item
+        raise AbiItemNotFoundException(f"Couldn't find '{name}' ABI")
+    except TypeError as ex:
```
karol-bisztyga, line 38: Could you please explain where possibly a `TypeError` could raise here?

kasperski95: If I remember correctly, `item["name"] == name` can raise `TypeError`. The error is thrown in `test_error_when_custom_abi_is_invalid`.

karol-bisztyga: I think we shouldn't catch a `TypeError` like this, I'd rather validate the abi JSON in the place where it is parsed `compiled_contract_reader.py::load_abi`. It seems that a valid abi looks like this `[{'inputs': [{...}, ...]}, {'outputs': [{...}, ...]}, {...}, ...]` while an invalid one like this `{'abi': [{'inputs': [{...}, ...]}, {'outputs': [{...}, ...]}, {...}, ...]}`.

I think we should either immediately throw or maybe fallback like this:
```
if isinstance(abi, 'dict'):
  abi = abi.get("abi")
```
What you're doing right now is you're validating the abi JSON a couple of levels deeper into the implementation, which is bad because you're handling an invalid input only for one case.

protostar/compiler/compiled_contract_reader.py
kasperski95: We do this in the Call command. This class is unused. I removed it in this PR https://github.com/software-mansion/protostar/pull/1382

protostar/starknet/abi.py
kasperski95: [use load_abi](https://github.com/software-mansion/protostar/pull/1373/commits/6cec71984b61c830903a01eeaae5b6372e761186)

protostar/starknet/abi.py
```diff
@@ -28,14 +32,29 @@ def get_function_parameters(contract_abi: AbiType, name: str) -> Dict[str, Cairo
...
+def load_abi(abi_path: Path) -> AbiType:
+    assert abi_path.suffix == ".json"
+    potential_abi = json.loads(abi_path.read_text())
+    panic_if_invalid_abi(potential_abi)
+    return potential_abi
+
+
+def panic_if_invalid_abi(abi: Any):
+    if not isinstance(abi, list):
+        raise ProtostarException("Invalid ABI.")
```
karol-bisztyga, line 60: Nit: Why this is a separate method? It's a single check used only in this one very place. I don't see why we should exclude every single check to a separate function. That would generate an unmaintainable code.

website/docs/cli-reference.md
```diff
@@ -40,6 +40,8 @@ Input to the account's constructor.
 Print machine-readable output in [NDJSON](https://github.com/ndjson/ndjson-spec) format.
 ### `call`
 Calls a contract on StarkNet with given parameters
+#### `--abi PATH`
+Path to the ABI file to be used by Data Transformer. If not provided, Protostar will get the ABI from the StarkNet.
```
karol-bisztyga, line 44:
```suggestion
Path to the ABI file to be used by Data Transformer. If not provided, Protostar will get the ABI from StarkNet.
```

Grammarly

protostar/cli/common_arguments.py
```diff
@@ -87,3 +87,12 @@
     type="bool",
     default=False,
 )
+
+ABI_PATH_ARG = ProtostarArgument(
+    name="abi",
+    description=(
+        "Path to the ABI file to be used by Data Transformer. "
+        "If not provided, Protostar will get the ABI from the StarkNet."
```
karol-bisztyga, line 95:
```suggestion
        "Path to the ABI file to be used by Data Transformer. "
        "If not provided, Protostar will get the ABI from StarkNet."
```

Grammarly

karol-bisztyga (approved):

protostar/starknet/abi.py
kasperski95: It's not about separating checks. It's more about operating on one level of abstraction. ABI validation is quite a separate concern compared to loading ABI.

https://www.fabrizioduroni.it/2019/07/28/clean-code-functions/

PR #1386, opened 2023-01-18
Extract `protostar.cairo` from `protostar.starknet`

kasperski95:
Create `protostar.cairo` and move Cairo related code there.


MaksymilianDemitraszek:
I will look on it after it is merged with #1368 , so to avoid unnecessary work

protostar/cairo/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .cairo_migrator import Cairo010Migrator
+from .cairo_pass_manager import (
+    PassManagerFactory,
+    CairoPassManagerFactory,
+    PassManagerConfig,
+)
+from .cairo_compiler import CairoCompiler, CairoCompilerConfig
+from .hint_local import HintLocal
```
Arcticae, line 8: I don't really understand the need to re-export things from modules in this project. It only causes more problems with cyclic imports, and we're not exactly exposing this API to anybody, so it doesn't have to be pretty

protostar/cairo/cairo_pass_manager.py
```diff
@@ -0,0 +1,32 @@
...
+class PassManagerConfig:
+    include_paths: list[str]
+    disable_hint_validation: bool
+
+
+class PassManagerFactory(ABC):
+    @staticmethod
+    @abstractmethod
+    def build(config: PassManagerConfig) -> PassManager:
+        ...
```
Arcticae, line 22: This should be in a separate file, preferably not in `cairo` module imho

Arcticae (approved):
Looks good, apart from the nitpicks

protostar/cairo/__init__.py
kasperski95: > It only causes more problems with cyclic imports

Bad design causes cyclic imports problem. Barrel files only highlight this issue.

> and we're not exactly exposing this API to anybody, so it doesn't have to be pretty

We expose component API to ourselves. Barrel files encapsulate information about filenames, so that renames don't propagate across the codebase. They also allow importing multiple items in one line.

PR #1392, opened 2023-01-19
Add `__setup__` for test runner

Arcticae:
Related: #1281


protostar/cairo_testing/cairo_setup_execution_environment.py
```diff
@@ -0,0 +1,15 @@
+import asyncio
+
+from protostar.cairo_testing.cairo_test_execution_environment import (
+    CairoTestExecutionEnvironment,
+)
+
+
+class CairoSetupExecutionEnvironment(CairoTestExecutionEnvironment):
```
kasperski95, line 8: I think Setup Environment shouldn't inherit from Test Environment. Not all cheatcodes should be available in `setup`.

Arcticae: We could split up `_get_hint_locals` into `_get_cheatcodes` and `_get_custom_hint_locals` - then overriding it in the subclass would make it easier to put a different hint factory there. For now i don't see an example in which it's useful.

kasperski95: For example,`expect_events` doesn't make much sense in the setup context. Also, I think this code violates LSP principle, because if `CairoTestExecutionEnvironment` is expected, `CairoSetupExecutionEnvironment` could be passed resulting in unexpected behavior.

> We could split up _get_hint_locals into _get_cheatcodes and _get_custom_hint_locals - then overriding it in the subclass would make it easier to put a different hint factory there. For now i don't see an example in which it's useful.

I wouldn't split `_get_hint_locals`, because cheatcodes are `hint_locals`. This change could be done either by inheritance or preferably by composition. 

- by inheritance, I'd create `CairoExecutionEnvironment` (CEE) that contains `run_cairo_function` concrete method and abstract `_get_hint_locals`.  This class could be implemented by SetupCEE or TestCEE.
- by composition, I'd create `CheatableCairoFunctionRunner` that expects `CheatcodeFactory` and used that runner in `CairoTestExecutionEnvironment` and `SetupTestExecutionEnvironment`.

Arcticae: I'm all for composition. Will attempt to split that up right now so it's clearer later what we established.

kasperski95 (approved):

PR #1404, opened 2023-01-20
Handle Cairo cheatcodes errors

kasperski95:
Return validation errors instead of raising them.


protostar/cheatable_starknet/cheatable_exception.py
```diff
@@ -0,0 +1,7 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class CheatableException(Exception):
```
Arcticae, line 5: Why is it "cheatable"?

protostar/cheatable_starknet/cheatable_entry_point.py
```diff
@@ -280,3 +300,11 @@ def execute(
             general_config=general_config,
             tx_execution_context=tx_execution_context,
         )
+
+    def _map_stark_exception(self, stark_exception: StarkException):
```
Arcticae, line 304:
```suggestion
    def _wrap_stark_exception(self, stark_exception: StarkException):
```

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -1,12 +1,65 @@
...
+]
+
 
 class CairoCheatcode(HintLocal, ABC):
     def __init__(self, cheaters: "CairoCheaters"):
         self.cheaters = cheaters
+
+    @abstractmethod
+    def _build(self) -> Callable:
+        ...
```
Arcticae, line 51: maybe we should just call this differently? `get_implementation` ?

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -1,12 +1,65 @@
...
+        raise self.err
+
+
+@dataclass
+class CairoCheatcodeValidExecution:
+    ok: Any
+    err = None
+
+    def unwrap(self) -> Any:
+        return self.ok
```
Arcticae, line 37: Shouldn't `unwrap` be only callable on a dataclass with valid result? The code will result in an error if we unwrap and then pass the error as a "result" to later calls. We should probably "panic" when unwrapping a result which has an error underneath, so that the code will crash on unwrap instead of a typeerror when passing error to further functions

protostar/cheatable_starknet/cheatable_entry_point.py
```diff
@@ -280,3 +300,11 @@ def execute(
...
+        results = re.findall("Error message: (.*)", stark_exception.message or "")
+        if len(results) > 0:
+            return CheatableException(message=results[0], raw_ex=stark_exception)
+        return CheatableException(
+            message=stark_exception.message or "", raw_ex=stark_exception
+        )
```
karol-bisztyga, line 310:
```suggestion
        message = stark_exception.message or ""
        results = re.findall("Error message: (.*)", message)
        if len(results) > 0:
            message = results[0]
        return CheatableException(message=message, raw_ex=stark_exception)
```

protostar/cheatable_starknet/cheatable_entry_point.py
```diff
@@ -280,3 +300,11 @@ def execute(
             general_config=general_config,
             tx_execution_context=tx_execution_context,
         )
+
+    def _map_stark_exception(self, stark_exception: StarkException):
+        results = re.findall("Error message: (.*)", stark_exception.message or "")
+        if len(results) > 0:
+            return CheatableException(message=results[0], raw_ex=stark_exception)
```
karol-bisztyga, line 307: Why are we poppin' just one element here?

protostar/cheatable_starknet/cheatable_entry_point.py
```diff
@@ -258,6 +261,23 @@ def execute(
...
+            return self._execute(
+                state,
+                resources_manager,
+                general_config,
+                tx_execution_context,
+            )
+        except StarkException as ex:
+            raise self._map_stark_exception(ex)
+
+    def _execute(
```
karol-bisztyga, line 275: do we need to create an extra function for this? Wouldn't it be enough to wrap `return super().execute(...` with a `try ... except`?

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -1,12 +1,65 @@
...
+        def wrapper(*args: Any, **kwargs: Any):
+            try:
+                result = self._build()(*args, **kwargs)
+                return CairoCheatcodeValidExecution(ok=result)
+            except CheatableException as ex:
+                return CairoCheatcodeInvalidExecution(
+                    err=CheatcodeException(
+                        code=encode_shortstring(ex.message), ex=ex.raw_ex
+                    )
+                )
```
karol-bisztyga, line 63:
```suggestion
            try:
                result = self._build()(*args, **kwargs)
                return CairoCheatcodeExecutionResult(ok=result)
            except CheatableException as ex:
                return CairoCheatcodeExecutionResult(
                    err=CheatcodeException(
                        code=encode_shortstring(ex.message), ex=ex.raw_ex
                    )
                )
```

I don't really understand why we need separation into 2 different types here (`CairoCheatcodeInvalidExecution`, `CairoCheatcodeValidExecution`), I think we could have just one type and either pass `ok` or `err`. `unwrap` would check what isn't `None` and behave accordingly (maybe I missed something?).

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: > Shouldn't unwrap be only callable on a dataclass with valid result? The code will result in an error if we unwrap and then pass the error as a "result" to later calls.

`CairoCheatcodeInvalidExecution::unwrap` raises (not returns) an error, which should result in panic.

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: I am not sure if `get_implemention` improves the readability significantly. It should be a private method.

tests/integration/pure_cairo_vm/cheatcodes/deploy/deploy_test.cairo
```diff
@@ -6,10 +6,10 @@ func test_deploying_pipeline_with_path(){
     local deployed_hash;
 
     %{
-        declared = declare("./src/main.cairo")
+        declared = declare("./src/main.cairo").unwrap()
```
karol-bisztyga, line 9: It's a bit confusing that sometimes we use `unwrap()` and sometimes `ok`... Why don't we always use `unwrap()`?

protostar/cheatable_starknet/cheatable_exception.py
kasperski95: `CheaterException` makes more sense?

tests/integration/pure_cairo_vm/cheatcodes/deploy/deploy_test.cairo
kasperski95: I added `unwrap` later, and I didn't change older code.

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: Proper Type-checking. You can't provide a value and an error at the same time. Also, typechecker can detect if ok is not `None`.

```py
if result.err:
  raise ...
result.ok # typechecker deduces it's not None
```

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -1,12 +1,65 @@
...
+
+from starknet_py.cairo.felt import encode_shortstring
 
 from protostar.cairo import HintLocal
+from protostar.cheatable_starknet.cheatable_exception import CheatableException
 
 if TYPE_CHECKING:
     from protostar.cheatable_starknet.cheaters import CairoCheaters
 
+Felt = int
```
MaksymilianDemitraszek, line 13:
```suggestion
```
It is misleading, this is int not a field element

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
MaksymilianDemitraszek: Both are fine imo

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -1,12 +1,65 @@
...
+    err: CheatcodeException
+
+    def unwrap(self) -> Any:
+        raise self.err
+
+
+@dataclass
+class CairoCheatcodeValidExecution:
+    ok: Any
+    err = None
```
MaksymilianDemitraszek, line 34: It should default to `0`

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -1,12 +1,65 @@
...
+    ok = None
+    err: CheatcodeException
+
+    def unwrap(self) -> Any:
+        raise self.err
+
+
+@dataclass
+class CairoCheatcodeValidExecution:
+    ok: Any
```
MaksymilianDemitraszek, line 33: [[Nitpick]] We could maybe make it a generic?

protostar/cheatable_starknet/cheatable_exception.py
MaksymilianDemitraszek: TransactionException? I think it better captures the class of errors we want to capture

MaksymilianDemitraszek: TransactionRevertException maybe?

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -1,12 +1,65 @@
...
+    @abstractmethod
+    def _build(self) -> Callable:
+        ...
+
+    def build(self):
+        def wrapper(*args: Any, **kwargs: Any):
+            try:
+                result = self._build()(*args, **kwargs)
+                return CairoCheatcodeValidExecution(ok=result)
+            except CheatableException as ex:
```
MaksymilianDemitraszek, line 58: I am not sure about this generalisation, we can leave it for now but it may be necessary to differentiate handling between cheatcodes

protostar/cheatable_starknet/cheatable_entry_point.py
kasperski95: We need to operate on error codes, so it must be a one message. This code will change anyway once Starknet will be integrated with Cairo 1.

protostar/cheatable_starknet/cheatable_entry_point.py
kasperski95: [warp stark ex](https://github.com/software-mansion/protostar/pull/1404/commits/96716f39a98a256b80eaf4763bf0307a79413fd1)

protostar/cheatable_starknet/cheatable_exception.py
kasperski95: [rename exception again](https://github.com/software-mansion/protostar/pull/1404/commits/7b51366559cd7ff6aee5d2df347dd068a68973ea)

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: Removed `unwrap`.

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: I don't see a potential use case. If this exception is too concrete, it can be abstracted.

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: [remove felt](https://github.com/software-mansion/protostar/pull/1404/commits/10ef1c98c597b049f223197328e8fca2c052ab15)

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: Preserved the original name.

protostar/cheatable_starknet/cheatable_entry_point.py
```diff
@@ -280,3 +302,11 @@ def execute(
             general_config=general_config,
             tx_execution_context=tx_execution_context,
         )
+
+    def _warp_stark_exception(self, stark_exception: StarkException):
```
Arcticae, line 306:
```suggestion
    def _wrap_stark_exception(self, stark_exception: StarkException):
```

protostar/cheatable_starknet/cheatable_entry_point.py
```diff
@@ -258,6 +263,23 @@ def execute(
...
+    ):
+        try:
+            return self._execute(
+                state,
+                resources_manager,
+                general_config,
+                tx_execution_context,
+            )
+        except StarkException as ex:
+            raise self._warp_stark_exception(ex)
```
Arcticae, line 275:
```suggestion
            raise self._wrap_stark_exception(ex)
```

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
```diff
@@ -1,12 +1,56 @@
...
     from protostar.cheatable_starknet.cheaters import CairoCheaters
 
 
+@dataclass(init=False)
+class CairoCheatcodeInvalidExecution:
+    ok = None
+    err_code: int
+
+    def __init__(self, err_code: int):
+        assert err_code > 0
```
Arcticae, line 23: please add an assertion message so it's easier to debug

Arcticae (approved):

protostar/cheatable_starknet/cheatable_entry_point.py
kasperski95: > do we need to create an extra function for this?

It's not needed. I created it, because it's a good practice.

> Try block should have one line.
https://gist.github.com/bansalankit92/6119c4f7fabe11b7a174a99f5a7bb376#exceptions-error-handling

> I don't see why the logic with max_step cannot be in the execute method (instead of _execute) and why it should be in the try ... except block, it doesn't seem to raise a StarkException.
> Wouldn't it be enough to wrap return super().execute(... with a try ... except?

It would be better. The existing code in the `execute` method wasn't super obvious, I used [wrap method technique](https://gist.github.com/jonnyjava/42883d4e464167f81e2ee60a488a5ded#wrap-method) to avoid touching it. However, after looking at the code in the `_execute` method again, it looks like the `max_steps` code doesn't have an impact anyway. 

[removed _execute, refined code in the CheatableEntryPoint](https://github.com/software-mansion/protostar/pull/1404/commits/e99e4f26a08d14c7b8a95fdbe0fa6cf116254be0)

protostar/cheatable_starknet/cheatable_entry_point.py
kasperski95: [rename warp to wrap](https://github.com/software-mansion/protostar/pull/1404/commits/919ad296758202de5c5244c412aeea1488a7c3e4)

protostar/cheatable_starknet/cheatcodes/cairo_cheatcode.py
kasperski95: [describe assertion](https://github.com/software-mansion/protostar/pull/1404/commits/3d72611785130d5bcd8382b71504473658dea1f9)

protostar/cheatable_starknet/cheatable_entry_point.py
kasperski95: [make max_steps affect execute](https://github.com/software-mansion/protostar/pull/1404/commits/d67e4d972358bcf4cda51c0ad6e1e73690d7ede5)

karol-bisztyga: Makes sense, looks better now

PR #1405, opened 2023-01-20
Improve `assert_cairo_test_cases`' feedback

kasperski95:
Improve 'assert_cairo_test_cases' feedback. `CairoCheatcodes` cairo tests can't be run with Protostar in the console. Checking test results with a debugger is annoying.

Before:
Screenshot 2023-01-20 at 16 27 23

After:
Screenshot 2023-01-20 at 16 33 03

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -3,7 +3,7 @@ import os
 from contextlib import contextmanager
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Callable, ContextManager, List, Optional, Set, cast
+from typing import Any, Callable, ContextManager, List, Optional, Set, cast
 
 import pytest
 from pytest import TempPathFactory
@@ -12,11 +12,12 @@ from starkware.starknet.public.abi import AbiType
 from typing_extensions import Protocol
 
 from protostar.commands.test.test_command import TestCommand
+from protostar.commands.test.test_result_formatter import format_test_result
 from protostar.compiler.project_cairo_path_builder import ProjectCairoPathBuilder
 from protostar.io.log_color_provider import LogColorProvider
 from protostar.testing import TestingSummary
 from protostar.cli import MessengerFactory
-
+from protostar.testing.test_results import TestCaseResult
 from tests.conftest import TESTS_ROOT_PATH, run_devnet
 from tests.integration._conftest import ProtostarFixture, create_protostar_fixture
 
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -81,9 +82,48 @@ def assert_cairo_test_cases(
...
+            if expected_test_case_name not in name_to_test_case_result:
+                lines.append("nothing — no such test case result found")
+                break
+            lines.append(
+                format_test_result(
+                    name_to_test_case_result[expected_test_case_name]
+                ).format_human(LogColorProvider())
+            )
+            lines.append("")
+            break
```
Arcticae, line 123: this whole flow with breaks and continues is confusing. i don't know if it's that important though since it's only used for test output

tests/integration/conftest.py
```diff
@@ -81,9 +82,48 @@ def assert_cairo_test_cases(
...
+    }
+    diff = show_diff_between_cairo_test_cases(
+        name_to_test_case_result, expected, actual
+    )
+    if diff:
+        assert False, diff
     assert actual == expected
 
 
+def show_diff_between_cairo_test_cases(
```
Arcticae, line 98: this doesn't "show" anything. it formats the differences into a string

tests/integration/conftest.py
```diff
@@ -81,9 +82,48 @@ def assert_cairo_test_cases(
...
+        test_result.test_case_name: test_result
+        for test_result in testing_summary.test_results
+        if isinstance(test_result, TestCaseResult)
+    }
+    diff = show_diff_between_cairo_test_cases(
+        name_to_test_case_result, expected, actual
+    )
+    if diff:
+        assert False, diff
     assert actual == expected
```
Arcticae, line 95: isn't this redundant since we're asserting above?

Arcticae (approved):
A few nitpicks, do as you feel on those, i don't have strong feelings against merging this, since it's only test code.

tests/integration/conftest.py
```diff
@@ -81,9 +82,48 @@ def assert_cairo_test_cases(
...
+    name_to_test_case_result = {
+        test_result.test_case_name: test_result
+        for test_result in testing_summary.test_results
+        if isinstance(test_result, TestCaseResult)
+    }
+    diff = show_diff_between_cairo_test_cases(
+        name_to_test_case_result, expected, actual
+    )
+    if diff:
+        assert False, diff
```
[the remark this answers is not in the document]
kasperski95: I excluded diff from the condition on purpose, to avoid duplicating diff information.

tests/integration/conftest.py
kasperski95: I am not sure if the diff assertion catches all edge cases. I left the second assertion as a "safety net".

tests/integration/conftest.py
```diff
@@ -81,9 +82,48 @@ def assert_cairo_test_cases(
...
 
 
+def show_diff_between_cairo_test_cases(
+    name_to_test_case_result: dict[str, Any],
+    expected: CairoTestCases,
+    actual: CairoTestCases,
+) -> str:
+    lines: list[str] = []
+    result_types = ["passed", "failed", "broken", "skipped"]
+    for expected_result_type in result_types:
```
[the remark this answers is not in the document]
kasperski95: I'll change it in the next PR.

Issue #1389, opened 2023-01-18
Investigate using cheatable state directly in test execution state


Arcticae:
_Originally posted by @MaksymilianDemitraszek in https://github.com/software-mansion/protostar/pull/1368#discussion_r1073639770_

Arcticae:
Moving to later releases, since it's a low-priority refactor task

MaksymilianDemitraszek:
Probably not relevant, and unnecessary refactor

PR #1406, opened 2023-01-20
Setup cases for cairo runner

### LINKED ISSUE #1281: Implement `setup_<case>`s and `__setup__` for pure cairo VM runner

Arcticae:
Closes #1281


tests/integration/pure_cairo_vm/test_runner_test/suite_with_setup_case_test.cairo
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func test_b_doesnt_leak_from_a_setup(){
+    alloc_locals;
+    local contract_deployed;
+
+    %{
+        ids.contract_deployed = 1 if hasattr(context, "contract_address") else 0
+    %}
+    assert contract_deployed = 0;
```
kasperski95, line 33: Assertions could be in hints.

```suggestion
    %{ assert not hasattr(context, "contract_address") %}
```

protostar/cairo_testing/execution_environments/cairo_setup_case_execution_environment.py
```diff
@@ -0,0 +1,44 @@
...
+        super().__init__(state, program, self._get_hint_locals(state))
+
+    async def execute(self, function_name: str):
+        with self.state.output_recorder.redirect("setup case"):
+            await self.run_cairo_function(function_name)
+
+    def _get_hint_locals(self, state: CairoTestExecutionState) -> HintLocalsDict:
+        hint_locals: HintLocalsDict = {}
+        cheatcode_factory = CairoSetupCheatcodeFactory(  # For now there is no reason to create another factory file
+            cheatable_state=cast(CheatableCachedState, state.starknet.state.state),
```
kasperski95, line 32:
```suggestion
            cheatable_state=state.cheatable_state,
```
I think this cast should be moved to `CairoTestExecutionState`.

protostar/cairo_testing/execution_environments/cairo_setup_case_execution_environment.py
```diff
@@ -0,0 +1,44 @@
...
+    ):
+        super().__init__(state, program, self._get_hint_locals(state))
+
+    async def execute(self, function_name: str):
+        with self.state.output_recorder.redirect("setup case"):
+            await self.run_cairo_function(function_name)
+
+    def _get_hint_locals(self, state: CairoTestExecutionState) -> HintLocalsDict:
+        hint_locals: HintLocalsDict = {}
+        cheatcode_factory = CairoSetupCheatcodeFactory(  # For now there is no reason to create another factory file
```
kasperski95, line 31:
```suggestion
        cheatcode_factory = CairoSetupCheatcodeFactory(
```

kasperski95 (approved):

protostar/cairo_testing/execution_environments/cairo_setup_case_execution_environment.py
Arcticae: That is not possible ATM. It will be, when we pass `CheatableCachedState` in the `TestExecutionState` class instead of `Starknet`. (#1389)

tests/integration/pure_cairo_vm/test_runner_test/suite_with_setup_case_test.cairo
Arcticae: It will make up for more work when migrating those files to cairo 1.0 IMO, that's why i try to keep the code as cairo-centered as possible.

protostar/cairo_testing/execution_environments/cairo_setup_case_execution_environment.py
kasperski95:
```py
    @property
    def cheatable_state(self):
        return cast(CheatableCachedState, self.starknet.state.state)
```

Arcticae: Ah right, that one is possible.

PR #1411, opened 2023-01-24
Fix using contracts in tests with name shortcut

### LINKED ISSUE #1366: Declare using the contract name doesn't work
It doesn't work with the new arch. Particularly, there's no ABI after this: https://github.com/software-mansion/protostar/blob/68a316eefd5e06d1a26470a6caa30e3f2bd86f43/protostar/compiler/project_compiler.py#L80

Repro branch: https://github.com/software-mansion/protostar/tree/kb/declare-by-name-repro

I might've done something wrong on the way, but wanted to report this as it may be a problem.

I used `deploy_contract` but this essentially comes down to the `declare` cheat code.

Arcticae:
Fixes by doing 2 things:
- Using `ProtostarFixture` directly
- Declaring proper contract names using the same `ProtostarFixture`

Closes #1366


tests/integration/_conftest/protostar_fixture.py
```diff
@@ -32,8 +43,19 @@
...
 
 # pylint: disable=too-many-instance-attributes
+def to_contract_name(contract_identifier: str) -> str:
+    """
+    Converts identifier (path/ contract name) into contract name which can be put into protostar.toml
+    For example:
+        ./src/basic.cairo -> basic
+        basic -> basic
+    """
+    return Path(contract_identifier).stem
```
kasperski95, line 56: I think this code is unused.

tests/integration/_conftest/protostar_fixture.py
```diff
@@ -318,6 +340,95 @@ def create_files(
...
+            profile_name_to_project_config={},
+            profile_name_to_commands_config={},
+        )
+        content_factory = ConfigurationFileV2ContentFactory(
+            content_builder=ConfigurationTOMLContentBuilder()
+        )
+        file_content = content_factory.create_file_content(
+            overriden_config_file_model_v2
+        )
+        protostar_toml_path.write_text(file_content)
```
kasperski95, line 405: I struggle to understand why this code works. The configuration file is loaded when ProtostarFixture is created, so these changes shouldn't work, because the file is already cached.

kasperski95 (approved):

tests/integration/_conftest/protostar_fixture.py
Arcticae: ProjectCompiler is instantiated in the worker, also configuration file is created there, so in tests runs themselves we have always the most recent data

PR #1412, opened 2023-01-24
Cairo Cheatcode — Expect events

kasperski95:
Closes https://github.com/software-mansion/protostar/issues/1343


protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
```diff
@@ -59,3 +72,10 @@ def _get_hint_locals(self, state: CairoTestExecutionState) -> HintLocalsDict:
             hint_locals[custom_hint_local.name] = custom_hint_local.build()
 
         return hint_locals
+
+    def _wrap_hint_exceptions(self, ex: Exception):
+        if isinstance(ex, ExpectEventsMismatchException):
+            return ExpectEventsMismatchReportedException(
+                message=ex.message, event_matching_result=ex.matching_result
+            )
+        return ex
```
karol-bisztyga, line 81: Can you explain the purpose of this function? It looks totally redundant. What you do is you catch `ExpectEventsMismatchException` and then you conditionally re-raise either `ExpectEventsMismatchException` or `ExpectEventsMismatchReportedException` if the original error is `ExpectEventsMismatchException`, which happens in 100% situations.

protostar/cairo_testing/cheatcode_factories/cairo_test_cheatcode_factory.py
```diff
@@ -40,18 +44,26 @@
...
 from protostar.cheatable_starknet.cheaters.storage import StorageCairoCheater
+from protostar.testing import Hook
 
 
 class CairoTestCheatcodeFactory:
     def __init__(
         self,
         cheatable_starknet_facade: CheatableStarknetFacade,
         project_compiler: ProjectCompiler,
+        test_finish_hook: Hook,
```
karol-bisztyga, line 60: good idea with the hook 👍

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -0,0 +1,142 @@
...
+        try:
+            event_matchings.append(
+                match_expected_and_emitted_event(
+                    emitted_event=emitted_event,
+                    expected_event=expected_events_queue[0],
+                    on_accepted=expected_events_queue.popleft,
+                )
+            )
+        except IndexError:
+            event_matchings.append(SkippedEventMatching(emitted_event=emitted_event))
```
karol-bisztyga, line 105: I'd rather see a check for a list's item but it's a NIT, for me it just would be cleaner.

karol-bisztyga (changes requested):
For me this whole structure you built here seems like an overkill.

I don't see why we need 2 distinct types for errors: `ExpectEventsMismatchException` and `ExpectEventsMismatchReportedException`; as well as such extensive logic for the controller while all we need is to check whether all expected events have been called.

Could you specify why we need this complicated structure instead of a few simple checks and one type of error that would just print a message what went wrong?

protostar/cairo_testing/cairo_test_execution_state.py
```diff
@@ -22,6 +25,7 @@ class CairoTestExecutionState:
     context: TestContext
     config: TestConfig
     project_compiler: ProjectCompiler
+    event_expectations: list[EventsExpectation] = field(default_factory=list)
```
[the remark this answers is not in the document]
kasperski95: event expectations ==> `list[list[Event]]`

protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
kasperski95: Using `ExpectEventsMismatchReportedException` in the controller would create a circular dependency.

protostar/cheatable_starknet/cheaters/expect_events_controller.py
Arcticae: IndexError might be raised from `match_expected_and_emitted_event` method, it's not a nit @karol-bisztyga  :)

protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
Arcticae: why not just "ReportedException" ? We don't seem to catch it anywhere anyways

protostar/cheatable_starknet/cheatables/cheatable_starknet_facade.py
```diff
@@ -72,10 +71,11 @@ async def invoke(
             entry_point_selector=get_selector_from_name(function_name),
         )
         with self.cheatable_state.copy_and_apply() as state_copy:
-            await entry_point.execute_for_testing(
+            call_info = await entry_point.execute_for_testing(
                 state=state_copy,
                 general_config=StarknetGeneralConfig(),
             )
+            self._starknet.state.events += call_info.get_sorted_events()
```
Arcticae, line 78: this seems sketchy to me, very not obvious what happens here and why

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -0,0 +1,142 @@
...
+            )
+            if not matching_result.should_be_accepted:
+                raise ExpectEventsMismatchException(matching_result)
+
+
+def match_events(
+    expected_events: list[Event], emitted_events: list[Event]
+) -> EventMatchingResult:
+    assert len(expected_events) > 0
+    expected_events_queue = deque(expected_events)
```
Arcticae, line 93: I don't think we need a queue here

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -0,0 +1,142 @@
...
+
+def match_expected_and_emitted_event(
+    emitted_event: Event,
+    expected_event: Event,
+    on_accepted: Callable,
+):
+    if should_accept_event_matching(
+        expected_event=expected_event, emitted_event=emitted_event
+    ):
+        on_accepted()
```
Arcticae, line 123: this is not very pythonic solution

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__(cheaters)
+        self._controller = controller
+
+    @property
+    def name(self) -> str:
+        return "expect_events"
+
+    def _build(self) -> Callable:
+        def expect_events(
+            *raw_expected_events: RawExpectedEvent,
```
Arcticae, line 32: this input should be validated

Arcticae (changes requested):

protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
kasperski95: Ok, I'll change that. I thought `ReportedException` is in the `protostar.testing` (which actually would make more sense) not in the `protostar.starknet`. Circular dependency won't be created.

karol-bisztyga: That doesn't change the fact that you can simply do
```
except ExpectEventsMismatchException as ex:
                raise ExpectEventsMismatchReportedException(...)
```
in `execute_test_case` and remove `_wrap_hint_exceptions`

protostar/cairo_testing/cairo_test_execution_state.py
```diff
@@ -44,4 +48,5 @@ def fork(self) -> Self:
             output_recorder=self.output_recorder.fork(),
             stopwatch=self.stopwatch.fork(),
             cheatable_starknet_facade=self.cheatable_starknet_facade.fork(),
+            event_expectations=self.event_expectations.copy(),
```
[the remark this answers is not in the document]
kasperski95: Theoretically, someone could use expect events in "setup", if we plan to make this hint available there.

protostar/cairo_testing/cheatcode_factories/cairo_test_cheatcode_factory.py
```diff
@@ -101,4 +113,12 @@ def build_cheatcodes(self) -> List[CairoCheatcode]:
...
                 cheaters=cheaters,
             ),
+            ExpectEventsCheatcode(
+                cheaters=cheaters,
+                controller=ExpectEventsController(
+                    test_finish_hook=self._test_finish_hook,
+                    test_execution_state=self._test_execution_state,
+                    cheatable_starknet_facade=self._cheatable_starknet_facade,
+                ),
+            ),
```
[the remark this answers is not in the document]
kasperski95: How does it collide?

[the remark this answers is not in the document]
Arcticae: Cheaters are essentially controllers, so that will need to be incorporated into `Controllers`

protostar/cairo_testing/cairo_test_execution_state.py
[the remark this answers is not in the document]
Arcticae: I don't think we should allow that TBH, even if we would, i think the lifespan of the expectation should be the scope of the test/setup function. Otherwise it would make some bizzare scenarios. 

It's up for discussion though.

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
kasperski95: It is in the method below.

protostar/cairo_testing/cheatcode_factories/cairo_test_cheatcode_factory.py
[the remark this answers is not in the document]
Arcticae: It seems easier to construct them once, and let the cheatcodes use whichever one they need. 

> A cheat code uses one controller

I'm not completely sure that's always the case

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
Arcticae: No it's not, 
`raw_expected_event["name"]`
`raw_expected_event["from_address"]`
those will raise `KeyError`

protostar/cairo_testing/cheatcode_factories/cairo_test_cheatcode_factory.py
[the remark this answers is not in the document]
Arcticae: Anyways it's not the subject of this PR anymore, resolving this

protostar/cheatable_starknet/cheaters/expect_events_controller.py
kasperski95: [use list](https://github.com/software-mansion/protostar/pull/1412/commits/e4176a011ef70b764c23617a2f857b2ed8291a92)

protostar/cheatable_starknet/cheaters/expect_events_controller.py
kasperski95: [simplify match_events](https://github.com/software-mansion/protostar/pull/1412/commits/0a93f31859a66511d69a776cc6b4edb15bb0ab46)

protostar/cairo_testing/execution_environments/cairo_test_execution_environment.py
kasperski95: [use one ExpectEventsMismatch exception](https://github.com/software-mansion/protostar/pull/1412/commits/d2f9490978db998738a2896fc8ba6232831a47b3)

protostar/cairo_testing/cairo_test_execution_state.py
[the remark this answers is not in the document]
kasperski95: [use expected events list](https://github.com/software-mansion/protostar/pull/1412/commits/57e932ed8262333e0134a7e76034d376f109637c)

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
Arcticae: Good point

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -85,24 +84,22 @@ def compare_expected_and_actual_results(self):
...
         else:
             event_matchings.append(
                 AcceptedEventMatching(
                     emitted_event=emitted_event,
-                    expected_event=expected_events_queue[0],
+                    expected_event=expected_events[0],
                 )
             )
-            expected_events_queue.popleft()
+            expected_events.pop(0)
```
Arcticae, line 100: Nit: Not that i care much about that, but this one is O(n) in python (for lists)

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -85,24 +84,22 @@ def compare_expected_and_actual_results(self):
...
     failed_event_matchings = [
-        FailedEventMatching(expected_event)
-        for expected_event in list(expected_events_queue)
+        FailedEventMatching(expected_event) for expected_event in list(expected_events)
```
Arcticae, line 102: `expected_events` is already a list

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -85,24 +84,22 @@ def compare_expected_and_actual_results(self):
 def match_events(
     expected_events: list[Event], emitted_events: list[Event]
 ) -> EventMatchingResult:
-    expected_events_queue = deque(expected_events)
     event_matchings: list[EventMatching] = []
     for emitted_event in emitted_events:
-        if len(expected_events_queue) == 0 or not should_accept_event_matching(
-            expected_event=expected_events_queue[0], emitted_event=emitted_event
+        if len(expected_events) == 0 or not should_accept_event_matching(
```
Arcticae, line 89: `len(expected_events) == 0` is simplifiable to `not expected_events`

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+            expected_events.pop(0)
+    failed_event_matchings = [
+        FailedEventMatching(expected_event) for expected_event in list(expected_events)
+    ]
+    return EventMatchingResult(
+        event_matchings=[*event_matchings, *failed_event_matchings],
+    )
+
+
+def should_accept_event_matching(expected_event: Event, emitted_event: Event) -> bool:
```
Arcticae, line 109: `events_match` would be a nicer name

Arcticae: Or even better, we could add a `.matches()` method to the `Event` dataclass

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+                raise ExpectEventsMismatchReportedException(matching_result)
+
+
+def match_events(
+    expected_events: list[Event], emitted_events: list[Event]
+) -> EventMatchingResult:
+    event_matchings: list[EventMatching] = []
+    for emitted_event in emitted_events:
+        if len(expected_events) == 0 or not should_accept_event_matching(
+            expected_event=expected_events[0], emitted_event=emitted_event
```
Arcticae, line 90: we should probably extract `expected_events[0]` to the variable. And is there any reason we get the first element?

Arcticae (approved):
It's mergable IMO - some style suggestions/questions, but those are left up for you to decide.

protostar/cheatable_starknet/cheaters/expect_events_controller.py
kasperski95: I changed it to the list because you requested it.

> I don't think we need a queue here

I am leaving as it is. This list won't contain many items anyway.

protostar/cheatable_starknet/cheaters/expect_events_controller.py
kasperski95: I forgot to change it after removing the queue.

[remove unnecessary converting to list](https://github.com/software-mansion/protostar/pull/1412/commits/2c6f9de34f0d7eb00c54d18726b4ae3b9a64e070)_

protostar/cheatable_starknet/cheaters/expect_events_controller.py
kasperski95: I'd avoid adding it to the data class because it's not obvious whether expected_event should match emitted_event or emitted_event should match expected_event.

protostar/cheatable_starknet/cheaters/expect_events_controller.py
kasperski95: The first element is extracted because the first condition ensures that this list contains at least one element. I didn't extract it to a variable because I'd have to create another `if`.

protostar/cheatable_starknet/cheaters/expect_events_controller.py
kasperski95: I'd keep this name for the consistency reasons — there's also`EventMatching` data structure.

protostar/cheatable_starknet/cheaters/expect_events_controller.py
kasperski95: I chose to use `len` variant, to highlight the existence of `expected_events_queue[0]` in the next condition.

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
```diff
@@ -0,0 +1,54 @@
...
+class RawExpectedEvent(TypedDict):
+    from_address: int
+    name: str
+    data: NotRequired[CairoData]
+
+
+class ExpectEventsCheatcode(CairoCheatcode):
+    def __init__(self, controller: ExpectEventsController, cheaters: "CairoCheaters"):
+        super().__init__(cheaters)
+        self._controller = controller
```
MaksymilianDemitraszek, line 24: I don't think we should pass this one as argument atm, there is no reason to do so. Can we just create it inside cheatcode?

protostar/cheatable_starknet/cheaters/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+
+@dataclass
+class Event:
+    from_address: Address
+    key: Selector
+    data: Optional[CairoData] = None
+
+
+@dataclass
+class EventMatching:
```
MaksymilianDemitraszek, line 22: This should be Union type instead

kasperski95: Why? With this approach, I don't have to remember to "register" or "unregister" concrete `EventMatching`. This is also consistent with `test_results.py`.

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
kasperski95: No. `ExpectEventsCheatcode` would have inherited `ExpectEventsController` dependencies.

protostar/cheatable_starknet/cheaters/expect_events_controller.py
MaksymilianDemitraszek: For me it seems more natural to represent composite types as union instead of subtyping, this should be marked as [[Nitpick]]

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
MaksymilianDemitraszek: You are right, but I am concerned that we have inconsistency between cheatcodes (which lead to my confusion), other cheatcodes are using cheaters to pass those dependencies. Can we create a task for removing Cheaters and consolidating it to your approach?

kasperski95: Do we need to create tasks for refactoring? It's hard to forget about it.

Arcticae: It doesn't hurt to have it

tests/integration/pure_cairo_vm/cheatcodes/expect_events/expect_event_cairo_cheatcode_test.py
```diff
@@ -0,0 +1,53 @@
...
+from protostar.testing.test_results import TestCaseResult
+from protostar.testing.testing_summary import TestingSummary
+from tests.integration.conftest import (
+    CreateProtostarProjectFixture,
+    assert_cairo_test_cases,
+)
+from tests.integration._conftest import ProtostarFixture
+from tests.integration.pure_cairo_vm.conftest import CONTRACTS_PATH
+
+TEST_PATH = Path(__file__).parent / "expect_events_test.cairo"
```
Arcticae, line 14: That variable is redundant

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+                raise ExpectEventsMismatchReportedException(matching_result)
+
+
+def match_events(
+    expected_events: list[Event], emitted_events: list[Event]
+) -> EventMatchingResult:
+    event_matchings: list[EventMatching] = []
+    for emitted_event in emitted_events:
+        if len(expected_events) == 0 or not should_accept_event_matching(
+            expected_event=expected_events[0], emitted_event=emitted_event
```
Arcticae, line 90: this is still using queue-style logic

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+    def failed_event_matchings(self):
+        return [
+            matching
+            for matching in self.event_matchings
+            if isinstance(matching, FailedEventMatching)
+        ]
+
+    @property
+    def should_be_accepted(self) -> bool:
+        return len(self.failed_event_matchings) == 0
```
Arcticae, line 56:
```suggestion
        return not self.failed_event_matchings
```

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+            if not matching_result.should_be_accepted:
+                raise ExpectEventsMismatchReportedException(matching_result)
+
+
+def match_events(
+    expected_events: list[Event], emitted_events: list[Event]
+) -> EventMatchingResult:
+    event_matchings: list[EventMatching] = []
+    for emitted_event in emitted_events:
+        if len(expected_events) == 0 or not should_accept_event_matching(
```
Arcticae, line 89:
```suggestion
        if not expected_events or not should_accept_event_matching(
```

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+@dataclass
+class EventMatching:
+    pass
+
+
+@dataclass
+class SkippedEventMatching(EventMatching):
+    emitted_event: Event
```
Arcticae, line 28: Do we need to acknowledge they are even present? The only thing we're interested in are failed matchings (meaning ones that are expected, or not there, or mismatched messages)

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+        self._test_execution_state = test_execution_state
+        self._test_finish_hook = test_finish_hook
+        self._cheatable_state = cheatable_state
+        self._test_finish_hook.on(self.compare_expected_and_actual_results)
+
+    def add_expected_events(self, expected_events: list[Event]):
+        self._test_execution_state.expected_events_list.append(expected_events)
+
+    def compare_expected_and_actual_results(self):
+        for expected_events in self._test_execution_state.expected_events_list:
```
Arcticae, line 75: I have trouble understanding why this list is not flattened, why is that the case?

Arcticae (approved):
Seems okay logic-wise, some nitpicks below

protostar/cheatable_starknet/controllers/expect_events_controller.py
kasperski95: We need to tell the user what events were actually emitted. Without that, it's hard to say what went wrong.

protostar/cheatable_starknet/controllers/expect_events_controller.py
kasperski95: Because a user can call `expect_events` multiple times, with many events. 

```python
expect_events(event_A, event_B) # order matters, A must be before B
expect_events(event_C)
```

protostar/cheatable_starknet/controllers/expect_events_controller.py
kasperski95: I don't know what you want me to change.

tests/integration/pure_cairo_vm/cheatcodes/expect_events/expect_event_cairo_cheatcode_test.py
kasperski95: [remove constant](https://github.com/software-mansion/protostar/pull/1412/commits/2222f1c7ac8d829fc9f8a4a2f57d742a9e91a57d)

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+    def failed_event_matchings(self):
+        return [
+            matching
+            for matching in self.event_matchings
+            if isinstance(matching, FailedEventMatching)
+        ]
+
+    @property
+    def should_be_accepted(self) -> bool:
+        return not self.failed_event_matchings
```
karol-bisztyga, line 56:
```suggestion
@dataclass
class EventMatchingResult:
    accepted_event_matchings: list[AcceptedEventMatching]
    failed_event_matchings: list[FailedEventMatching]
    skipped_event_matchings: list[SkippedEventMatching]

    @property
    def should_be_accepted(self) -> bool:
        return not self.failed_event_matchings
```

How about storing specific results in separate collections rather than querying them all the time like this and spawning new comprehension lists?

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def _format_accepted_event_matching(self, event_matching: AcceptedEventMatching):
+        return self._format_event(event=event_matching.emitted_event, prefix="[pass]")
+
+    def _format_failed_event_matching(self, event_matching: FailedEventMatching):
+        return self._format_event(event=event_matching.expected_event, prefix="[fail]")
+
+    def _format_skipped_event_matching(self, event_matching: SkippedEventMatching):
+        return self._format_event(event=event_matching.emitted_event, prefix="[skip]")
+
```
karol-bisztyga, line 158:
```suggestion
    def _get_event_matches_formatted_lines(self):
        lines: list[str] = []
        for event_matching in self.event_matching_result.event_matchings:
            if isinstance(event_matching, AcceptedEventMatching):
                lines.append(self._format_event(event=event_matching.emitted_event, prefix="[pass]"))
            elif isinstance(event_matching, SkippedEventMatching):
                lines.append(self._format_event(event=event_matching.emitted_event, prefix="[skip]"))
            elif isinstance(event_matching, FailedEventMatching):
                lines.append(self._format_event(event=event_matching.expected_event, prefix="[fail]"))
            else:
                assert False, f"Unexpected event_matching: {event_matching}"
        return lines
```
redundant methods:
- `_format_accepted_event_matching`
- `_format_failed_event_matching`
- `_format_skipped_event_matching`

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+    ]
+    return EventMatchingResult(
+        event_matchings=[*event_matchings, *failed_event_matchings],
+    )
+
+
+def should_accept_event_matching(expected_event: Event, emitted_event: Event) -> bool:
+    return (
+        expected_event.key == emitted_event.key
+        and (expected_event.data is None or (expected_event.data == emitted_event.data))
```
karol-bisztyga, line 112: Hmm so if we expect no data but receive some data then we just ignore this data and consider this correct, right? It feels weird, afaik, for the expected call we went for the strict match, why wouldn't we do the same here?

protostar/cheatable_starknet/controllers/expect_events_controller.py
```diff
@@ -0,0 +1,166 @@
...
+            if not matching_result.should_be_accepted:
+                raise ExpectEventsMismatchReportedException(matching_result)
+
+
+def match_events(
+    expected_events: list[Event], emitted_events: list[Event]
+) -> EventMatchingResult:
+    event_matchings: list[EventMatching] = []
+    for emitted_event in emitted_events:
+        if not expected_events or not should_accept_event_matching(
```
karol-bisztyga, line 89: Maybe when this is correct: `if not expected_events`, we could create the `event_matchings` as a list of `SkippedEventMatching` before the loop using comprehension? Would be far cleaner I think.

protostar/cheatable_starknet/controllers/expect_events_controller.py
kasperski95: If data is not specified, then data is not compared. However, a user can check if there's no data was emitted by providing an empty list. This approach gives the user a choice.

protostar/cheatable_starknet/controllers/expect_events_controller.py
Arcticae: hmm honestly i think it's fine, i think i misunderstood the logic behind it

protostar/cheatable_starknet/controllers/expect_events_controller.py
kasperski95: I wouldn't say it cleaner. The loop must be in `else` scope, which would create 3 level nesting. I removed negations from the predicate to make it easier to reason about.

[remove negations](https://github.com/software-mansion/protostar/pull/1412/commits/5a57a99a499d2582c6faa4b6bcad266f741a3cc5)

protostar/cheatable_starknet/controllers/expect_events_controller.py
kasperski95: We need to maintain the order to display matchings correctly.

protostar/cheatable_starknet/controllers/expect_events_controller.py
kasperski95: [remove functions](https://github.com/software-mansion/protostar/pull/1412/commits/4b29d7257b3235d3c6dc1ef9943d35230208f526)

protostar/cheatable_starknet/controllers/expect_events_controller.py
karol-bisztyga: I understand, it looks better now.

protostar/cheatable_starknet/controllers/expect_events_controller.py
karol-bisztyga: I understand, it feels like you basically explained what we both know. I still don't understand why we don't go for a "strict" match here (so if the emitted data is different than expected, it will be a fail for sure) and instead we use the "loose" match (if the expected emitted data is different than the actual emitted data, it's not always a fail).

kasperski95: To give a user a choice. If he wants to compare the data, he will provide the data.

@Arcticae WDYT

Arcticae: Seems reasonable to me to have no expectations from the data if the user doesn't

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
MaksymilianDemitraszek: We definiately need to for parts which we decided to refactor. We are using tasks as async communication mechanism. Keeping stuff as tribe knowledge is always a bad idea

protostar/cheatable_starknet/controllers/expect_events_controller.py
karol-bisztyga: Ok, resolving then.

karol-bisztyga (approved):

protostar/cheatable_starknet/cheatcodes/expect_events_cairo_cheatcode.py
Arcticae: https://github.com/software-mansion/protostar/issues/1440

protostar/cairo_testing/cairo_test_execution_state.py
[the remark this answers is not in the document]
Arcticae: Transferring this discussion to private. Will merge as-is, we can follow-up later, this PR has been up for long enough, and i figure the code is backwards-compatible as-is

PR #1413, opened 2023-01-25
Move cheatables to `cheatables` directory

kasperski95:
Move cheatable files to cheatables directory. Current structure implicitly suggests, that Cheatables use Cheaters (Controller) which shouldn't be a case.


kasperski95:
Closing. Moved again on a different more carefully to avoid creating cyclic imports.

PR #1417, opened 2023-01-25
Remove dead code (cheaters)

kasperski95:
Remove dead code.

### DIFF
--- a/protostar/cheatable_starknet/cheatables/cheatable_entry_point.py
+++ b/protostar/cheatable_starknet/cheatables/cheatable_entry_point.py
@@ -2,7 +2,7 @@
 import logging
 import re
 from dataclasses import dataclass
-from typing import Optional, Tuple, cast, List, TYPE_CHECKING, Any
+from typing import Optional, Tuple, cast, List, TYPE_CHECKING
 from copy import deepcopy
 
 from starkware.cairo.common.cairo_function_runner import CairoFunctionRunner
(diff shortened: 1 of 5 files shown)


protostar/cheatable_starknet/cheatables/cheatable_entry_point.py
```diff
@@ -55,37 +55,27 @@
 
 @dataclass(frozen=True)
 class CheatableExecuteEntryPoint(ExecuteEntryPoint):
-    cheaters: "CairoCheaters"
     max_steps: Optional[int] = None
     "``None`` means default Cairo value."
 
     @classmethod
-    def create_with_cheaters(
+    def create_for_protostar(
```
Arcticae, line 62: this method can be removed as well

Arcticae (approved):

protostar/cheatable_starknet/cheatables/cheatable_entry_point.py
kasperski95: This method actually has some purpose. Callers don't have to specify every `ExecuteEntryPoint` constructor parameter — sensible default are used here. Also, this method can use `Address` and `Selector` types to avoid converting every time `ExecuteEntryPoint` is created.

Arcticae: If you feel like it's useful - you can leave it, but the same defaults are in `create`. Please reuse that method then in this one

PR #1418, opened 2023-01-26
CheatableStarknetFacade

kasperski95:
Move Starknet's implementation details in `CheatableStarknetFacade`. Simplify `CairoTestExecutionState` and `ContractsCheater`. Code responsible for deploying and declaring wasn't moved in this PR. I plan to use `CheatableStarknetFacade` in PR with Expect Events.

### DIFF
--- a/protostar/cairo_testing/cairo_test_execution_state.py
+++ b/protostar/cairo_testing/cairo_test_execution_state.py
@@ -1,25 +1,13 @@
 import dataclasses
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import cast
 
-from starkware.crypto.signature.fast_pedersen_hash import pedersen_hash_func
-from starkware.starknet.business_logic.fact_state.patricia_state import (
-    PatriciaStateReader,
-)
-from starkware.starknet.business_logic.fact_state.state import SharedState
-from starkware.starknet.business_logic.state.state_api_objects import BlockInfo
-from starkware.starknet.definitions.general_config import StarknetGeneralConfig
-from starkware.starknet.testing.starknet import Starknet
-from starkware.starknet.testing.state import StarknetState
-from starkware.storage.dict_storage import DictStorage
-from starkware.storage.storage import FactFetchingContext
 from typing_extensions import Self
 
-from protostar.compiler import ProjectCompiler
-from protostar.cheatable_starknet.cheatables.cheatable_cached_state import (
-    CheatableCachedState,
+from protostar.cheatable_starknet.cheatables.cheatable_starknet_facade import (
+    CheatableStarknetFacade,
 )
+from protostar.compiler import ProjectCompiler
 from protostar.testing.stopwatch import Stopwatch
 from protostar.testing.test_config import TestConfig
 from protostar.testing.test_context import TestContext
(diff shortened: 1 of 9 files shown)


Arcticae (commented):
I don't really understand why we need this abstraction to be present, and moreover it's coupling the starknet interaction code with starknet object itself (by inhertiance), which seems sub-optimal in terms of future code growth.

Can you please explain the reasoning? I think it'd be feasible to extract cheatables' interaction with starknet to a separate module, just i don't think that's the correct way of doing it.

kasperski95:
> I don't really understand why we need this abstraction to be present. it's coupling the starknet interaction code with starknet object itself (by inhertiance), which seems sub-optimal in terms of future code growth.

I am confused. I don't see the inheritance.

> Can you please explain the reasoning? I think it'd be feasible to extract cheatables' interaction with starknet to a separate module, just i don't think that's the correct way of doing it.

This class hides Starknet's internals from Protostar. Protostar shouldn't know about these classes:
- `StarknetGeneralConfig`
- `FactFetchingContext`
- `DictStorage`
- `SharedState`
- `PatriciaStateReader`
- `ExecuteEntryPoint` 

Also, it provides Starknet's interface made for Protostar needs. In Expect Events PR I need to get emitted events. This data isn't stored in CachedState, and I don't want to build code on top of unstable Starknet internals, especially because a refactoring in planned on the Starknet side.

Arcticae:
> This class hides Starknet's internals from Protostar
doesn't `CairoTestExecutionState` do the same thing ATM? This is a stretch IMO

> and I don't want to build code on top of unstable Starknet internals
well that's what we mainly do here 😬 also if extracted to a controller (cheater ATM), and passing the StarkNet object, it'd fit into our current design nicely, without the need to compilcate things

kasperski95:
> doesn't CairoTestExecutionState do the same thing ATM? This is a stretch IMO

`CairoTestExecutionState`, as the name suggest, keeps test execution state.

> well that's what we mainly do here 😬 also if extracted to a controller (cheater ATM), and passing the StarkNet object, it'd fit into our current design nicely, without the need to compilcate things

I don't get this part. `CheatableStarknetFacade` would focus on providing `Starknet` interface for Protostar needs, Contracts Cheater/Controller would focus on business logic. The class is now big enough to justify separation.

Arcticae:
> CairoTestExecutionState, as the name suggest, keeps test execution state.

well yes, but it contains the knowledge on how to build the new state as well. that seems reasonable to me. 

> CheatableStarknetFacade would focus on providing Starknet interface for Protostar needs, Contracts Cheater/Controller 
would focus on business logic. The class is now big enough to justify separation

Yes i agree, but it doesn't conform to the design we agreed upon, in which controllers would be divided by the technical aspect (transactions, expectations, state manipulation, etc..). It makes us pass yet another weird object around which contains state and some operations directly on that state (which current cheatcodes don't do - they can be treated as a function). 

I just don't get the need for a separate abstraction over "starknet" since we established kind of "MVC" here -
- `HintLocal` as a view
- `Controllers` (previous cheaters) as a (stateless) controller
- `CheatableCachedState`/ `StarknetState` as  a model

This adds another concept which is not optimal for understanding the architecture and what belongs where

Arcticae:
You also said one time that you wanted to avoid dividing things by the domain, and do the division instead by a technical concept - `StarknetFacade` is the opposite of that in my understanding

MaksymilianDemitraszek (changes requested):
I appreciate the proactive effort to simpIify the codebase, but unfortunately have to agree with @Arcticae on this.

Facades are useful when you use only a subset of logic from certain code and want to hide some details, which is not our use-case. In our scenario we just creating a godclass which is what we exactly want to avoid with the new architecture (we wanted to decompose the responsibility of previous StarkNet class into smaller chunks). The logic you implement in facade in the next PR should be a responsibility of ExpectController. Can we restrain from such change? I don't say changes in this PR are bad, I can see the benefits from such modifications. Those are just not alligned with our current goals.

PR #1423, opened 2023-01-30
Fix invalid type in parser test

karol-bisztyga:


Arcticae (changes requested):
The correct solution here would be to annotate the `parse_test_suite` with `str` instead of `Any`

karol-bisztyga:
> The correct solution here would be to annotate the `parse_test_suite` with `str` instead of `Any`

That's another thing, thanks! Added this in [addr fdbck](https://github.com/software-mansion/protostar/pull/1423/commits/e31aa4ef9d43a6fe0601d1dbb1418107eab44ff6)

Arcticae (approved):

PR #1426, opened 2023-01-31
Use shared cheatcode factory

kasperski95:
Use `CairoSharedCheatcodeFactory` to avoid duplicating code. Move factories to one file.


Arcticae (approved):
A thought: 

I am wondering if we should remove other factories overall - doesn't seem like we need cheatcodes to be separated (for now). And if it's not needed now, it'd be better to remove them, and just add them later. 

This is mergable though, LGTM

kasperski95:
> I am wondering if we should remove other factories overall - doesn't seem like we need cheatcodes to be separated (for now). And if it's not needed now, it'd be better to remove them, and just add them later.

I was thinking about it. We may need this distinction if we decide to reimplement fuzzing.

PR #1429, opened 2023-02-01
`stop_warp` & `stop_roll` cheatcodes

### LINKED ISSUE #1335: Implement `stop_warp` and `stop_roll`

Arcticae:
Closes #1335

### DIFF
--- a/protostar/cheatable_starknet/callable_hint_locals/roll_hint_local.py
+++ b/protostar/cheatable_starknet/callable_hint_locals/roll_hint_local.py
@@ -18,7 +18,7 @@ class RollHintLocal(CallableHintLocal):
         self,
         target_contract_address: int,
         blk_number: int,
-    ):
+    ) -> None:
         self.controllers.block_info.cheat(
             contract_address=Address.from_user_input(target_contract_address),
             block_number=blk_number,
--- a/protostar/cheatable_starknet/callable_hint_locals/roll_hint_local.py
+++ b/protostar/cheatable_starknet/callable_hint_locals/roll_hint_local.py
@@ -19,7 +19,7 @@ class RollHintLocal(CallableHintLocal):
         target_contract_address: int,
         blk_number: int,
     ) -> None:
-        self.controllers.block_info.cheat(
+        self.controllers.block_info.set_block_number(
             contract_address=Address.from_user_input(target_contract_address),
             block_number=blk_number,
         )
--- a/protostar/cheatable_starknet/callable_hint_locals/roll_hint_local.py
+++ b/protostar/cheatable_starknet/callable_hint_locals/roll_hint_local.py
@@ -3,7 +3,7 @@ from typing import Any, Callable
 from protostar.cheatable_starknet.callable_hint_locals.callable_hint_local import (
     CallableHintLocal,
 )
-from protostar.starknet import RawAddress, Address
+from protostar.starknet import Address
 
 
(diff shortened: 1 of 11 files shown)


karol-bisztyga:
This looks ok but I'm wondering:
- what's wrong with the idea of returning the stop callback from the original cheat code?
- are we going to split the logic for other cheat codes like expect call/event as well?

Arcticae:
> * what's wrong with the idea of returning the stop callback from the original cheat code?

it's not going to be doable in cairo 1.0, since there are no hints

> * are we going to split the logic for other cheat codes like expect call/event as well?

yes, that is the idea. i mentioned somewhere a logic like `setTimeout`/`clearTimeout` - should be pretty similar (returning an id, and using separate function to cancel)

protostar/cheatable_starknet/callable_hint_locals/roll_hint_local.py
```diff
@@ -16,10 +16,10 @@ def _build(self) -> Callable[..., Any]:
 
     def roll(
         self,
-        target_contract_address: RawAddress,
+        target_contract_address: int,
         blk_number: int,
-    ) -> Callable[[], None]:
-        return self.controllers.block_info.cheat(
+    ):
```
karol-bisztyga, line 21:
```suggestion
    ) -> None:
```
nit, maybe it's not the most important thing in the universe but removing the return type just feels like a step back.

protostar/cheatable_starknet/callable_hint_locals/warp_hint_local.py
```diff
@@ -16,10 +14,10 @@ def _build(self):
 
     def warp(
         self,
-        target_contract_address: RawAddress,
+        target_contract_address: int,
         blk_timestamp: int,
-    ) -> Callable[[], None]:
-        return self.controllers.block_info.cheat(
+    ):
+        self.controllers.block_info.cheat(
```
karol-bisztyga, line 20: Why is this method called `cheat`? Couldn't we use something like `set_block_timestamp` + `set_block_number` instead?

protostar/cheatable_starknet/callable_hint_locals/warp_hint_local.py
```diff
@@ -16,10 +14,10 @@ def _build(self):
 
     def warp(
         self,
-        target_contract_address: RawAddress,
+        target_contract_address: int,
         blk_timestamp: int,
-    ) -> Callable[[], None]:
-        return self.controllers.block_info.cheat(
+    ):
```
karol-bisztyga, line 19:
```suggestion
    ) -> None:
```
nit, maybe it's not the most important thing in the universe but removing the return type just feels like a step back.

karol-bisztyga (approved):
Ok, cool, thanks for the answer, this code LGTM mostly, maybe it'd be good to get one more review for this.

protostar/cheatable_starknet/callable_hint_locals/warp_hint_local.py
Arcticae: It's a legacy thing, it was copied over. If you feel like it, we can change the structure in this PR.

Arcticae: https://github.com/software-mansion/protostar/pull/1429/commits/3f458f11e2f42aa50d0590e2b5ecb963d10185e1

PR #1451, opened 2023-02-07
Refactor test state related code

kasperski95:
Closes: https://github.com/software-mansion/protostar/issues/1389

### DIFF
--- a/protostar/cairo_testing/cairo_test_execution_state.py
+++ b/protostar/cairo_testing/cairo_test_execution_state.py
@@ -36,12 +36,14 @@ class ContractsControllerState:
         self,
         emitted_events: Optional[list[Event]] = None,
         class_hash_to_contract_abi: Optional[dict[ClassHashType, AbiType]] = None,
+        event_key_to_event_selector: Optional[dict[int, Selector]] = None,
         event_selector_to_event_abi: Optional[dict[Selector, AbiType]] = None,
         address_to_class_hash: Optional[dict[Address, ClassHashType]] = None,
         target_address_to_pranked_address: Optional[dict[Address, Address]] = None,
     ) -> None:
         self._emitted_events = emitted_events or []
         self._class_hash_to_contract_abi = class_hash_to_contract_abi or {}
+        self._event_key_to_event_selector = event_key_to_event_selector or {}
         self._event_selector_to_event_abi = event_selector_to_event_abi or {}
         self._contract_address_to_class_hash = address_to_class_hash or {}
         self._target_address_to_pranked_address = (
@@ -67,6 +69,12 @@ class ContractsControllerState:
     ) -> AbiType:
         class_hash = self._contract_address_to_class_hash[contract_address]
         return self._class_hash_to_contract_abi[class_hash]
+
+    def get_event_selector_from_event_key(self, key: int) -> Selector:
+        return self._event_key_to_event_selector[key]
+
+    def bind_event_key_to_event_selector(self, key: int, event_selector: Selector):
+        self._event_key_to_event_selector[key] = event_selector
 
     def bind_contract_address_to_class_hash(
         self, address: Address, class_hash: ClassHashType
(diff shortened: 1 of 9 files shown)


protostar/starknet/__init__.py
```diff
@@ -1,3 +1,5 @@
+from starkware.starknet.business_logic.state.state import CachedState as StarknetState
```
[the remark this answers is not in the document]
Arcticae: Moreover it may be confused with `starkware.starknet.testing.state.StarknetState` which is a totally different thing

protostar/cheatable_starknet/cheatables/cheatable_syscall_handler.py
```diff
@@ -31,31 +41,68 @@ def __init__(self, message: str):
 
 
 class CheatableSysCallHandler(BusinessLogicSysCallHandler):
-    def __init__(self, state: StateSyncifier, **kwargs: Any):
-        self.cheatable_state: "CheatableCachedState" = cast(
-            "CheatableCachedState", state.async_state
-        )
+    block_info_controller_state: Optional["BlockInfoControllerState"] = None
+    contracts_controller_state: Optional["ContractsControllerState"] = None
```
Arcticae, line 45: This is not a dataclass, which makes those class variables

protostar/cairo_testing/cairo_test_execution_state.py
```diff
@@ -10,45 +11,164 @@
...
+        self._emitted_events.extend(emitted_events)
+
+    def set_pranked_address(self, target_address: Address, pranked_address: Address):
+        self._target_address_to_pranked_address[target_address] = pranked_address
+
+    def remove_pranked_address(self, target_address: Address):
+        if target_address in self._target_address_to_pranked_address:
+            del self._target_address_to_pranked_address[target_address]
+
+    def clone(self):
```
Arcticae, line 98: The naming is not consistent, we have `fork`, `copy`, and now `clone`. I don't see a particular difference between those.

protostar/cairo_testing/cairo_test_execution_state.py
```diff
@@ -10,45 +11,164 @@
...
+    def get_class_hash_from_address(self, contract_address: Address) -> ClassHashType:
+        return self._contract_address_to_class_hash[contract_address]
+
+    def get_contract_abi_from_class_hash(self, class_hash: ClassHashType) -> AbiType:
+        return self._class_hash_to_contract_abi[class_hash]
+
+    def get_contract_abi_from_contract_address(
+        self, contract_address: Address
+    ) -> AbiType:
+        class_hash = self._contract_address_to_class_hash[contract_address]
```
Arcticae, line 64: This can use `self.get_class_hash_from_address`

protostar/cairo_testing/execution_environments/cairo_execution_environment.py
```diff
@@ -28,6 +31,12 @@ async def run_cairo_function(
         *args: Any,
         **kwargs: Any,
     ):
+        CheatableSysCallHandler.block_info_controller_state = (
+            self.state.block_info_controller_state
+        )
+        CheatableSysCallHandler.contracts_controller_state = (
+            self.state.contracts_controller_state
+        )
```
Arcticae, line 39: Setting a static class property with a state instance is a big no-no, it was implemented before and was marked for removal in future versions. We'd like to avoid this.

Arcticae (changes requested):
I like the separation of concerns in the state, although `starknet_state` and `state` are not the names i agree with, it might be correct from user standpoint but not from a technical one. 

I also feel like the states & controllers are now tightly coupled and could be merged into one. I don't see a case where we need a state without a controller right now, or vice-versa. Please correct me if i'm wrong though.

protostar/starknet/__init__.py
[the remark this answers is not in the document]
kasperski95: `CachedState` doesn't make much sense from Protostar perspective.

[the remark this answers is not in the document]
MaksymilianDemitraszek: Even though it is the case, it is still hard to follow. It is better to use one name for one thing, even though it is imperfect. It is easier to know what we are talking about. Something like this will cause a lot of clarification questions in tech discussions "Do we talk about renamed cached state?"

protostar/cheatable_starknet/cheatables/cheatable_syscall_handler.py
kasperski95: Resolving in favor of https://github.com/software-mansion/protostar/pull/1451#discussion_r1101478245

kasperski95:
> I also feel like the states & controllers are now tightly coupled and could be merged into one. I don't see a case where we need a state without a controller right now, or vice-versa. Please correct me if i'm wrong though.

`ExpectEventsController` depends on `ContractsControllerState`. ContractsController is already big and messy. It acts as a Controller (Business logic) and an Interfacer (to Starknet). Adding another concern will create more mess. State need to be forkable.

Instead of joining controllers and state, I think I would be better to join `ContractsContractState` and `BlockInfoContractState` into one.

protostar/cairo_testing/execution_environments/cairo_execution_environment.py
MaksymilianDemitraszek: Agreed, there is no obvious reason to do that, though there are some subtle, but important ones. First, it is much easier to follow and explore the codebase without monkey patching, it is always somewhat confisuing to reason about state, you can't never be sure if there is no race condition. Second reason is we are close to rust migration, and Rust does not allow such hacks. I would prefer to not encourage any form of dependence on them, to simplify the integration in the future. Hope it makes sense.

protostar/cairo_testing/cairo_test_execution_state.py
```diff
@@ -10,45 +11,164 @@
...
         return dataclasses.replace(
             self,
             context=deepcopy(self.context),
             config=deepcopy(self.config),
             output_recorder=self.output_recorder.fork(),
             stopwatch=self.stopwatch.fork(),
-            starknet=self.starknet.copy(),
+            # pylint: disable=protected-access
+            starknet_state=self.starknet_state._copy(),
+            contracts_controller_state=self.contracts_controller_state.clone(),
```
MaksymilianDemitraszek, line 170: I believe it does not work properly. In the new architecture single test can execute multiple transactions, but you clone it only per test, not per transaction. We need copy apply mechanism implemented, thats why we used cached state before. Right now if you invoke transaction which reverts, protostar state may contain some lefovers from this transaciton. Does it make sense?

MaksymilianDemitraszek (changes requested):

protostar/cairo_testing/cairo_test_execution_state.py
kasperski95: It would make sense, if Protostar's hint local were available in a contract. However, they are not. 
That means contract execution doesn't mutate Protostar state. Changes to Protostar state are done after contract execution. Therefore, we don't need to implement the copy and apply mechanism.

```py
with self._starknet_state.copy_and_apply() as state_copy:
            call_info = await entry_point.execute_for_testing(
                state=state_copy,
            )
            self._add_emitted_events(call_info.get_sorted_events()) # <= this won't execute in case of revert
```

protostar/starknet/__init__.py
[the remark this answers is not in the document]
Arcticae: this is also the term that will continue to be used in rust implementation, so it will be easier to reason about the migration to rust, and `legacy <-> new codebase` relation when we get to it.

protostar/cairo_testing/cairo_test_execution_state.py
MaksymilianDemitraszek: Yea, but we store there deployed contracts with syscalls as well

Issue #1280, opened 2022-12-15
Implement caching mechanism for StarknetCompiler and CairoCompiler


Arcticae:
This caching mechanism should be capable of:
- Saving the compiled artifacts somewhere in a file, alongside with a hash of the compiled code
- Comparing the stored artifacts' hashes with a hash of the files to be compiled, and determining if the files need re-compilation 
- Supporting storing pre-processed and compiled files in-memory in order to avoid compiling/processing the same file twice in one program run (thus making the code cleaner, so we don't have to pass the compiled/preprocessed programs around)
- Above in-memory compilation data should be accessible through the workers of test runner. This might come with a challenge to make those picklable

A performant hash function has to be chosen, and some decorators to compilation methods, that compare the parameters and check for cached values.

MaksymilianDemitraszek:
It has low priority

kasperski95:
Reference — Caching in Buck: https://youtu.be/uvNI_E0ZgZU?t=976

kasperski95:
Scarb should be responsible for caching Cairo compilation results.

kasperski95:
Contract name in tests should resolve to the build output.
Caching would be a functionality of `protostar build`.

MaksymilianDemitraszek:
Moving it to the next milestone

PR #1452, opened 2023-02-08
Using Compiled Contracts in tests

kasperski95:
Related https://github.com/software-mansion/protostar/issues/1280

### DIFF
--- a/protostar/commands/test/test_command.py
+++ b/protostar/commands/test/test_command.py
@@ -155,6 +155,11 @@ A glob or globs to a directory or a test suite, for example:
                 type="bool",
                 description="Show gas estimation for each test case. Estimations might be inaccurate.",
             ),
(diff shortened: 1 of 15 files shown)


protostar/commands/test/test_command.py
```diff
@@ -155,6 +155,11 @@ def arguments(self):
                 type="bool",
                 description="Show gas estimation for each test case. Estimations might be inaccurate.",
             ),
+            ProtostarArgument(
+                name="compiled-contracts-dir",
```
MaksymilianDemitraszek, line 159: Shouldn't this one be stored in .protostar by default? Do you think that it is necessary to give user option to change the location?

kasperski95:
We need to introduce the concept of "targets". We should introduce targets with or after Cairo 1 integration.

PR #1453, opened 2023-02-08
Remove address parameter from deploy-account command

piotmag769:
Removed: 
- address parameter from deploy-account command
- redundant address calculation in tests/_conftest/devnet/devnet_account_preparator.py


MaksymilianDemitraszek (changes requested):
Congrats on the first PR 🔥 

We also have to remove the argument from CLI, because right now it is required. It is located in 
```
            *SignableCommandUtil.signable_arguments,
```
after that you will need to generate new CLI reference with `poetry run poe update_cli_docs`

protostar/cli/signable_command_util.py
```diff
@@ -1,97 +1,80 @@
...
 
-        if not private_key_str:
-            private_key_str = os.environ.get(PRIVATE_KEY_ENV_VAR_NAME)
+    key_pair = KeyPair.from_private_key(private_key)
 
-        if not private_key_str or not self._args.account_address:
-            raise ProtostarException(
-                "Signing is mandatory, please provide an account and a private key in order to sign transactions."
-            )
+    account_address = (
```
MaksymilianDemitraszek, line 52: I think the logic here becomes a little convoluted, you can provide account address with two arguments. We could extract address from args in commands that have it and make `account_address` a required argument of `get_signer`. It would separate logic a little bit, as the way account-address is provided is command specific. What do you think?

MaksymilianDemitraszek (changes requested):
Great job, logic-wise it is fine. Can we remove address extracting from the `get_signer` method?

protostar/cli/signable_command_util.py
piotmag769: Agree, modified the function

protostar/cli/signable_command_util.py
```diff
@@ -1,97 +1,73 @@
...
-from .protostar_argument import ProtostarArgument
+from protostar.starknet import Address
+from protostar.cli.common_arguments import PRIVATE_KEY_ENV_VAR_NAME
 
-PRIVATE_KEY_ENV_VAR_NAME = "PROTOSTAR_ACCOUNT_PRIVATE_KEY"
 
+def get_signer(
+    args: Any,
+    network_config: NetworkConfig,
+    account_address: Optional[Union[Address, int, str]],
```
MaksymilianDemitraszek, line 17: We can assume it is there if it is required by command, we also assume it for `signer_class` as well.
I suggest the type
```
account_address: Address
```
and using `from_user_input` constructor. It will be more consistent and `Address` type supports conversion to int.

piotmag769: Changed to Optional[Address] to split account_address validation (in Address class) from get_signer logic.
Optional is required as account_address may not be necessary if args.signer_class is provided.

protostar/cli/signable_command_util_test.py
```diff
@@ -56,7 +54,7 @@ def test_custom_signer_class(mocker: MockerFixture):
     args.private_key_path = None
     args.account_address = None
```
MaksymilianDemitraszek, line 55: We could just pass None to function

protostar/cli/signable_command_util_test.py
```diff
@@ -72,7 +70,7 @@ def test_default_signer_class(
     args.private_key_path = str(pkey_file_factory("0x123"))
     args.account_address = Address.from_user_input("0x123")
 
-    signer = SignableCommandUtil(args).get_signer(network_config)
+    signer = get_signer(args, network_config, args.account_address)
```
MaksymilianDemitraszek, line 73: We could just pass Address, there is no point in adding it to args

protostar/commands/multicall_command.py
```diff
@@ -113,7 +120,9 @@ async def run(self, args: Namespace):
         network_util = NetworkCommandUtil(args)
         network_config = network_util.get_network_config()
         gateway_client = network_util.get_gateway_client()
-        signer = SignableCommandUtil(args).get_signer(network_config=network_config)
+        signer = get_signer(
+            args, network_config=network_config, account_address=args.account_address
```
MaksymilianDemitraszek, line 124: We probably should build Address in those commands

protostar/cli/signable_command_util_test.py
```diff
@@ -86,13 +84,13 @@ def test_wrong_format_of_private_key_env(
 
     args.signer_class = None
     args.private_key_path = None
-    args.account_address = "0x123"
+    args.account_address = Address.from_user_input("0x123")
```
MaksymilianDemitraszek, line 87: We probably don't need to add it to args, we can just pass it to function

MaksymilianDemitraszek (changes requested):

MaksymilianDemitraszek (approved):
Good job, feel free to merge!

Issue #1290, opened 2022-12-16
Display more information when git command fails


kasperski95:
"Error while executing command: ..." doesn't provide enough information to debug issues.

PR #1455, opened 2023-02-09
Remove duplications in Git module

kasperski95:
Reduce duplications. Prepare code for https://github.com/software-mansion/protostar/issues/1290.


protostar/git/git_repository.py
```diff
@@ -120,122 +80,47 @@ def add_submodule(
...
-            cwd=self.repo_path,
-        )
+        self._git(["fetch", "--tags"])
 
-    def get_submodules(self) -> Dict[str, Submodule]:
-        """
-        Returns a dictionary of form:
-        submodule_name: PackageInfo
-        """
+    def _git(self, args: list[str]) -> str:
```
Arcticae, line 120: [nitpick]: I think it is a good case for starred-expression arguments. Although this one works as well, so i'll leave it up to you
```suggestion
    def _git(self, *args: str) -> str:
```

protostar/git/git_repository.py
```diff
@@ -19,98 +21,56 @@ class Submodule:
...
-    def get_repo_root(repo_path: Path):
-        credentials = [] if has_git_credentials() else DEFAULT_CREDENTIALS
-        process = subprocess.run(
-            ["git", *credentials, "rev-parse", "--show-toplevel"],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.DEVNULL,
-            cwd=repo_path,
-            check=False,
-        )
+    def from_existing(repo_path: Path):
```
Arcticae, line 34: I think this can be a `@classmethod`

protostar/git/git.py
```diff
@@ -1,52 +1,72 @@
 import subprocess
```
Arcticae, line 1: Can we use this opportunity to rename this file? Maybe to `ops`/`global_ops`/`commands` ?

protostar/git/git_repository.py
```diff
@@ -19,98 +21,56 @@ class Submodule:
...
+        ensure_user_has_git()
+        path = GitRepository.get_repo_root(repo_path)
+        return GitRepository(path)
 
-        if process.returncode:
+    @staticmethod
+    def get_repo_root(repo_path: Path) -> Path:
```
Arcticae, line 40: This could be an instance method - it'd work like this:

```
GitRepository(repo_path).get_repo_root()
```

But it's not a big deal if we leave it like this

protostar/git/git_repository.py
Arcticae: Also we could allow a broader spectrum of arguments like: `Union[str, int, Path]` and do the conversion inside the method

protostar/git/git_repository.py
```diff
@@ -120,122 +80,47 @@ def add_submodule(
...
 
-    def get_submodules(self) -> Dict[str, Submodule]:
-        """
-        Returns a dictionary of form:
-        submodule_name: PackageInfo
-        """
+    def _git(self, args: list[str]) -> str:
+        return run_git(args=args, cwd=self.repo_path)
 
+    def get_submodule_name_to_submodule(self) -> dict[str, Submodule]:
```
Arcticae, line 123: I actually liked the previous name better, or we could do `get_submodules_dict`? Submodule appears 2 time in the name and it bugs me 😬

Arcticae (approved):
A few nitpicks, but looks good overall 👍

protostar/git/git.py
kasperski95: `ops` doesn't say much. `Commands` can be confused with Protostar commands. I can change it to `git_utils`.

Arcticae: i think we don't like `utils` that much as well

protostar/git/git_repository.py
kasperski95: The previous name was better from the perspective of this file, but it was obscure for callers of that method because there was no info about what is the key without reading the comment.

`get_name_to_submodule`?

protostar/git/git.py
kasperski95: `GitService`?

protostar/git/git_repository.py
Arcticae: `get_name_to_submodule` is fine with me, although usually i don't need that much verbosity in code, it actually obscures what is important in the function logic. this is my personal opinion though. Do as you please - just bringing this up for the sake of discussion and future consideration

protostar/git/git.py
kasperski95: > They typically sit dormant waiting for other objects to call them, but because they’re often relied upon by so many objects, they may contain responsibilities that relatively stable, not changing often.

> For example, utility classes like TextUtil, DateUtil, NumberUtil provide respective text, date, and number utilities and probably don’t change dramatically (or at all for that matter). For this reason, we’d say they’re stable. Just think about what would happen if we had to change the JSON.stringify method in JavaScript.

Arcticae: > Commands can be confused with Protostar commands

this part i don't get, every function here has `git` in its' name, so how can we confuse it?

MaksymilianDemitraszek: I don't get why we don't like utils that much 🤔

protostar/git/git_repository.py
MaksymilianDemitraszek: I agree that type provides enough verbosity

protostar/git/git.py
Arcticae: @MaksymilianDemitraszek those above are git commands wrappers.
There is always a better name than `utils`, `utils` is just a lazy way out for a module name that you don't have an immediate idea for :P Moreover, it does not provide any information about the functionality of the module

kasperski95: This file contains small, git related functions, and often such files are named utils or helpers.

`git_helpers`?

> Commands can be confused with Protostar commands

this part i don't get, every function here has git in its' name, so how can we confuse it?
In this codebase, commands usually refer to Protostar CLI commands. Also, using git CLI through a subprocess is an implementation detail which should be hidden from the outside world. I think we are trying too hard to avoid the u-word. IMO, utils is fine as long as it's prefixed and it's not a directory.

protostar/git/git_repository.py
kasperski95: 1) The type doesn't say what is a string. You have to either guess it, have an explanatory comment, or read the code. 2) The name (`get_submodules`) suggests a list, which is misleading.
3) The bigger the scope, the more explicit the name should be.

kasperski95: And this comment is an example why they should be avoided. The comment says the dictionary contains `PackageInfo` which is true.

Arcticae: You can alias `SubmoduleName` as a `str` and the problem is solved.
When reading the code i did not understand what was happening in a function when given the new verbose name. Describing dictionaries with this verbosity in python is not a commonly seen pattern. 

Why have the type system if we don't use all of it, and keep relying on names to provide full information? 
Isn't the above information duplication as well? 

My mental process when wanting to use a method:
1. Check module/class name
2. Skim through available method names, to check how can i get what i want (submodules)
3. Check what it returns
4. Use it

Where do you usually go differently?

protostar/git/git.py
MaksymilianDemitraszek: Why not git_wrappers?

kasperski95: > A descriptive name for this file could be git_interface.py. This name indicates that the file contains a high-level interface for interacting with Git and encapsulates the underlying subprocess calls to Git. The name also conveys the purpose of the file and what it is used for, which can be useful for someone unfamiliar with the code.
~ChadGPT

protostar/git/git_repository.py
kasperski95: [make git fn variadic](https://github.com/software-mansion/protostar/pull/1455/commits/f03b7783169afd37fe8840c81a22368fb7c3faf9)

protostar/git/git.py
Arcticae: Interesting, but not relevant since it's text-based AI, has no knowledge about our project and our needs/preferences. Thus i would kindly ask to refrain from posting output from ChatGPT as an argument for the discussion.

Arcticae: `git_globals_facade`/`git_facade`? i think this one fits nicely

protostar/git/git_repository.py
kasperski95: Moved this function to `git_helpers.py`.

protostar/git/git.py
kasperski95: > git_globals_facade/git_facade? i think this one fits nicely

I don't know how to interpret globals. I am not sure about the facade.


> Interesting, but not relevant since it's text-based AI, has no knowledge about our project and our needs/preferences. Thus i would kindly ask to refrain from posting output from ChatGPT as an argument for the discussion.

ChatGPT has many flaws, but it's quite good with summarization. It had a knowledge about the file contents and that what matters. It suggested names that we both came up with.

Screenshot 2023-02-10 at 17 13 01

protostar/git/git_repository.py
kasperski95: [refine](https://github.com/software-mansion/protostar/pull/1455/commits/05285d1116239bb1c07a98c4c2550632f5a6093e)

protostar/git/git_repository.py
kasperski95: [use implicit name](https://github.com/software-mansion/protostar/pull/1455/commits/5804133c45c09f8177acbb405b7e7db6bd950e64)

PR #1473, opened 2023-02-13
Make git error message more descriptive

kasperski95:
Make git errors more descriptive. Don't crash when committing unstaged files.
Closes: https://github.com/software-mansion/protostar/issues/1290

Before:
Screenshot 2023-02-13 at 16 52 08

After (but before fixing this concrete problem):
Screenshot 2023-02-13 at 16 51 34

### DIFF
--- a/protostar/git/git_repository.py
+++ b/protostar/git/git_repository.py
@@ -17,6 +17,11 @@ class Submodule:
     path: Optional[Path] = None
 
 
+@dataclass
+class GitStatusResult:
+    files_to_be_committed: list[Path]
+
+
 class GitRepository:
     @classmethod
     def create(cls, repo_path: Path):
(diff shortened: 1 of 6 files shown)


protostar/git/git_repository.py
```diff
@@ -17,6 +17,11 @@ class Submodule:
     path: Optional[Path] = None
 
 
+@dataclass
+class GitStatusResult:
+    staged_file_paths: list[Path]
```
Arcticae, line 22: This is not used anywhere but in tests, we don't need that value

Arcticae (approved):

protostar/git/git_repository.py
kasperski95: It's used in `install_package_from_repo.py`.

I could've removed this dataclass in favor of a more specific function, however I chose to keep the interface of `GitRepository` similar to the `git cli`.

PR #1480, opened 2023-02-13
Store contract's ABI in deploy syscalls

### LINKED ISSUE #1456: Take care of TODO from `CheatableSysCallHandler`
`protostar/cheatable_starknet/cheatables/cheatable_syscall_handler.py` has some leftover todos, we need to take care of them, I believe at least the one in the `_deploy` method is a bug.

Also please include a test which fail this scenario.

Arcticae:
This concerns the new architecture
closes #1456


tests/integration/pure_cairo_vm/test_runner_test/cheatable_syscall_handler_test.cairo
```diff
@@ -0,0 +1,19 @@
+func test_syscall_deployment_abi_storing() {
```
kasperski95, line 1: The test's name doesn't tell why we need to store abi. I think this should be a test of `store` or/and `load` hint local.

Suggestion: `test_loading_from_contract_deployed_by_syscall`

kasperski95 (approved):

MaksymilianDemitraszek (approved):

PR #1481, opened 2023-02-14
Add expect call hint local

karol-bisztyga:

PR #1501, opened 2023-02-16
Add assert expect call hint local

karol-bisztyga:
Depends on #1481

### DIFF
--- a/protostar/cheatable_starknet/controllers/expect_call_controller.py
+++ b/protostar/cheatable_starknet/controllers/expect_call_controller.py
@@ -49,7 +49,7 @@ class ExpectCallController:
                 expected_call
             ]
 
-    def stop_expecting_call(self, expected_call: ExpectedCall):
+    def assert_expect_call(self, expected_call: ExpectedCall):
         contract_address = expected_call.address
         expected_calls = self._cheatable_state.expected_contract_calls.get(
             contract_address
--- a/protostar/cheatable_starknet/controllers/expect_call_controller.py
+++ b/protostar/cheatable_starknet/controllers/expect_call_controller.py
@@ -49,7 +49,7 @@ class ExpectCallController:
                 expected_call
             ]
 
-    def stop_expecing_call(self, expected_call: ExpectedCall):
+    def stop_expecting_call(self, expected_call: ExpectedCall):
         contract_address = expected_call.address
         expected_calls = self._cheatable_state.expected_contract_calls.get(
             contract_address
--- a/protostar/cheatable_starknet/controllers/expect_call_controller.py
+++ b/protostar/cheatable_starknet/controllers/expect_call_controller.py
@@ -51,22 +51,19 @@ class ExpectCallController:
 
     def stop_expecing_call(self, expected_call: ExpectedCall):
         contract_address = expected_call.address
-        data_for_address = self._cheatable_state.expected_contract_calls.get(
+        expected_calls = self._cheatable_state.expected_contract_calls.get(
             contract_address
         )
-        if not data_for_address:
+        if not expected_calls:
             return
-        for selector, calldata in data_for_address:
-            if (
-                int(expected_call.fn_selector) == selector
-                and expected_call.calldata == calldata
-            ):
+        for expected_call_item in expected_calls:
+            if expected_call == expected_call_item:
                 raise ExpectedCallException(
                     contract_address=contract_address,
                     fn_name=str(expected_call.fn_selector),
                     calldata=expected_call.calldata,
                 )
-            
+
     @staticmethod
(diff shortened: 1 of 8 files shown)


Arcticae (commented):
I think this works, and implementation seems fine but i have some doubts about ergonomics of usage. This is not something that i'm familiar with from other frameworks, but there is a similar thing in `jest` i think, which are spies:

1. User constructs a spy to spy on a contract
2. User makes some operations
3. User makes some assertions about calls, or retrieves them from a spy for their use.

@MaksymilianDemitraszek WDYT?

karol-bisztyga:
Interesting. Is there anything else that this approach would give us besides `was_called` (or `get_number_of_calls`) and `get_calls_results`?
I'm just concerned that the benefit of such a redesign would be insufficient compared to the amount of work we'd have to put into it to implement it. That's only my intuition about this for now.

MaksymilianDemitraszek:
@Arcticae yep, this is a nice idea, we could actually add an issue about it and explore it after initial release. I don't want to go to far right now as I am not even sure if we will expose all of the implemented cheatcodes in the initial release. 

@karol-bisztyga wdyt about renaming it to `assert_expect_call`?

karol-bisztyga:
@MaksymilianDemitraszek @Arcticae can we merge this?

karol-bisztyga:
Can we merge this @MaksymilianDemitraszek @Arcticae ?

MaksymilianDemitraszek:
For me it's fine, @Arcticae 's call

protostar/cheatable_starknet/controllers/expect_call_controller.py
```diff
@@ -49,6 +49,21 @@ def add_expected_call(self, expected_call: ExpectedCall):
                 expected_call
             ]
 
+    def assert_expect_call(self, expected_call: ExpectedCall):
+        contract_address = expected_call.address
+        expected_calls = self._cheatable_state.expected_contract_calls.get(
+            contract_address
+        )
+        if not expected_calls:
+            return
```
Arcticae, line 58: This should raise the exception as well

tests/integration/pure_cairo_vm/cheatcodes/expect_call/expect_call_test.cairo
```diff
@@ -10,6 +10,17 @@ func test_expect_call_success() {
   return ();
```
Arcticae, line 10: we should add a test case for failing test when nothing was expected and an assertion was made

Arcticae (changes requested):
I think this cheatcode has a flaw right now:

Suppose a situation where we did not make an expectation before using this cheatcode, it's going to go through even though the call could've not been made.

protostar/cheatable_starknet/controllers/expect_call_controller.py
karol-bisztyga: I don't think so, because it also means that there might be calls that were fulfilled - in this case `expected_calls` will also be empty because all the expected calls were removed. Also, if there are other expected calls that were not called, would we also have to raise an error? Example:
```
expect_call(a)
expect_call(b)
call(a)
assert_expect_call(a) // this is ok, a was called
assert_expect_call(c) // this should raise an error then
```
I don't think we should rise an error in such cases.

What `assert_expect_call` does is it checks that there's no hanging expected call that matches the specified criteria.

What you want `assert_expect_call` to do is: check if there were any expected calls with specified params that were not called.

To achieve that, I think we'd have to have two separate collections: `expected_calls` and `performed_calls`. This is because we'd have to check whether an expectation was made. I think this is a strong edge case and it brings some complexity.

Another thing is that it used to be implemented like that in the old cheat code.

I wouldn't go for it, what do you think?

Arcticae (approved):

protostar/cheatable_starknet/controllers/expect_call_controller.py
karol-bisztyga: Ok let's discuss there: https://github.com/software-mansion/protostar/issues/1509#issuecomment-1441461319

karol-bisztyga:
> I think this works, and implementation seems fine but i have some doubts about ergonomics of usage. This is not something that i'm familiar with from other frameworks, but there is a similar thing in `jest` i think, which are spies:
> 
> 1. User constructs a spy to spy on a contract
> 2. User makes some operations
> 3. User makes some assertions about calls, or retrieves them from a spy for their use.
> 
> @MaksymilianDemitraszek WDYT?

We can continue this discussion at https://github.com/software-mansion/protostar/issues/1525 - I think we forgot about this but if someone actually created an issue for that, you can close mine, but please paste a link here.

PR #1503, opened 2023-02-16
Update bindings

MaksymilianDemitraszek:
Added new function for test compilation
Removed unused functions as we do not need non-test, pure cairo, compilation


Arcticae (approved):
I like it, it removes code. Always nice.

karol-bisztyga (approved):
image

Issue #1436, opened 2023-02-03
Improve installation process of cairo bindings


karol-bisztyga:
This is related to https://github.com/software-mansion/protostar/pull/1316

For now, we use `maturin develop` in a custom bash script. That isn't perfect and should be improved in the future. `maturin build` produces a wheel but using `maturin` is still a custom step that needs to be scripted.

PR #1504, opened 2023-02-16
Use `protostar_cairo_py`

kasperski95:
Use Python package from https://github.com/software-mansion-labs/protostar-cairo-py.

Closes https://github.com/software-mansion/protostar/issues/1436

---

## TODO:
- [ ] fix Rust error `Failed to find development corelib`
- [ ] fix Rust error `Failed to build database` (probably caused by corelib issue)

## Problem
[`corelib`](https://github.com/software-mansion-labs/cairo/tree/main/corelib) is a dynamically linked library and is not shipped with `protostar_cairo_py`.

`corelib` is detected in the following way:
  1. `Cargo.toml` directory (development only `CARGO_MANIFEST_DIR` must be present)
  1. hardcoded relative path to the executable directory (`../../corelib`)

https://github.com/software-mansion-labs/cairo/blob/main/crates/cairo-lang-filesystem/src/detect.rs

## Solution
- ship `corelib` with Protostar
- modify Protostar Cairo Python bindings and the Cairo code so that Protostar can pass a path to `corelib`


MaksymilianDemitraszek:
Can we don't remove `scripts/install_cairo_bindings.sh`? It will be very useful for development

PR #1506, opened 2023-02-16
Simplify cairo1 compiler bindings tests

MaksymilianDemitraszek:
- Move all relevant tests to one directory
- Removed project-based testing flow
- Used datadirs for data


karol-bisztyga (approved):
This looks ok. What do you think about adding the suffix `_data` to the folders in the `cairo1_compiler_bindings`? So for example it would be `parser_test` => `parser_test_data`.

MaksymilianDemitraszek:
I think it would be better but it is the convention of `pytest-datadir` i believe

karol-bisztyga:
Right

Arcticae (approved):

PR #1507, opened 2023-02-17
cairo1 parser test for start_prank and declare

karol-bisztyga:


karol-bisztyga:
this requires https://github.com/software-mansion-labs/cairo/pull/17 to be merged

tests/integration/cairo1/parser_test.py
```diff
@@ -16,23 +16,38 @@ def test_suite_json_fixture(datadir: Path) -> str:
...
+                    "__find_element_max_size": 2**20,
+                    "__squash_dict_max_size": 2**20,
+                    "__keccak_max_size": 2**20,
+                    "__usort_max_size": 2**20,
+                    "__chained_ec_op_max_len": 1000,
+                },
+                run_resources=RunResources(n_steps=100000000000000000),
+                verify_secure=False,
+            )
+            assert mocked_error_code == runner.get_return_values(3)[0]
```
karol-bisztyga, line 53: I couldn't find a less hacky way of checking the result for now :/

tests/integration/cairo1/parser_test.py
```diff
@@ -10,29 +10,44 @@
...
-                "__squash_dict_max_size": 2**20,
-                "__keccak_max_size": 2**20,
-                "__usort_max_size": 2**20,
-                "__chained_ec_op_max_len": 1000,
-            },
-            run_resources=RunResources(n_steps=100000000000000000),
-            verify_secure=False,
-        )
-    assert cheat_mock.call_count == 6
+    for mocked_error_code in [0, 1]:
```
MaksymilianDemitraszek, line 19: Could we add compilation to this pipeline? json is a bit hard to modify, it was used before we had bindings

tests/integration/cairo1/parser_test.py
MaksymilianDemitraszek: We could probably do it like it is done here:
https://github.com/software-mansion-labs/cairo/blob/aba4b69ef925214c77ed57672313ca25f9de4014/crates/cairo-lang-runner/src/lib.rs#L106

tests/integration/cairo1/parser_test.py
```diff
@@ -10,29 +10,44 @@
...
+            "return_value", (object,), {"err_code": mocked_error_code, "ok": ok}
+        )()
+
+        for case in test_suite.test_cases:
+            runner = CairoFunctionRunner(program=test_suite.program, layout="all")
+            runner.run_from_entrypoint(
+                case.offset,
+                *[],
+                hint_locals={
+                    "roll": cheat_mock,
```
MaksymilianDemitraszek, line 39: WDYT splitting each cheat into separate test? I am not sure about that, would be easier to debug though.

karol-bisztyga: I just pushed changes with this.

tests/integration/cairo1/parser_test.py
karol-bisztyga: Done with the new changes.

tests/integration/cairo1/parser_test.py
```diff
@@ -10,29 +10,44 @@
 
 @pytest.fixture(name="test_suite_json")
 def test_suite_json_fixture(datadir: Path) -> str:
-    # Cairo source code of the tested fixture - ../cairo_compiler/contracts/roll_test.cairo
+    # Cairo source code of the tested fixture - ../cairo_compiler/contracts/hint_locals_test.cairo
     with open(datadir / "compiled_test_suite.json", "r") as file:
         return file.read()
 
 
 def test_parse(mocker: MockerFixture, test_suite_json: str):
```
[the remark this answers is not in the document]
karol-bisztyga: Not relevant anymore, I decided to keep 2 separate files for this as they're different tests and may get different over time, plus it's easier to navigate and it's not that much repetition tbh.

tests/integration/cairo1/parser_test.py
karol-bisztyga: From what I understand this is basically the same as [here](https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/cairo/common/cairo_function_runner.py#L185). I guess the function "signature" (rust) may be equivalent to the "identifiers" (py) which we're missing, so we are unable to automatically retrieve the return data, but it's still there in the memory.

karol-bisztyga: I'm going to try to solve this

karol-bisztyga: I don't think it's quickly doable but might've missed something. I think we should unblock for now https://github.com/software-mansion/protostar/issues/1519

MaksymilianDemitraszek: Sure

MaksymilianDemitraszek (approved):

PR #1512, opened 2023-02-20
Rename bindings

karol-bisztyga:


Arcticae (commented):
I think we should wrap this in a module (or a class) if we're refactoring. My idea:

- `import protostar.cairo.cairo_bindings as cairo1` +  `cairo1.compile_starknet_contract()`
- `Cairo1.compile_starknet_contract()` (class approach)

PR #1516, opened 2023-02-20
Catch wrong value of private key

piotmag769:
Added:
- catching obviously invalid (too large/small) values of private keys in signable_command_util.get_signer
- appropriate tests

Fixed:
- typo in README


protostar/cli/signable_command_util.py
```diff
@@ -47,7 +47,12 @@ def get_signer(
             f"Invalid private key format ({private_key_str}). Please provide hex-encoded number."
         ) from v_err
 
-    key_pair = KeyPair.from_private_key(private_key)
+    try:
+        key_pair = KeyPair.from_private_key(private_key)
+    except AssertionError as a_err:
+        raise ProtostarException(
+            "Invalid private key value. Please provide valid private key."
```
MaksymilianDemitraszek, line 54:
```suggestion
            "Unable to parse provided private key. Please provide a valid private key."
```
maybe a bit more descriptive

MaksymilianDemitraszek (approved):

protostar/cli/signable_command_util.py
piotmag769: The issue here is not the fact that we are not able to parse it, but the fact that the key doesn't satisfy the constraint in starkware package (so its value is either too large r too small)

`assert 0 < priv_key < EC_ORDER`

MaksymilianDemitraszek: your call

Arcticae (approved):

PR #1532, opened 2023-02-24
Added depth 2 tests for warp and roll

piotmag769:
Added:
- depth 2 tests for warp and roll

Updated:
- refactored remaining tests

Fixed:
- roll_warp_tester contract - marked setters @ external and made them return no value


tests/integration/pure_cairo_vm/contracts/roll_warp_tester.cairo
```diff
@@ -3,49 +3,71 @@
...
-@view
-func block_number_setter{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() -> (res: felt) {
+@external
+func block_number_setter{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}() {
     let (block_number) = get_block_number();
     stored_block_number.write(block_number);
-    return (res=block_number);
+
+    return ();
 }
```
karol-bisztyga, line 52: I think we should change the naming of these functions.

The naming is really confusing to me and the names don't really describe what they do exactly. I think we should change them similarly to the bindings - to indicate actions, not the roles (and do it in a slightly better/more-readable manner).

| current name | what it does | proposed new name | 
| ------------- | ------------- | ------------- | 
| stored_block_timestamp_getter | returns stored timestamp | get_stored_block_timestamp |
| stored_block_number_getter | returns stored block number | get_stored_block_number |
| timestamp_getter | returns a timestamp using a syscall  | get_block_timestamp_wrapper |
| block_number_getter | returns block number using a syscall | get_block_number_wrapper |
| block_timestamp_setter | sets the stored timestamp to the one from the syscall | set_stored_timestamp_to_syscall_value |
| block_number_setter | sets the stored block number to the one from the syscall | set_block_number_to_syscall_value |

If you don't like any of the proposed names, that's ok, but please consider making these functions' names more descriptive.

tests/integration/pure_cairo_vm/contracts/roll_warp_tester.cairo
```diff
@@ -3,49 +3,71 @@
...
+func call_block_timestamp_setter{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(contract_address: felt) {
+    TimestampAndBlockNumberTesterContract.block_timestamp_setter(contract_address=contract_address);
+    return ();
+}
+
+@external
+func call_block_number_setter{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(contract_address: felt) {
+    TimestampAndBlockNumberTesterContract.block_number_setter(contract_address=contract_address);
+    return ();
+}
```
karol-bisztyga: Maybe that should be in a separate file?

tests/integration/pure_cairo_vm/contracts/roll_warp_tester.cairo
```diff
@@ -3,49 +3,71 @@
 from starkware.cairo.common.cairo_builtins import HashBuiltin
```
karol-bisztyga, line 3: Don't you think we should split this into 2 files, one for `roll` and one for `warp`?

tests/integration/pure_cairo_vm/contracts/roll_warp_tester.cairo
piotmag769: Agree, will do.

tests/integration/pure_cairo_vm/contracts/roll_warp_tester.cairo
piotmag769: Do you mean the functions calling external contract? I don't see a reason to do that honestly. Especially since we already want to split the contract into 2.

tests/integration/pure_cairo_vm/contracts/l1_handler_contract.cairo
```diff
@@ -36,15 +36,19 @@ func on_l1_msg_set_block_timestamp{syscall_ptr: felt*, pedersen_ptr: HashBuiltin
...
+    func set_stored_block_timestamp_to_syscall_value() {
+    }
+
+    func get_stored_block_timestamp() -> (res:felt) {
     }
 }
 
 @l1_handler
-func call_set_block_timestamp{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
+func call_set_set_stored_block_timestamp_to_syscall_value_and_get_its_value{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
```
Arcticae, line 47: this name is too long IMHO and becomes unreadable, can you shorten it?

tests/integration/pure_cairo_vm/contracts/warp_contract.cairo
```diff
@@ -0,0 +1,40 @@
...
+}
+
+@contract_interface
+namespace TimestampTesterContract {
+    func set_stored_block_timestamp_to_syscall_value() {
+    }
+}
+
+@external
+func call_set_stored_block_timestamp_to_syscall_value{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(contract_address: felt) {
```
Arcticae, line 37: this sets the peer block timestamp right? `call` indicates no state changes as well

Arcticae (approved):
LGTM, some nitpicks below :)

tests/integration/pure_cairo_vm/contracts/warp_contract.cairo
piotmag769: Will change to `invoke`

PR #1538, opened 2023-02-27
Display more informative error message when account is not deployed (declare command)

piotmag769:
Screenshot 2023-02-27 at 12 58 23

### DIFF
--- a/protostar/starknet_gateway/gateway_facade.py
+++ b/protostar/starknet_gateway/gateway_facade.py
@@ -226,9 +226,10 @@ class GatewayFacade(MulticallClientProtocol):
                 auto_estimate=max_fee == "auto",
             )
         except ClientError as ex:
+            account_address_found_in_message = hex(int(account_address)) in ex.message
             message = (
                 "No account associated with provided account address found. Contact your wallet provider."
-                if "StarknetErrorCode.UNINITIALIZED_CONTRACT" in ex.message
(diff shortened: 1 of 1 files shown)


karol-bisztyga:
Hmm, you could add information that it depends on https://github.com/software-mansion/protostar/pull/1532 or you could set the target branch to `add-depth-2-tests-for-warp-and-roll` so the changes related to this PR would be isolated and could be easily seen.

I'm going to wait with the review until #1532 is merged.

protostar/starknet_gateway/gateway_facade.py
```diff
@@ -218,11 +218,21 @@ async def declare(
...
+        try:
+            declare_tx = await account_client.sign_declare_transaction(
+                compiled_contract=compiled_contract,
+                max_fee=max_fee if isinstance(max_fee, int) else None,
+                auto_estimate=max_fee == "auto",
+            )
+        except ClientError as ex:
+            message = (
+                "No account associated with provided account address found. Contact your wallet provider."
+                if "StarknetErrorCode.UNINITIALIZED_CONTRACT" in ex.message
```
Arcticae, line 231: I think we want to also check for address of the account, matching the one given in the error.

piotmag769: Fine, but it can only be done by searching through the ex.message string :/

Arcticae: I think this is good enough

PR #1546, opened 2023-02-28
Flatten code in compiler

karol-bisztyga:
I don't see a gain in excluding one line of code into a separate function if it's going to be called just once.


MaksymilianDemitraszek (approved):
I am ok with merging this, but I am not sure why do we need to change this, is it useful in any way for build command?

Radinyn (approved):

karol-bisztyga:
It is useful because it makes the code less complicated and easier to read.

PR #1547, opened 2023-02-28
Set fixed poetry version to 1.3.2 in all Actions

Arcticae:
_"Whatever method you use, it is highly recommended to explicitly control the version of Poetry used, so that you are able to upgrade after performing your own validation."_

Source: https://python-poetry.org/docs/master/#ci-recommendations

Related issue on CI:
https://github.com/software-mansion/protostar/actions/runs/4292622238/jobs/7479310114

### DIFF
--- a/.github/workflows/build_binaries.yml
+++ b/.github/workflows/build_binaries.yml
@@ -16,12 +16,12 @@ jobs:
       - name: Set up Python 3.9.14
         uses: actions/setup-python@v2
         with:
-          python-version: 3.9.14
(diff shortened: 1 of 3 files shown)


karol-bisztyga (approved):
It's ok, just wondering if there's a way maybe of defining it only once and re-using it across these scripts? Maybe we could use env vars?

Arcticae:
> It's ok, just wondering if there's a way maybe of defining it only once and re-using it across these scripts? Maybe we could use env vars?

Can do

Issue #1544, opened 2023-02-28
`test-cairo1` extensive testing


MaksymilianDemitraszek:
We need to test
- [x] glob querying
- [x] ignore command
- [x] `::test_case_name` syntax
- [ ] `no-progress-bar` (probably impossible to test, but leaving it here)
- [x] `exit-first`
- [ ] `report-slowest-tests` (probably impossible to test, but leaving it here)
- [x] `last-failed`

MaksymilianDemitraszek:
Some of the tests can be e2e as well

Arcticae:
Partially blocked by #1550

PR #1548, opened 2023-02-28
Cairo1 tests e2e

MaksymilianDemitraszek:


karol-bisztyga:
Can you make the CI right? Maybe you need to update the base branch and do a rebase?

protostar/protostar_cli.py
```diff
@@ -68,6 +68,11 @@ def __init__(
                     type="bool",
                     description="Disable colors.",
                 ),
+                ProtostarArgument(
+                    name="cairo1",
+                    type="bool",
+                    description="Use cairo1 variant of commands",
+                ),
```
Arcticae, line 75: I think this one is unused

protostar/protostar_cli.py
```diff
@@ -114,6 +119,12 @@ async def _run_command_from_args(self, args: Any) -> None:
             self._extend_pythonpath_with_cairo_path(cairo_path_arg)
             _apply_pythonpath()
 
+        # This is only temporary until all commands will support cairo1 variants
+        if args.cairo1 and args.command not in ["test"]:
+            raise ProtostarException(
+                f"Command {args.command} does not have cairo1 variant for now"
+            )
```
Arcticae, line 126: That seems unused as well

protostar/commands/test_cairo1/test_cairo1_command.py
```diff
@@ -0,0 +1,139 @@
...
+            no_progress_bar=args.no_progress_bar,
+            safe_collecting=False,
+            exit_first=args.exit_first,
+            seed=None,
+            max_steps=None,
+            slowest_tests_to_report_count=args.report_slowest_tests,
+            gas_estimation_enabled=False,
+            messenger=messenger,
+            use_cairo1_test_runner=True,
+        )
```
Arcticae, line 135: Shouldn't we use this opportunity to prune some unused and irrelevant code from this command? Some redundancy would be introduced but i just wonder if you considered it. 

It's not a deal breaker for this PR though.

protostar/commands/test_cairo1/test_cairo1_command.py
```diff
@@ -0,0 +1,139 @@
...
+                type="int",
+                description="Print slowest tests at the end.",
+                default=0,
+            ),
+            ProtostarArgument(
+                name="last-failed",
+                short_name="lf",
+                type="bool",
+                description="Only re-run failed and broken test cases.",
+            ),
```
Arcticae, line 103: Do all of these work? We should add some tests for checking that

protostar/compiler/project_cairo_path_builder.py
```diff
@@ -1,7 +1,14 @@
 from pathlib import Path
 
 
-class ProjectCairoPathBuilder:
+class LinkedLibrariesBuilder:
```
Arcticae, line 4: [nitpick]: This is kind of confusing since linked libraries is a term for cairo1 exclusively, and this class is a required interface in CLI class. If we extracted the interface, named it like `CairoPathGetter` or `CompilationPathBuilder` and inherited in both classes, we could require the interface in CLI class. As per the name of the method (`build_project_cairo_path_list`): this is a method which is implemented in the `LikedLibrariesBuilder` and is kind of mixed into the cairo1 context where the cairo path doesn't really exist anymore. 

Now that i'm thinking after writing this, perhaps we should divide the 2 concepts from each other, since we changed argument name anyway? Even if we have to add an extra argument to the legacy command.

I understand that it's a workaround, but it's not ideal nonetheless. I'm marking it as nitpick, to be addressed now or later if we don't have time for it rn.

Arcticae (approved):
This seems to work, but i am missing some tests for this i.e. Broken Suites test.
Also some arguments we are accepting require testing. This can be done in a separate PR or this one (please link an issue if delegating to another PR).

protostar/commands/test_cairo1/test_cairo1_command.py
MaksymilianDemitraszek: Yup I know, there is a task for that 😉 
[#tests exten](https://github.com/software-mansion/protostar/issues/1544)

MaksymilianDemitraszek:
We need to add e2e test for failing as well but it is blocked for now

protostar/compiler/project_cairo_path_builder.py
MaksymilianDemitraszek: #1559

MaksymilianDemitraszek: I will address this but I want first to have cairo0 runner removed

protostar/commands/test_cairo1/test_cairo1_command.py
```diff
@@ -0,0 +1,139 @@
+from argparse import Namespace
```
karol-bisztyga, line 1: as discussed offline, I think we should move this to a separate folder like `protostar/commands/cairo1_commands` in the same manner as in  https://github.com/software-mansion/protostar/pull/1551. I'm open with using a different name, but I think it should be in one, separate directory

protostar/commands/test_cairo1/test_cairo1_command.py
```diff
@@ -0,0 +1,139 @@
...
+    def name(self) -> str:
+        return "test-cairo1"
+
+    @property
+    def description(self) -> str:
+        return "Executes cairo1 tests."
+
+    @property
+    def example(self) -> Optional[str]:
+        return "$ protostar test"
```
karol-bisztyga, line 48:
```suggestion
        return "$ protostar test-cairo1"
```
?

tests/e2e/fixtures/cairo1_project/tests/test_a.cairo
```diff
@@ -0,0 +1,12 @@
+#[test]
+fn test_assert_trueA() {
+    let x = 1;
```
karol-bisztyga, line 3: This doesn't check anything, shouldn't we `assert`, especially since the method name itself says we do so? Or if `assert` is not an option, I'd rather rename the function and leave it empty 🤷 Better do less than do confusing stuff if you ask me.

tests/e2e/fixtures/cairo1_project/libraries/external_lib_bar/bar.cairo
```diff
@@ -0,0 +1,3 @@
+fn bar() {
+    external_lib_foo::foo::foo();
+}
```
karol-bisztyga, line 3: All these files for external libs are just copied from `tests/integration/cairo1_compiler_bindings/data`, isn't there a way we could reuse already-existent files?

protostar/compiler/project_cairo_path_builder.py
```diff
@@ -1,7 +1,14 @@
...
+    def build_project_cairo_path_list(
+        self, relative_cairo_path_list: list[Path]
+    ) -> list[Path]:
+        return relative_cairo_path_list
+
+
+class ProjectCairoPathBuilder(LinkedLibrariesBuilder):
```
karol-bisztyga, line 11: Do we need this inheritance here? What does it give us?

tests/e2e/conftest.py
```diff
@@ -50,7 +50,13 @@ def cairo_fixtures_dir(protostar_repo_root: Path) -> Path:
 def copy_fixture(
     cairo_fixtures_dir: Path,
 ) -> CopyFixture:
-    return lambda file, dst: shutil.copy(cairo_fixtures_dir / file, dst)
+    def copy_content(file: Union[Path, str], dst: Union[Path, str]):
+        if (cairo_fixtures_dir / file).is_dir():
+            shutil.copytree(cairo_fixtures_dir / file, dst)
+        else:
+            shutil.copy(cairo_fixtures_dir / file, dst)
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: good idea

tests/e2e/fixtures/cairo1_project/tests/test_a.cairo
MaksymilianDemitraszek: Yup thats an important detail

tests/e2e/fixtures/cairo1_project/libraries/external_lib_bar/bar.cairo
MaksymilianDemitraszek: I am not really a fan of over anti-duplicating. IMO we should keep stuff from e2e tests separated. This code is tightly coupled to the tests and abstracting it make it much harder to modify and adjust.

protostar/compiler/project_cairo_path_builder.py
MaksymilianDemitraszek: We do not need it here, but it is hard to remove for now and makes cope of the task much bigger
First i want to remove cairo0 runner
Then we can work on https://github.com/software-mansion/protostar/issues/1559 which includes the change you suggested

tests/e2e/fixtures/cairo1_project/libraries/external_lib_bar/bar.cairo
karol-bisztyga: Sure, it was just a thought, fair enough.

PR #1551, opened 2023-03-01
Cairo1 build command

karol-bisztyga:
TODO:
- [x]  tests - I think we should first do `protostar init-cairo1` because e2e build tests depend on `protostar init` and we need to have a properly set-up cairo1 project to be tested.


tests/e2e/conftest.py
```diff
@@ -78,14 +82,29 @@ def init_project(
...
+            fixture_path = (
+                Path(__file__).parent.parent.parent
+                / "tests"
+                / "e2e"
+                / "fixtures"
+                / "basic_cairo1.cairo"
+            )
+            assert main_path.exists() and fixture_path.exists()
+            main_path.write_text(fixture_path.read_text())
+            Path(project_dir / "build").mkdir()
```
karol-bisztyga, line 107: Once `init-cairo1` is done in https://github.com/software-mansion/protostar/issues/1496, we can just probably do
```
child = pexpect.spawn(f"{protostar_bin} init-cairo1 {real_project_name}")
```

protostar/cairo/cairo_bindings.py
```diff
@@ -12,11 +12,13 @@ class TestCollectorOutput:
...
 
 
-def compile_starknet_contract(
+def compile_starknet_contract_from_path(
     input_path: Path,
     output_path: Optional[Path] = None,
     cairo_path: Optional[list[Path]] = None,
 ) -> Optional[str]:
+    if output_path:
+        Path(output_path).parent.mkdir(parents=True, exist_ok=True)
```
Arcticae, line 21: [nitpick]: can this be extracted to a function? `ensure_output_path` would be an okay name

tests/integration/_conftest/protostar_fixture.py
```diff
@@ -223,6 +226,14 @@ def build_sync(self):
         args = self._prepare_build_args()
         return asyncio.run(self._build_command.run(args))
 
+    async def cairo1build(self):
```
Arcticae, line 229:
```suggestion
    async def cairo1_build(self):
```

tests/integration/_conftest/protostar_fixture.py
```diff
@@ -223,6 +226,14 @@ def build_sync(self):
...
+        args = self._prepare_build_args()
+        return await self._cairo1_build_command.run(args)
+
+    def cairo1build_sync(self):
```
Arcticae, line 233:
```suggestion
    def cairo1_build_sync(self):
```

protostar/commands/cairo1_commands/cairo1_build_command.py
```diff
@@ -0,0 +1,79 @@
...
+from protostar.compiler import (
+    ProjectCompiler,
+)
+import protostar.cairo.cairo_bindings as cairo1
+
+
+class Cairo1BuildCommand(ProtostarCommand):
+    def __init__(
+        self,
+        project_compiler: ProjectCompiler,
```
Arcticae, line 16: It seems weird that we use a compiler from previous versions here. Can't we make it separate, or pass higher level objects?

protostar/commands/cairo1_commands/cairo1_build_command.py
```diff
@@ -0,0 +1,79 @@
...
+    ):
+        super().__init__()
+        self._project_compiler = project_compiler
```
Arcticae, line 19: what we really need here is only `configuration_file`, not a legacy project compiler

protostar/commands/cairo1_commands/cairo1_build_command.py
```diff
@@ -0,0 +1,79 @@
...
+        logging.info("Contracts built successfully")
+
+    async def build(
+        self,
+        output_dir: Path,
+        relative_cairo_path: Optional[List[Path]] = None,
+    ) -> None:
+        configuration_file = self._project_compiler.configuration_file
+        cairo_path = relative_cairo_path or []
+        for contract_name in configuration_file.get_contract_names():
```
Arcticae, line 65: This will compile all contracts as cairo1 contracts, which will be wrong for most of the projects

tests/e2e/cairo1/test_building.py
```diff
@@ -0,0 +1,12 @@
...
+import pytest
+
+from tests.e2e.conftest import ProtostarFixture
+
+
+@pytest.mark.usefixtures("init_cairo1")
+def test_default_build(protostar: ProtostarFixture):
+    protostar(["build-cairo1"])
+    dirs = listdir()
+    assert "build" in dirs
```
Arcticae, line 12: we should assert something more about the result, other than the directory exists

Arcticae (changes requested):

protostar/commands/cairo1_commands/cairo1_build_command.py
karol-bisztyga: My understanding was that if someone calls `build-cairo1`, they want to build cairo1 contracts. So I assume that now projects may contain mixed contracts - some of them will be written in cairo 0 and some of them in cairo 1, right?

Could we do this as a separate task, especially since I think some discussions regarding this are in progress?

Arcticae: taking into account everything we discussed, i think we should leave behavior as-is, if users want to build cairo0 contacts they will use the legacy command

protostar/commands/cairo1_commands/cairo1_build_command.py
Arcticae: i think this point still stands

protostar/commands/cairo1_commands/cairo1_build_command.py
Arcticae: i think this point still stands (after discussion about interop)

protostar/cairo/cairo_bindings.py
karol-bisztyga: [add ensure_output_path](https://github.com/software-mansion/protostar/pull/1551/commits/f9d1dcc794a4ffe3136b8e16da1dc39ef03611d5)

tests/integration/_conftest/protostar_fixture.py
karol-bisztyga: [fix naming](https://github.com/software-mansion/protostar/pull/1551/commits/cee9313463e0d794f240111aecb98ae3a4eafe30)

protostar/commands/cairo1_commands/cairo1_build_command.py
karol-bisztyga: [avoid passing proj compiler to the build command](https://github.com/software-mansion/protostar/pull/1551/commits/2ae4aa91169bacb80b98de40fee8fe1799529653) please let me know if that's ok

protostar/commands/cairo1_commands/cairo1_build_command.py
karol-bisztyga: I think there are two comments about the same thing, let's follow up in https://github.com/software-mansion/protostar/pull/1551#discussion_r1127040435

Arcticae (approved):

PR #1555, opened 2023-03-02
Separate updating bindings and installation

Arcticae:
Closes https://github.com/software-mansion/protostar/issues/1556

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,24 @@ curl -L https://raw.githubusercontent.com/software-mansion/protostar/master/inst
 1. Build Protostar: `poe build`
     - You can find the newly created binary at `dist/protostar/protostar`
 
+#### Bumping cairo bindings version
+To bump cairo bindings version to the latest commit on fork (master branch), run: 
+
+```shell
+poetry run poe bump_cairo_bindings
+```
+
+After this, you can pull & rebuild your local version of bindings by running:
+```shell
+poetry run poe install_cairo_bindings
+```
+
+##### Caveats:
+Remember to have your working tree clean, since it creates a commit on the current branch.
+
+The command will check if the tracking branch is master, so you don't commit/push to master by accident (would be rejected by branch protection).
+
+
 #### Submodules development 
 You can use submodules from a different branch than main. Run commands
 ```shell
--- a/scripts/bump_cairo_bindings.sh
+++ b/scripts/bump_cairo_bindings.sh
@@ -16,7 +16,7 @@ fi
 
 git submodule update --recursive --remote --init cairo
 if [ -z "$(git status --porcelain)" ]; then
-  >&2 echo "Bindings are up to date!"
+  >&2 echo "Bindings are already up to date!"
   exit 0
 fi
 
--- a/scripts/bump_cairo_bindings.sh
+++ b/scripts/bump_cairo_bindings.sh
@@ -15,5 +15,10 @@ fi
 
 
 git submodule update --recursive --remote --init cairo
+if [ -z "$(git status --porcelain)" ]; then
+  >&2 echo "Bindings are up to date!"
+  exit 0
+fi
+
 git add ./cairo
 git commit -m "Bump cairo bindings" --no-verify
--- a/scripts/bump_cairo_bindings.sh
+++ b/scripts/bump_cairo_bindings.sh
@@ -1,6 +1,19 @@
 #!/usr/bin/env bash
 set -e
 
+if [ -n "$(git status --porcelain)" ]; then
+  >&2 echo "Working tree has uncommited changes, please stash/commit them before bumping submodule version"
+  exit 1
+fi
+
+tracking_branch = "$(git for-each-ref --format='%(upstream:short)' "$(git symbolic-ref -q HEAD)")"
+
+if [ tracking_branch == "origin/master" ] ; then
+  >&2 echo "You are currently tracking master branch. Checkout to another branch before bumping submodule version"
+  exit 1
+fi
+
+
 git submodule update --recursive --remote --init cairo
 git add ./cairo
 git commit -m "Bump cairo bindings" --no-verify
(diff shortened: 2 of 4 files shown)


MaksymilianDemitraszek (approved):

scripts/bump_cairo_bindings.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+
+git submodule update --recursive --remote --init cairo
+git add ./cairo
+git commit -m "Bump cairo bindings" --no-verify
```
MaksymilianDemitraszek, line 6: Can we include update info in README

MaksymilianDemitraszek (approved):

scripts/bump_cairo_bindings.sh
Arcticae: Yes i'm doing that, and posting on a channel. Script is not finished yet though, i will add a check if the working tree is clean (so we don't commit any more changes that we don't want)

scripts/bump_cairo_bindings.sh
```diff
@@ -0,0 +1,24 @@
...
+set -e
+
+if [ -n "$(git status --porcelain)" ]; then
+  >&2 echo "Working tree has uncommited changes, please stash/commit them before bumping submodule version"
+  exit 1
+fi
+
+tracking_branch="$(git for-each-ref --format='%(upstream:short)' "$(git symbolic-ref -q HEAD)")"
+
+if [ $tracking_branch == "origin/master" ] ; then
```
karol-bisztyga, line 11:
```suggestion
if [ "$tracking_branch" == "origin/master" ] ; then
```
shellcheck

scripts/bump_cairo_bindings.sh
```diff
@@ -0,0 +1,24 @@
...
+  >&2 echo "You are currently tracking master branch. Checkout to another branch before bumping submodule version"
```
karol-bisztyga, line 12: shouldn't we do something like `git fetch --all` first? Or did I miss something?

scripts/bump_cairo_bindings.sh
```diff
@@ -0,0 +1,24 @@
...
+
+
+git submodule update --recursive --remote --init cairo
+if [ -z "$(git status --porcelain)" ]; then
+  >&2 echo "Bindings are already up to date!"
+  exit 0
+fi
+
+git add ./cairo
+git commit -m "Bump cairo bindings" --no-verify
```
karol-bisztyga, line 24: This doesn't seem to be safe, we could have some staged changes already, I think we shouldn't automate this at all, I'd remove it.

karol-bisztyga: NIT

Another thing, I'd add a log like this:
```
echo "Bindings have been updated/bumped ({old_commit_hash}=>{new_commit_hash})"
```
the `*_comit_hash` thing is not necessary (command: `git rev-parse --short HEAD`)
So I think just "Bindings have been updated\bumped" is ok, the rest is just an idea.

README.md
```diff
@@ -78,6 +78,24 @@ curl -L https://raw.githubusercontent.com/software-mansion/protostar/master/inst
...
+poetry run poe bump_cairo_bindings
+```
+
+After this, you can pull & rebuild your local version of bindings by running:
+```shell
+poetry run poe install_cairo_bindings
+```
+
+##### Caveats:
+Remember to have your working tree clean, since it creates a commit on the current branch.
```
MaksymilianDemitraszek, line 94: Maybe it shouldn't commit on the current branch then 🤔 Not sure, your call

MaksymilianDemitraszek (approved):

scripts/bump_cairo_bindings.sh
Arcticae: > This doesn't seem to be safe, we could have some staged changes already, I think we shouldn't automate this at all, I'd remove it.

i check whether working tree is clean.

README.md
Arcticae: We should be ok. this is not a big deal since it's a development script.

scripts/bump_cairo_bindings.sh
Arcticae: For what reason?

scripts/bump_cairo_bindings.sh
Arcticae: Btw not automating this will make this whole thing confusing, and we'll have problems (just like i did today).
With this script we're abstracting away how we're doing things, we just want to do a thing.

Arcticae: > the *_comit_hash thing is not necessary (command: git rev-parse --short HEAD)

i'm afraid i don't understand what you mean by that

scripts/bump_cairo_bindings.sh
karol-bisztyga: Let's say you install the bindings, then the master of the bindings is updated and then you run this script. This check is going to pass, right? Maybe I missed the concept, so in such a case, we're supposed to run install script again?

Arcticae: > in such a case, we're supposed to run the install script again?

yes. it's mentioned in readme

scripts/bump_cairo_bindings.sh
karol-bisztyga: > > This doesn't seem to be safe, we could have some staged changes already, I think we shouldn't automate this at all, I'd remove it.
> 
> i check whether working tree is clean.

Ah, right, sorry. Ok, we're good.

scripts/bump_cairo_bindings.sh
karol-bisztyga: Ok, thx

karol-bisztyga (approved):

Issue #1550, opened 2023-03-01
Support failing tests in cairo1 runner


Arcticae:
We need to identify panics and fail the test when it panics

PR #1557, opened 2023-03-02
Added panic support PoC

MaksymilianDemitraszek:
Related issue #1550

### DIFF
--- a/protostar/cairo/cairo_function_runner_facade.py
+++ b/protostar/cairo/cairo_function_runner_facade.py
@@ -59,3 +59,21 @@ class CairoRunnerFacade:
             self._previous_runner
         ), "No runs were performed, so no return values are available!"
         return self._previous_runner.get_return_values(n_ret)
+
+    def is_panic(self) -> bool:
+        result = self.get_return_values(3)
+        return result[0] != 0
+
+    def get_panic_data(self) -> list[int]:
+        assert self.is_panic()
+        assert self._previous_runner
+        result = self.get_return_values(3)
+
+        panic_data_start = result[1]
+        panic_data_end = result[2]
+        iterator = panic_data_start
+        panic_data = []
+        while iterator != panic_data_end:
+            panic_data.append(self._previous_runner.vm_memory.data[iterator])
+            iterator = iterator + 1
+        return panic_data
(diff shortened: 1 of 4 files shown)


protostar/cairo/cairo_function_runner_facade.py
```diff
@@ -59,3 +59,21 @@ def get_return_values(self, n_ret: int):
             self._previous_runner
         ), "No runs were performed, so no return values are available!"
         return self._previous_runner.get_return_values(n_ret)
+
+    def is_panic(self) -> bool:
+        result = self.get_return_values(3)
+        return result[0] != 0
+
+    def get_panic_data(self) -> list[int]:
+        assert self.is_panic()
```
karol-bisztyga, line 68: This way we're forced to use this pattern:
```
if is_panic():
  get_panic_data()
```

Can't we just do `get_panic_data` and if there was no panic, it will return an empty array? And then, `is_panic` could be just:
```
def is_panic(self) -> bool:
  return not self.get_panic_data()
```

protostar/cairo/cairo_function_runner_facade.py
```diff
@@ -59,3 +59,21 @@ def get_return_values(self, n_ret: int):
...
+        result = self.get_return_values(3)
+
+        panic_data_start = result[1]
+        panic_data_end = result[2]
+        iterator = panic_data_start
+        panic_data = []
+        while iterator != panic_data_end:
+            panic_data.append(self._previous_runner.vm_memory.data[iterator])
+            iterator = iterator + 1
+        return panic_data
```
karol-bisztyga, line 79:
```suggestion
    def is_panic(self) -> bool:
        return not self.get_panic_data()

    def get_panic_data(self) -> list[int]:
        assert self._previous_runner
        result = self.get_return_values(3)
        if result[0] == 0:
            return []

        panic_data_start = result[1]
        panic_data_end = result[2]
        iterator = panic_data_start
        panic_data = []
        while iterator != panic_data_end:
            panic_data.append(self._previous_runner.vm_memory.data[iterator])
            iterator = iterator + 1
        return panic_data
```

MaksymilianDemitraszek:
@karol-bisztyga This is only a PoC how to extract panic data from vm, I do not plan to work on the top of this PR

PR #1564, opened 2023-03-04
Add tests for `init-cairo1`

piotmag769:


templates/cairo0/protostar.toml
```diff
@@ -1,3 +1,7 @@
-["protostar.project"]
+[project]
+protostar-version = "0.0.0"
+lib-path = "lib"
+
+[contracts]
+main = ["src/main.cairo"]
 
-["protostar.contracts"]
```
piotmag769, line 3: protostar.toml in templates has no impact on how it is created during `init` command call, but it has to be this way to ensure it is the same as the one we create [here](https://github.com/software-mansion/protostar/blob/816cba84923c2d002e677ebc14289b82777682af/protostar/commands/init/project_creator/_project_creator.py#L28) for tests

tests/integration/commands/init_command_test.py
```diff
@@ -0,0 +1,43 @@
...
+    "create_protostar_project, cairo_version",
+    [
+        (CairoVersion.cairo0, CairoVersion.cairo0),
+        (CairoVersion.cairo1, CairoVersion.cairo1),
+    ],
+    indirect=["create_protostar_project"],
+)
+def test_init(
+    create_protostar_project: CreateProtostarProjectFixture, cairo_version: CairoVersion
+):
```
piotmag769, line 21: Read [this comment](https://github.com/software-mansion/protostar/pull/1564/files#r1126353789) first.

There we pass "create_protostar_project" as [indirect](https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments) argument to let pytest know it has pass the first argument of the tuple (the second one cairo_version) to the request object that is passed to the fixture [there](https://github.com/software-mansion/protostar/pull/1564/files#diff-c87c5173e88bb0f60b00107f4586304f4869a976a4f998425efe3481c0db9f56R142-R147)

tests/integration/conftest.py
```diff
@@ -139,9 +141,15 @@ def __call__(self) -> ContextManager[ProtostarFixture]:
...
 @pytest.fixture(name="create_protostar_project", scope="module")
 def create_protostar_project_fixture(
+    request: FixtureRequest,
     session_mocker: MockerFixture,
     tmp_path_factory: TempPathFactory,
 ) -> CreateProtostarProjectFixture:
+    cairo_version = (
+        CairoVersion.cairo0 if not hasattr(request, "param") else request.param
+    )
+    assert isinstance(cairo_version, CairoVersion)
```
piotmag769, line 151: If request doesn't have an attribute "param" we default to CairoVersion.cairo0 - this way we don't break any existing tests.

tests/integration/conftest.py
```diff
@@ -139,9 +141,15 @@ def __call__(self) -> ContextManager[ProtostarFixture]:
 
 @pytest.fixture(name="create_protostar_project", scope="module")
 def create_protostar_project_fixture(
+    request: FixtureRequest,
     session_mocker: MockerFixture,
     tmp_path_factory: TempPathFactory,
 ) -> CreateProtostarProjectFixture:
+    cairo_version = (
+        CairoVersion.cairo0 if not hasattr(request, "param") else request.param
```
karol-bisztyga, line 149: Can you check if there's a way to do this without using `hasattr`? I think we want to avoid it if possible.

tests/integration/commands/init_command_test.py
```diff
@@ -0,0 +1,43 @@
...
+
+        _, diff, errors = cmpfiles(
+            dcmp.left, dcmp.right, dcmp.same_files, shallow=False
+        )
+        if diff or errors:
+            return False
+
+        for sub_dcmp in dcmp.subdirs.values():
+            if not compare_directories(sub_dcmp):
+                return False
```
karol-bisztyga, line 34: So this function goes only 1 level deep, right? Wouldn't it be better to go as deep as necessary? Maybe we should do a recursion here? Or maybe `subdirs` contains **all** the subdirectories no matter how deep they are? Quick look at https://docs.python.org/3/library/filecmp.html#filecmp.dircmp.subdirs didn't help :/

tests/integration/conftest.py
```diff
@@ -139,9 +141,15 @@ def __call__(self) -> ContextManager[ProtostarFixture]:
 
 @pytest.fixture(name="create_protostar_project", scope="module")
 def create_protostar_project_fixture(
+    request: FixtureRequest,
     session_mocker: MockerFixture,
     tmp_path_factory: TempPathFactory,
 ) -> CreateProtostarProjectFixture:
```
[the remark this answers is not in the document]
piotmag769: After creating every cairo1 command we can probably just parametriaze the fixture there instead of doing it on the test level -> pytest.fixture(params=...).

templates/cairo1/protostar.toml
```diff
@@ -1,3 +1,7 @@
-["protostar.project"]
+[project]
+protostar-version = "0.0.0"
+lib-path = "lib"
```
karol-bisztyga, line 3: Is this necessary?

karol-bisztyga (commented):
Would be cool to get a review from someone who has more pytest expertise.

MaksymilianDemitraszek (changes requested):
Please make `main` a cairo functions for now, not a contract as we do not support contracts  for now. Please write simple test for this function in the test file.

templates/cairo1/protostar.toml
piotmag769: Yes, you are correct

tests/integration/commands/init_command_test.py
piotmag769: But there is a recursive call inside the `if` statement... Or am I missing sth?

tests/integration/commands/init_command_test.py
```diff
@@ -0,0 +1,43 @@
...
+
+        return True
```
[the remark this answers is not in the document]
piotmag769: I can refactor it to be more aligned with project's code style then.

tests/integration/conftest.py
piotmag769: I could check the class instance but I think that would make it worse. From source code of pytest:
```python
# Notes on the type of `param`:
# -`request.param` is only defined in parametrized fixtures, and will raise
#   AttributeError otherwise. Python typing has no notion of "undefined", so
#   this cannot be reflected in the type.
# - Technically `param` is only (possibly) defined on SubRequest, not
#   FixtureRequest, but the typing of that is still in flux so this cheats.
# - In the future we might consider using a generic for the param type, but
#   for now just using Any.
self.param: Any
```

tests/integration/commands/init_command_test.py
```diff
@@ -0,0 +1,43 @@
...
+from filecmp import dircmp, cmpfiles
+
+import pytest
+
+from protostar.cairo import CairoVersion
+from tests.integration.conftest import CreateProtostarProjectFixture
+from tests.conftest import PROJECT_ROOT
+
+
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
piotmag769: Will do + make it an optional argument.

tests/integration/commands/init_command_test.py
karol-bisztyga: ok, right `compare_directories` is recursive.

tests/integration/commands/init_command_test.py
```diff
@@ -0,0 +1,35 @@
...
+
+    _, diff, errors = cmpfiles(dcmp.left, dcmp.right, dcmp.same_files, shallow=False)
+    if diff or errors:
+        return False
+
+    for sub_dcmp in dcmp.subdirs.values():
+        if not compare_directories(sub_dcmp):
+            return False
+
+    return True
```
Arcticae, line 35: [nitpick]: this could take a form of series of assertions, instead of returning boolean values. there will be no feedback what to fix when the test fails

Arcticae (approved):

PR #1575, opened 2023-03-06
`test-cairo1` comprehensive e2e tests

### LINKED ISSUE #1544: `test-cairo1` extensive testing

Arcticae:
Closes #1544


Arcticae:
@MaksymilianDemitraszek i wouldn't sacrifise brevity for perf in tests anyway. This is not much overhead over using the test collector directly anyway, right?

tests/e2e/cairo1/test_cairo1_testing.py
```diff
@@ -43,3 +42,98 @@ def test_failing_tests(protostar: ProtostarFixture, copy_fixture: CopyFixture):
...
+
+    result = protostar(
+        ["--no-color", "test-cairo1", "--last-failed"], ignore_exit_code=True
+    )
+
+    # Only ran 2 failed ones
+    assert "Running previously failed tests" in result
+    assert "Collected 1 suite, and 2 test cases" in result
+    assert "test_panic_multiple_values" in result
+    assert "test_panic_single_value" in result
```
karol-bisztyga, line 124:
```suggestion
    assert "test_panic_multiple_values" in result
    assert "test_panic_single_value" in result
    assert "test_ok" not in result
```

tests/e2e/cairo1/test_cairo1_testing.py
```diff
@@ -22,10 +22,9 @@ def test_no_tests_found(protostar: ProtostarFixture):
 
 def test_failing_tests(protostar: ProtostarFixture, copy_fixture: CopyFixture):
     copy_fixture("cairo1_project", "./cairo1_project")
-    template_test_files = Path("cairo1_project/tests").glob("**/*")
-    for template_test_file in template_test_files:
-        template_test_file.unlink()
-    copy_fixture("cairo1/failing_test.cairo", "./cairo1_project/tests")
+    copy_fixture(
+        "cairo1/failing_test.cairo", "./cairo1_project/tests/failing_test.cairo"
+    )
```
karol-bisztyga, line 27: How does this work without removing existing test files? Below, we explicitly check that this line is in the output: `"Collected 1 suite, and 3 test cases"`

Arcticae: I removed the default test from the template in this PR, so adding tests is now explicit. It's much better to add what you want IMHO than remove all tests (and also it caused some errors for me in case of nested directories god knows why)

karol-bisztyga: I see now, right, that totally makes sense! Nice.

karol-bisztyga (approved):

PR #1583, opened 2023-03-08
Fix the example cheatcode

piotmag769:
We had not been waiting for async test function to execute.

### DIFF
--- a/protostar/testing/environments/fuzz_test_execution_environment.py
+++ b/protostar/testing/environments/fuzz_test_execution_environment.py
@@ -192,7 +192,9 @@ class FuzzTestExecutionEnvironment(ContractBasedTestExecutionEnvironment):
                                 inputs=inputs,
                             ) from reported_ex
 
-            if hasattr(test, "hypothesis"):  # this checks only if "given" cheatcode is used
+            if hasattr(
+                test, "hypothesis"
+            ):  # this checks only if "given" cheatcode is used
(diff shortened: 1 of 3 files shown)


tests/integration/fuzzing/parameterized_test.cairo
```diff
@@ -82,3 +82,21 @@ func test_check_exact_example{syscall_ptr: felt*, range_check_ptr}(a: felt, b: f
...
+        example(a=11)
+    %}
+    return ();
+}
+
+@external
+func test_negative_example{syscall_ptr: felt*, range_check_ptr}(a: felt) {
+    %{
+        assert ids.a == 12
+    %}
```
piotmag769, line 98:
```suggestion
    assert a = 12;
```

protostar/testing/environments/fuzz_test_execution_environment.py
```diff
@@ -1,3 +1,4 @@
+import asyncio
 import dataclasses
 from asyncio import to_thread
```
piotmag769, line 3: Maybe we could use asyncio.to_thread instead since we already import asyncio

piotmag769 (commented):
Some nitpicks

tests/integration/fuzzing/parameterized_test.cairo
Arcticae: Actually that should do the job. Good idea.

mkaput (approved):
Yep, this looks to be the smallest fix possible for this

PR #1586, opened 2023-03-08
Cairo1 testing docs

### LINKED ISSUE #1535: Cairo 1 tests docs
Document how to use `test-cairo1`

Arcticae:
I will gladly accept any suggestions about the content of this page. 

Wondering if we should include a verbose example, (step-by-step) besides the test cases that are already presented.

Closes #1535


website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,76 @@
...
+
+`__setup__` and `<test_name>_setup` from previous version are not supported for now.
+
+### 3. Test function type
+
+A test function must not return any values, and be panickable for correct test result assessment
+
+If your last statement returns a value, you must follow it with a line with blank statement (`;`).
+
+If you fail to comply to this, the test function will not pass the type check, and test collecting will fail.
```
karol-bisztyga, line 78:
```suggestion
---
sidebar_label: Testing
---

# Running tests with Cairo 1

:::info
This functionality is in the alpha stage, expect rapid iteration
:::


## Writing your first test

To make a test function, you need to mark the function with a decorator `#[test]`

```
#[test]
fn test_bool_operators() {
    assert(1 == 1);
}
```

To write a test that fails, you will need to use `panic`, here's how you do it:

```
use array::ArrayTrait;

// Single value in the panic payload
#[test]
fn test_panic_single_value() {
    let mut data = array_new::<felt>();
    array_append::<felt>(ref data, 21);
    panic(data)
}

// Multiple values in the panic payload
#[test]
fn test_panic_multiple_values() {
    let mut data = array_new::<felt>();
    array_append::<felt>(ref data, 101);
    array_append::<felt>(ref data, 102);
    array_append::<felt>(ref data, 103);
    panic(data)
}
```

## Running the tests

To run cairo 1 tests, there is a special command called `test-cairo1`.
It is a sister command to the `test` command, it will collect all the tests in the given directory/module, run them, and print out a summary.

Tests are run on Cairo VM, so no Starknet syscalls are available from the test code.

:::info
There is no support currently for starknet contracts at the moment, it's a work in progress.
:::

## Caveats
### 1. Test collecting
`test-cairo1` will collect all tests ending with `.cairo` since there's no distinction between cairo 0 and cairo 1 files in terms of extension right now.

That means that you will either have to specify a regex to match your test names (see [command reference](../cli-reference.md#test-cairo1)), or keep them in a separate directory to avoid syntax errors.

A `test_` file prefix or `_test` postfix is required as well, to mark the files as test suites.

### 2. Test state

`__setup__` and `<test_name>_setup` from the previous version are not supported for now.

### 3. Test function type

A test function must not return any values and be panickable for correct test result assessment

If your last statement returns a value, you must follow it with a line with a blank statement (`;`).

If you fail to comply with this, the test function will not pass the type check, and test collecting will fail.
```

grammarly

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,76 @@
...
+    array_append::<felt>(ref data, 102);
+    array_append::<felt>(ref data, 103);
+    panic(data)
+}
+```
+
+## Running the tests
+
+To run cairo 1 tests, there is a special command called `test-cairo1`.
+It is a sister-command to `test` command, it will collect all the tests in the given directory/module, run them, and print out a summary.
```
karol-bisztyga, line 50: I'd use "sibling" instead of "sister" here

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,76 @@
...
+
+A `test_` file prefix or `_test` postfix is required as well, to mark the files as test suites.
+
+### 2. Test state
+
+`__setup__` and `<test_name>_setup` from previous version are not supported for now.
+
+### 3. Test function type
+
+A test function must not return any values, and be panickable for correct test result assessment
```
karol-bisztyga, line 74: I think we should also mention that it mustn't have arguments

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,76 @@
...
+
+If your last statement returns a value, you must follow it with a line with blank statement (`;`).
```
karol-bisztyga, line 76: This is a bit confusing. I assume you it's about this syntax that if you miss the trailing `;` for the last command it is a return value for the function. If so, I'd rearrange this sentence a bit:
```
Please note, that cairo1 uses the rust-like syntax for the return values, so if the last instruction in a function doesn't have a ';' at the end, cairo1 treats this instruction as a returning one. That may lead to errors.
```

I don't know, I guess my suggested line isn't perfect but this is what I have in my mind more or less.

karol-bisztyga (changes requested):

website/docs/tutorials/08-cairo-1-support/01-testing.md
karol-bisztyga: Ok this is broken, I'll do this again

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,78 @@
...
+To make a test function, you need to mark the function with a decorator `#[test]`
+
+```
+#[test]
+fn test_bool_operators() {
+    assert(1 == 1);
+}
+```
+
+To write a test which fails, you will need to use `panic`, here's how you do it:
```
karol-bisztyga, line 23:
```suggestion
To write a test that fails, you will need to use `panic`, here's how you do it:
```

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,78 @@
...
+    array_append::<felt>(ref data, 102);
+    array_append::<felt>(ref data, 103);
+    panic(data)
+}
+```
+
+## Running the tests
+
+To run cairo 1 tests, there is a special command called `test-cairo1`.
+It is a sister-command to `test` command, it will collect all the tests in the given directory/module, run them, and print out a summary.
```
karol-bisztyga, line 50:
```suggestion
To run cairo 1 tests, there is a special command called `test-cairo1`.
It is a sister command to the `test` command, it will collect all the tests in the given directory/module, run them, and print out a summary.
```

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,78 @@
...
+
+Tests are ran on Cairo VM, so no Starknet syscalls are available from the test code.
```
karol-bisztyga, line 52:
```suggestion
Tests are run on Cairo VM, so no Starknet syscalls are available from the test code.
```

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,78 @@
...
+
+See [command reference](../../cli-reference.md#test-cairo1) for more details on usage.
+
+:::info
+There is no support currently for starknet contracts at the moment, it's a work in progress.
+:::
+
+## Caveats
+### 1. Test collecting
+`test-cairo1` will collect all tests ending with `.cairo`, since there's no distinction between cairo 0 and cairo 1 files in terms of extension right now.
```
karol-bisztyga, line 62:
```suggestion
`test-cairo1` will collect all tests ending with `.cairo` since there's no distinction between cairo 0 and cairo 1 files in terms of extension right now.
```

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,78 @@
...
+
+That means that you will either have to specify a regex to match your test names (see [command reference](../../cli-reference.md#test-cairo1)), or keep them in a separate directory to avoid syntax errors.
+
+A `test_` file prefix or `_test` postfix is required as well, to mark the files as test suites.
+
+### 2. Test state
+
+`__setup__` and `<test_name>_setup` from previous version are not supported for now.
```
karol-bisztyga, line 70:
```suggestion
`__setup__` and `<test_name>_setup` from the previous version are not supported for now.
```

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,78 @@
...
+
+### 3. Test function type
+
+A test function must not return any values, and be panickable for correct test result assessment
```
karol-bisztyga, line 74:
```suggestion
A test function must not return any values and be panickable for correct test result assessment
```

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,78 @@
...
+
+If your last statement returns a value, you must follow it with a line with blank statement (`;`).
+
+If you fail to comply to this, the test function will not pass the type check, and test collecting will fail.
```
karol-bisztyga, line 78:
```suggestion
If you fail to comply with this, the test function will not pass the type check, and test collecting will fail.
```

website/docs/tutorials/08-cairo-1-support/01-testing.md
Arcticae: I added an example

website/docs/tutorials/08-cairo-1-support/01-testing.md
```diff
@@ -0,0 +1,90 @@
...
+```
+fn foo() -> felt {
+    1
+}
+
+#[test]
+fn test_foo() {
+    foo();
+    ; // foo returns a value, so this would make the test invalid 
+}
```
karol-bisztyga, line 86: This will not pass the type checker (not panicable), I think we should add code snippets that actually work.

Also, I don't fully understand this example, line 7 already has the `;` in the end, so a new line with `;` is not necessary.

Last but not least, in this comment `// foo returns a value, so this would make the test invalid` it is a bit confusing what `this` refers to - is it the `foo()` line or the `;` line? Actually, it looks like it refers to the `;` line but it should refer to the `foo()` line.

How about this:

```suggestion
fn foo() -> felt {
    assert(1 == 1, 'simple check');
    1
}

#[test]
fn test_foo() {
    foo()
    ; // foo returns a value, so without this line, the test would be invalid
}
```

karol-bisztyga (approved):

PR #1623, opened 2023-03-17
Modular protostar - architecture

MaksymilianDemitraszek:
You can view it formatted here https://github.com/software-mansion/protostar/blob/desing-doc/modular-architecture/design_docs/protostar-architecture.md


design_docs/protostar-architecture.md
```diff
@@ -0,0 +1,63 @@
...
+- Make workflow asynchronous between smaller sub-teams operating on contracts between modules
+- Allow incremental rewrite to Rust
+- Remove compilation responsibility from Protostar codebase
+
+
+
+## General overview
+First decomposition of Protostar will introduce 3 modules
+
+- Protostar Gateway
```
Arcticae, line 14: This name is a bit unfortunate since it coincides with SW's `Gateway`

design_docs/protostar-architecture.md
```diff
@@ -0,0 +1,63 @@
...
+This document proposes the idea how to modularize Protostar. Goals of the modularization are
+- Allow to introduce better ownerships of the codebase parts
+- Make workflow asynchronous between smaller sub-teams operating on contracts between modules
+- Allow incremental rewrite to Rust
+- Remove compilation responsibility from Protostar codebase
+
+
+
+## General overview
+First decomposition of Protostar will introduce 3 modules
```
Arcticae, line 12: What about CLI itself? I guess "gateway" will have this module inside (parsing args, etc) right?

design_docs/protostar-architecture.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Protostar Gateway
+- Protostar Starknet
+- Protostar Testing backend
+
+Each of the modules would be a separate poetry package (open question?), to simplify scenarios of updating different dependencies. Example of such scenario, Protostar-starknet relies on starknet.py and devnet but testing backend does not. As a final goal everything would be under one cargo workspace.
```
Arcticae, line 18: Some of the modules will be crates though (protostar gateway), right? What are the benefits of separating those into separate packages as of now?

design_docs/protostar-architecture.md
```diff
@@ -0,0 +1,63 @@
...
+
+Anything outside of this transaction will be encapsulated in the backend.
+
+
+### Starknet CLI
+This is a facade over starknet.py SDK, we can migrate it to Rust very soon and use Starknet.rs instead if its capabilities are sufficient.
+
+
+## Communication
+Communication layer between modules will be JSON, simillar to the current Protostar `--json` to make it language-independent. We can do communication either through FFI or calling as sub-processes, this is an open question.
```
Arcticae, line 41: I know that FFI is causing a lot of pain right now and might be harder to work with, WDYT?

Arcticae (commented):
The concept is really nice. Execution steps need to be explained well with the team, and i think we should start ASAP with whatever is unblocked.

design_docs/protostar-architecture.md
mkaput: Looks like so

design_docs/protostar-architecture.md
mkaput: Why not `Protostar CLI` ?

design_docs/protostar-architecture.md
MaksymilianDemitraszek: @Arcticae yup, communication contract with Scarb will be the reposnsibility of CLI

design_docs/protostar-architecture.md
MaksymilianDemitraszek: >Example of such scenario, Protostar-starknet relies on starknet.py and devnet but testing backend does not.
This was a scenario, easier version upgrades

design_docs/protostar-architecture.md
mkaput: Do you really need two-way live communication between gateway and deeper modules? I thought it would be more like command execution? Why not simply pass params via arguments/env/argfile, pipe stdin/stdout/stderr and wait for execution?

design_docs/protostar-architecture.md
mkaput: I think what would be beneficial is to split Protostar into two apps: test and starknet

design_docs/img/diagram.png
mkaput, line 1: It's not possible currently to inject a compiler into Scarb. Dojo forked Scarb and produces their own binary.

But I think I have come up with an elegant solution to this problem and we can prioritize implementing it -- worth discussin on call.

design_docs/protostar-architecture.md
Arcticae: that's a good point, we have to probably think about the implications though

MaksymilianDemitraszek: @mkaput I know that was a first idea, on the other hand to way communication removes formatting responsibility form specific teams which makes it much easier to keep consistent. I still consider it as an open question

mkaput: The problem is that human-readable and JSON formats are not in sub/superset relation. You would basically have to invent a third format for transport. I don't think it's worth it over copy-paste of a rather simple mechanism that is present currently in Protostar.

piotmag769: I wouldn't overcomplicate it - I think the solution proposed by @mkaput is just fine. End users doesn't really care how it is done and for us (developers) the cons outweigh the pros imo, at least for now.

design_docs/protostar-architecture.md
piotmag769: And if so - is it only for us (developers) to make the development easier or do we want to communicate that fact to end users?

design_docs/protostar-architecture.md
```diff
@@ -0,0 +1,63 @@
+# Protostar modular architecture
+
+This document proposes the idea how to modularize Protostar. Goals of the modularization are
+- Allow to introduce better ownerships of the codebase parts
+- Make workflow asynchronous between smaller sub-teams operating on contracts between modules
+- Allow incremental rewrite to Rust
+- Remove compilation responsibility from Protostar codebase
```
piotmag769, line 7: Won't we still have the compiler in Protostar CLI?

MaksymilianDemitraszek: Short-term yes, long term everything will be handled by scarb

design_docs/protostar-architecture.md
MaksymilianDemitraszek: @piotmag769 Undelrying architecture is for us, but we are going to communicate with users that protostar will be accesible through scarb

design_docs/protostar-architecture.md
```diff
@@ -0,0 +1,63 @@
...
+## Threats
+- Mixing technologies and using bindings will introduce additional overhead, hopefully we will be able to remove it completely at some point.
+
+## Testing
+Each module will be tested separately based on the cross-module contracts.
+Additionally we will introduce e2e tests set involving scarb layer testing some basing flows.  
+
+## Open questions
+- Communication flow between modules - subprocesses or FFI
+- Should we have a separate poetry package for each module?
```
karol-bisztyga, line 56: Did you do research about this or not yet? I wonder if poetry supports something like workspaces in node.js.

MaksymilianDemitraszek: I did, poetry itsef doesn't https://github.com/python-poetry/poetry/issues/936

There is some stuff like https://pypi.org/project/poetry-workspaces/ but it doesn't seem very mature

Arcticae (approved):

PR #1625, opened 2023-03-17
Declare cheatcode

### LINKED ISSUE #1454: Adapt `declare` and `deploy_contract` to work with cairo1 felt argument.

cptartur:
Closes #1454


protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -1,54 +1,65 @@
...
-        return DeclaredContract(class_hash)
+        return DeclaredContract(class_hash=declared_class.class_hash)
+
+
+def make_contract_class(sierra_contract: str) -> ContractClass:
+    return ContractClass.loads(sierra_contract)
+
+
+def compute_class_hash(casm_contract: str) -> int:
+    return compute_compiled_class_hash(parse_casm(json.loads(casm_contract)))
```
MaksymilianDemitraszek, line 65: Those functions seem a bit too anemic, can we avoid it? It is completely understandable what they do from their body. Introducing additional functions brings unnecessary mental overhead when jumping around code

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -119,10 +133,48 @@ async def _transform_calldata_to_cairo_data(
                 ) from dt_exc
         return calldata or []
 
+    async def declare_sierra_contract(
```
MaksymilianDemitraszek, line 136: Why do we need this one in `PreparedContract` 🤔

pyproject.toml
```diff
@@ -27,7 +27,7 @@ tqdm = "^4.65.0"
 typing-extensions = "^4.3.0"
 tomlkit = "^0.11.6"
 maturin = "^0.14.15"
-web3 = "5.31.4"
+web3 = "6.0.0b4"
```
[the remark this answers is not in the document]
cptartur: No. This was a temporary fix. Should be partially resolved on starknet-0.11 branch now.

protostar/cheatable_starknet/controllers/contracts.py
cptartur: It's ContractsController. Diff is just weird.

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -103,6 +131,12 @@ async def _transform_calldata_to_cairo_data(
     ) -> CairoData:
         if isinstance(calldata, collections.Mapping):
             contract_class = await self.cheatable_state.get_contract_class(class_hash)
+
+            if isinstance(contract_class, CompiledClass):
+                # TODO add handling
+                raise ValueError("Cair1 contracts do not support data transformation.")
```
cptartur, line 137: Do we need better handling here?

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -117,10 +151,50 @@ async def _transform_calldata_to_cairo_data(
...
+        )
+
+        with self.cheatable_state.copy_and_apply() as state_copy:
+            await tx.apply_state_updates(
+                state=state_copy, general_config=starknet_config
+            )
+
+        abi = contract_class.abi
+        # TODO figure out how it works
+        # self._add_event_abi_to_state(abi)
```
cptartur, line 187: Abis of cairo1 contract's are plain strings which is incompatible with abis we use here. Should the abi related logic be updated (and should it be done in this PR) so it's possible to to `_add_event_abi_to_state`?

tests/integration/cairo1_cheatcodes/declare/declare_test.cairo
```diff
@@ -0,0 +1,27 @@
...
+            array_append::<felt>(ref data, x);
+            panic(data)
+        },
+    }
+}
+
+#[test]
+fn test_declare_nonexistent() {
+    match declare('abcdef') {
+        Result::Ok(class_hash) => {
```
MaksymilianDemitraszek, line 18: Can we add tests
- declare cairo0 contract - should fail?
- declare broken cairo1 contract

tests/integration/cairo1_cheatcodes/declare/declare_test.cairo
```diff
@@ -0,0 +1,27 @@
...
+            assert(class_hash != 0, 'class_hash != 0');
+        },
+        Result::Err(x) => {
+            let mut data = array_new::<felt>();
```
MaksymilianDemitraszek, line 22: Can we assert error message content here instead?

tests/integration/cairo1_cheatcodes/declare/data/cairo_project.toml
```diff
@@ -0,0 +1,2 @@
+[crate_roots]
```
MaksymilianDemitraszek, line 1: I don't think contracts need cairo_project, am I right?

protostar/compiler/project_compiler.py
```diff
@@ -104,6 +117,128 @@ def compile_contract_from_contract_identifier(
             )
         return self.compile_contract_from_contract_name(contract_identifier, config)
 
+    def compile_contract_to_sierra_from_contract_identifier(
```
MaksymilianDemitraszek, line 120: ProjectCompiler is Cairo0 entity, can we rename it to Cairo0Compiler instead and create separate entity for Cairo1?

MaksymilianDemitraszek: `compile_contract_to_sierra_from_contract_identifier` shouldn't be supported for cairo1, we should only support `ContractName`

protostar/cheatable_starknet/controllers/contracts.py
MaksymilianDemitraszek: It depends it will be raised as Unexpected exception by protostar
If you want exception to be pretty printed you have to raise CheatcodeException

MaksymilianDemitraszek: But it should be catched and re-raised from HintLocal not from Controller for consistency

MaksymilianDemitraszek: @Arcticae

MaksymilianDemitraszek:
Generally LGTM, good job and congrats of first major PR 🔥 left couple comments

protostar/compiler/project_compiler.py
```diff
@@ -37,6 +45,11 @@ class ProjectCompilerConfig:
     hint_validation_disabled: bool = False
 
 
+class OutputType(Enum):
```
Arcticae, line 48: I don't understand why is that enum used in the compiler, do we need it outside the `ProjectCompiler` class?

Arcticae (changes requested):

protostar/cheatable_starknet/controllers/contracts.py
Arcticae: That's right 👍

cptartur: It's not used in Declare cheatcode, so perhaps this should be addressed when updating relevant cheatcodes (deploy_prepared and prepare I believe)?

protostar/cheatable_starknet/controllers/contracts.py
cptartur: Should I leave it as todo? I think it's a okay indicator of this feature missing in case someone touches this code.

MaksymilianDemitraszek: It is ok leaving a TODO but please also create and issue

tests/integration/cairo1_cheatcodes/declare/declare_test.cairo
cptartur: We've decided on raising `CheatcodeException` instead. Invalid contract names/paths are non-recoverable errors in tests.

protostar/cheatable_starknet/controllers/contracts.py
cptartur: #1663

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -138,6 +212,8 @@ async def declare_cairo0_contract(
             )
 
         abi = contract_class.abi
+        assert abi is not None
```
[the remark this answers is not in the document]
cptartur: ABI is now (or was always?) typed as Optional in `DeprecatedCompiledClass`.

protostar/compiler/project_compiler.py
cptartur: It's not needed outside, I just think using enum here provides stronger typing than relying on strings or other solution. Prefixed it with underscore so it's obvious it's internal.

protostar/compiler/project_compiler.py
Arcticae: let's remove dead code please

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -0,0 +1,71 @@
...
+    ):
+        self._contracts_controller = contracts_controller
+        self._project_compiler = project_compiler
+
+    @property
+    def name(self) -> str:
+        return "declare"
+
+    def _build(self) -> Callable:
+        return self.declare
```
Arcticae, line 39: the new style of defining `hint_local`s that was agreed upon is returning a function defined inside of this `_build`

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def declare(self, contract: int) -> DeclaredContract:
```
Arcticae, line 41: There shouldn't be this much code in `HintLocal`, can it be moved to the controller? (refer to `DeployHintLocal` as per how much code should be in the hintlocal itself)

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -77,6 +85,26 @@ class DeployedContract:
...
+        )
+
+    def run_validate_entrypoint(
+        self,
+        remaining_gas: int,
+        state: SyncState,
+        resources_manager: ExecutionResourcesManager,
+        general_config: StarknetGeneralConfig,
+    ) -> Tuple[Optional[CallInfo], int]:
+        # This hack might interfere with other things
```
Arcticae, line 104: If we're doing a hack, can we explain what it hacks? It might help with debugging some day

Arcticae (changes requested):

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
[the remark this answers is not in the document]
MaksymilianDemitraszek: @cptartur similarly to old cheatcodes documentation https://docs.swmansion.com/protostar/docs/tutorials/testing/cheatcodes

We can do it in a separate PR to not block anybody

protostar/compiler/project_compiler.py
cptartur, line 1: This works to the extend that is tested by declare hint local tests, but that's all.

Ideally this module should have it own tests (perhaps we can just copy and adapt Cairo0ProjectCompiler tests?). 
Should this be done in this or separate PR (and if in separate, should we prioritize testing this further for in the current milestone?)

@Arcticae @MaksymilianDemitraszek

Arcticae (approved):

protostar/compiler/project_compiler.py
MaksymilianDemitraszek: I wouldn't do it at all, I would spend the time instead to remove this abstraction, it should be faster than writing tests

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -73,6 +97,31 @@ class DeployedContract:
...
+        #
+        # By default InternalDeclare has a quite complicated run_validate_entrypoint method
+        # that among other things verifies if the sender_address of the Declare transaction is of a deployed contract.
+        # There's no account in protostar internals so this validation will always fail. This hack completely removes
```
MaksymilianDemitraszek, line 120: Please use docstring instead

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -55,11 +68,22 @@ class ConstructorInvocationException(ContractsCheaterException):
     pass
 
 
+class OperationNotSupportedForCairo1Exception(ProtostarException):
+    def __init__(self, message: str = "Operation not supported for Cairo1 contracts."):
+        super().__init__(message=message)
```
MaksymilianDemitraszek, line 73: I would raise CheatcodeException there instead, is there any reason why we need to create a new error type here?

protostar/cheatable_starknet/controllers/contracts.py
MaksymilianDemitraszek: but it is a nice use of comments, very informative ❤️

MaksymilianDemitraszek (approved):

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -113,10 +170,51 @@ async def _transform_calldata_to_cairo_data(
...
+        self,
+        contract_class: ContractClass,
+        compiled_class: CompiledClass,
+    ) -> DeclaredSierraClass:
+        """
+        Declare a sierra contract.
+
+        @param contract_class: sierra compiled contract to be declared
+        @param compiled_class_hash: compiled class hash of the casm compiled contract.
+        @param compiled_class: casm compiled contract to be declared
```
MaksymilianDemitraszek, line 183: Why we have the same `@param` repeated twice, it is some python docsting convention for alternative?

protostar/cheatable_starknet/controllers/contracts.py
cptartur: The previous suggestion would be to raise a specific error here so it can be handled by the cheatcodes. But I guess it doesn't make sense since it cannot be reasonably handled.

protostar/cheatable_starknet/controllers/contracts.py
cptartur: Just my error

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -0,0 +1,77 @@
...
+        def _compile_contract(
+            contract_identifier: ContractIdentifier,
+        ) -> Tuple[CompiledClass, ContractClass]:
+            contract_class = self._project_compiler.compile_contract_to_sierra_from_contract_identifier(
+                contract_identifier
+            )
+            compiled_class = self._project_compiler.compile_contract_to_casm_from_contract_identifier(
+                contract_identifier
+            )
+            return compiled_class, contract_class
```
karol-bisztyga, line 75: is there a good reason to put these two functions inside of the `_build` function? I think they should be just members of the class, they don't capture anything (ok, `declare` captures `_compile_contract` but it could call it with `self.`) and I think nesting like this is not the best practice.

protostar/configuration_file/configuration_file.py
```diff
@@ -106,6 +106,7 @@ def __init__(self, contract_name: str, expected_declaration_location: str):
         super().__init__(
             f"Couldn't find `{contract_name}` in `{expected_declaration_location}`"
         )
+        self.contract_name = contract_name
```
karol-bisztyga, line 109: what is this for?

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
cptartur: I think they can be made separate members.

MaksymilianDemitraszek:
@cptartur Can we prioritize merging this one? It is a heavy blocker 🥺

protostar/configuration_file/configuration_file.py
cptartur: I'm re-raising a `CheatcodeException` from `ContractNameNotFoundException` in declare hint local. I need a contract name for the exception message.

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
cptartur: We can't make this class methods due to #1693 and `_compile_contract` needs to capture self to use `contracts_controller`. So I don't think these can be made just normal functions.

karol-bisztyga: Discussed offline, we can follow up in the issue #1693

protostar/configuration_file/configuration_file.py
karol-bisztyga: right, thanks

karol-bisztyga (approved):
please fix the CI (related #1694)

PR #1637, opened 2023-03-20
Move hooks setup to corresponding HintLocal constructors

abulenok:


karol-bisztyga (approved):

Arcticae (changes requested):
I think this really moves the logic out of controller, but doesn't address the issue itself:
> _"The idea is to move the hooks' assignments to one place where the environment is being set up"_
But really we now have the same thing, just in a different constructor, which i think is even less intuitive (instantiating a `HintLocal` should not have side effects, we should consider `HintLocal` as HOF)

abulenok:
@Arcticae 
My motivation to put it into HintLocals' constructor was that these hooks are integral to the `ExpectEventsHintLocal` and `ExpectCallHintLocal`. Using these HintLocals without setting `_test_finish_hook` would make little sense.

Do you suggest doing smth like `expect_call_hint_local.setFinishHook(_test_finish_hook)` in `CairoSharedHintLocalFactory`?

Also, could you please elaborate on `consider HintLocal as HOF`?

protostar/cheatable_starknet/callable_hint_locals/expect_events_hint_local.py
```diff
@@ -20,8 +21,14 @@ class RawExpectedEvent(TypedDict):
...
 class ExpectEventsHintLocal(CallableHintLocal):
-    def __init__(self, controller: ExpectEventsController):
+    def __init__(
+        self,
+        controller: ExpectEventsController,
+        test_finish_hook: Hook,
+    ):
         self._controller = controller
+        self._test_finish_hook = test_finish_hook
+        self._test_finish_hook.on(self._controller.compare_expected_and_actual_results)
```
MaksymilianDemitraszek, line 31: Wouldn't be actually better to define method `finish_hook` in `HintLocal` and presence of the HintLocal in the vm environment would trigger test runner to run this method for each local at the end? We wouldn't have to pass this convoluted `Hook` object? 🤔 
cc @Arcticae

MaksymilianDemitraszek: Ok then maybe we should just keep those checks in a separate place, and call them from test runner?

abulenok:
@Arcticae @MaksymilianDemitraszek 
The other option I was thinking about is to create a `def get_finish_hooks() -> Optional[List[Callable]]:` in `CairoSharedHintLocalFactory`, but it feels less "tidy" to me. WDYT?

Arcticae (approved):
I guess it's better now

MaksymilianDemitraszek (approved):

PR #1641, opened 2023-03-21
Fetch linked-libraries from Scarb metadata

Radinyn:


protostar/commands/cairo1_commands/fetch_from_scarb.py
```diff
@@ -0,0 +1,89 @@
...
+    except subprocess.CalledProcessError as ex:
+        raise ScarbMetadataFetchException(str(ex)) from ex
+
+    if result.returncode != 0:
+        raise ScarbMetadataFetchException(
+            "Metadata fetch returned a non-zero exit code."
+        )
+
+    # only the last line of the output contains metadata
+    result = result.stdout.decode("utf-8").strip().split("\n")[-1]
```
mkaput, line 49: You shouldn't decode to UTF-8 here. I would suggest sticking with bytes. Reason: broken output from Scarb could yield non-UTF-8 byte sequences in the output; which could happen to be perfectly understandable by JSON parser

protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -79,7 +81,12 @@ async def build(
         relative_cairo_path: Optional[list[Path]] = None,
         target_contract_name: str = "",
     ) -> None:
+
         cairo_path = relative_cairo_path or []
+
+        libraries = maybe_fetch_linked_libraries(self._project_root_path)
+        cairo_path += libraries
```
mkaput, line 88: I think it should be an error if you simultaneously use Scarb and link libraries in protostar.toml

protostar/commands/cairo1_commands/fetch_from_scarb.py
piotmag769: "Changed in version 3.6: s can now be of type [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) or [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray). The input encoding should be UTF-8, UTF-16 or UTF-32."

https://docs.python.org/3/library/json.html#basic-usage

mkaput (changes requested):

MaksymilianDemitraszek (commented):
Please provide e2e test for that before merging

protostar/commands/cairo1_commands/fetch_from_scarb.py
```diff
@@ -124,3 +125,12 @@ def maybe_fetch_linked_libraries(project_root_path: Path) -> Optional[List[Path]
         raise ScarbMetadataFetchException("Error parsing metadata:\n" + str(ex)) from ex
 
     return paths
+
+
+def find_directory_with_cairo_project_toml(lib_cairo_path: Path) -> Path:
+    src_directory_path = lib_cairo_path.parent
+    return (
+        src_directory_path
+        if "cairo_project.toml" in os.listdir(src_directory_path)
```
mkaput, line 134: why??

website/docs/cli-reference.md
```diff
@@ -32,7 +32,7 @@ An output directory used to put the compiled contracts in.
 #### `--contract-name STRING`
 Specify a contract name that should be built.
 #### `--linked-libraries PATH[]`
-Paths to cairo1 modules
+Paths to cairo1 modules. Shouldn't be explicitly provided when managing dependencies using Scarb.
```
mkaput, line 35:
```suggestion
Paths to cairo1 modules. Should not be explicitly provided when managing dependencies using Scarb.
```

It's strange to me when software is using informal english

mkaput (changes requested):

protostar/commands/cairo1_commands/fetch_from_scarb.py
piotmag769: Path should point to the location with cairo_project.toml - some libraries like [quaireaux](https://github.com/keep-starknet-strange/quaireaux/tree/main/quaireaux/math) have cairo_project.toml outside of the src folder.

mkaput: `cairo_project.toml` does not have to match the project structure, it's merely a set of arguments for the compiler saved to a file, and anyway it is a dependency, so you shouldn't touch its structure at all -- all the paths are provided by Scarb already

piotmag769: As discussed offline - we have to do it this way until changes are made in the cairo fork.
Issue: https://github.com/software-mansion/protostar/issues/1704

mkaput (approved):

protostar/commands/cairo1_commands/fetch_from_scarb.py
```diff
@@ -0,0 +1,136 @@
...
+                == "starknet-contract",
+                matching_compilation_units,
+            )
+        )
+        matching_lib_units = list(
+            filter(
+                lambda compilation_unit: compilation_unit["target"]["kind"] == "lib",
+                matching_compilation_units,
+            )
+        )
```
Arcticae, line 94: Please consider list comprehension for better readability

protostar/commands/cairo1_commands/fetch_from_scarb.py
```diff
@@ -0,0 +1,136 @@
...
+            map(
+                lambda component: find_directory_with_cairo_project_toml(
+                    Path(component["source_path"])
+                ),
+                valuable_dependencies,
+            )
+        )
+
+    except (IndexError, KeyError) as ex:
+        raise ScarbMetadataFetchException("Error parsing metadata:\n" + str(ex)) from ex
```
Arcticae, line 125: catching general exceptions in a broad scope may lead to unexpected errors

protostar/commands/cairo1_commands/fetch_from_scarb.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+        )
+        matching_lib_units = list(
+            filter(
+                lambda compilation_unit: compilation_unit["target"]["kind"] == "lib",
+                matching_compilation_units,
+            )
+        )
+
+        #  lib is the default target
```
Arcticae, line 96: This comment seems inaccurate since we're getting contract and falling back to lib?

protostar/commands/cairo1_commands/fetch_from_scarb.py
```diff
@@ -0,0 +1,136 @@
...
+                "Scarb found multiple starknet-contract targets - using the target named %s.",
+                {unit_with_dependencies["target"]["name"]},
+            )
+
+        valuable_dependencies = filter(
+            lambda dependency: dependency["name"] != "core",
+            unit_with_dependencies["components_data"],
+        )
+
+        # using path.parent because component["source_path"] points to lib.cairo
```
Arcticae, line 114: This comment is misplaced, right?

Arcticae (approved):
A few nitpicks, but looks good overall. The logic itself is a bit cryptic to me (as i'm not that familiar with the scarb metadata scheme) but it looks correct.

protostar/commands/cairo1_commands/fetch_from_scarb.py
piotmag769: Agree, but I don't see how can we make it more specific in this case.

piotmag769:
Note: obtaining dependencies using `scarb metadata` can lead to some unexpected behaviours like being able to use modules that shouldn't really be available. For example you could still use `external_lib_foo` [here](https://github.com/software-mansion/protostar/pull/1641/files#diff-05da53e7e37b1393295d0915c5d89411e9214045dc54ef86413aaf70e64d6cea) even if you haven't had specified it [there](https://github.com/software-mansion/protostar/pull/1641/files#diff-96b69f20cf7e403a3e19397d1df66b7b8bf28aeb1ea3ed31926d2d125ba49e61), since `external_lib_bar` dependens on `external_lib_foo`. I guess the reason for this is passing every dependency to `linked-libraries` in protostar, which doesn't provide any distinction between outer and inner dependencies.

Arcticae:
> that shouldn't really be available

If it's in your dependencies, it means you can use it.

protostar/commands/cairo1_commands/fetch_from_scarb.py
```diff
@@ -71,29 +71,24 @@ def maybe_fetch_linked_libraries(project_root_path: Path) -> Optional[List[Path]
...
+        matching_contract_units = [
+            compilation_unit
+            for compilation_unit in matching_compilation_units
+            if compilation_unit["target"]["kind"] == "starknet-contract"
+        ]
+        matching_lib_units = [
+            compilation_unit
+            for compilation_unit in matching_compilation_units
+            if compilation_unit["target"]["kind"] == "lib"
+        ]
```
Arcticae, line 89: ❤️

tests/e2e/cairo1/test_building_contracts_scarb.py
```diff
@@ -0,0 +1,44 @@
...
+
+# TODO #1767: investigate differences in gas support between quaireaux and protostar
+#  building works for some reason even considering the differences mentioned above...
+def test_build_with_online_dependencies_using_scarb(
+    protostar: ProtostarFixture, copy_fixture: CopyFixture
+):
+    execute_build_test("online_dependencies", protostar, copy_fixture)
+
+
+def execute_build_test(
```
Arcticae, line 27: This should be a fixture

tests/e2e/cairo1/conftest.py
```diff
@@ -0,0 +1,21 @@
+import os
+
+from tests.e2e.conftest import ProtostarFixture, CopyFixture
+
+
+def execute_test(
```
Arcticae, line 6: This should be a fixture

piotmag769:
> > that shouldn't really be available
> 
> If it's in your dependencies, it means you can use it.

Right, for some reason I forgot it works this way everywhere.

tests/e2e/cairo1/test_testing_contracts_scarb.py
```diff
@@ -0,0 +1,39 @@
...
+def test_contract_with_libraries_using_scarb(
+    protostar: ProtostarFixture, copy_fixture: CopyFixture
+):
+    execute_test(
+        "libraries",
+        "test_contract_with_libraries_as_dependencies",
+        "tests/test_contract.cairo",
+        protostar,
+        copy_fixture,
+    )
```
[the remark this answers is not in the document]
piotmag769: I decided to use `@pytest.mark.parametrize` in those tests, I think it reduces the boilerplate code by a lot. Tried to make it as easy to modify as possible.

.github/workflows/checks.yml
```diff
@@ -141,6 +141,14 @@ jobs:
       - name: Build
         run: |
           poetry run poe build
+      - name: Install Scarb
+      # TODO: improve to download latest release - maybe https://github.com/marketplace/actions/release-downloader
```
Arcticae, line 145: You left a TODO here without an issue

Arcticae (approved):

.github/workflows/checks.yml
piotmag769: https://github.com/software-mansion/protostar/issues/1814

MaksymilianDemitraszek (changes requested):
Generally looks good! 🔥 

Some high level suggestions:
- Please investigate and write a test for a case we discussed on slack (no protostar.toml)
- We should move some of the tests to integration. I believe e2e test should only validate if e2e flow works (arguments are parsed correctly etc.) . Test validating complex scenarios should be moved to the integration layer
- Can we add a test case testing the scenario with multiple contracts in protostar toml?

MaksymilianDemitraszek:
We should lift the requirement of `has_scarb_toml` from   root

MaksymilianDemitraszek (dismissed):

MaksymilianDemitraszek (approved):

PR #1667, opened 2023-03-24
Use --linked-libraries instead of --cairo-path for build-cairo1

abulenok:


tests/e2e/fixtures/cairo_0_1_mixed/protostar.toml
```diff
@@ -1,7 +1,6 @@
 [project]
 protostar-version = "0.9.1"
 lib-path = "lib"
-linked-libraries = ["src"]
```
Arcticae, line 4: Why was this deleted?

protostar/cli/common_arguments.py
```diff
@@ -167,3 +167,10 @@
     type="str",
     default="",
 )
+
+LINKED_LIBRARIES = ProtostarArgument(
+    name="linked-libraries",
+    value_parser="list",
+    description="Libraries to include in compilation",
```
Arcticae, line 174: This could use a better explanation (paths to cairo1 libraries perhaps)?

tests/e2e/fixtures/cairo_0_1_mixed/protostar.toml
abulenok: `linked-libraries` paths should lead to folders with module declaration file (`lib.cairo`). There is no such file in `src`

abulenok: https://github.com/software-mansion/protostar/actions/runs/4513087426/jobs/7947598500#step:8:383

protostar/cli/common_arguments.py
abulenok: updated

Arcticae (approved):

PR #1668, opened 2023-03-24
Use Cairo tests configs

karol-bisztyga:
In order to properly handle `get_gas` in the unit tests we should handle the test config that comes from the cairo code.
Here https://github.com/software-mansion-labs/cairo/blob/7628ee34558bb7d89a9c091542bcf5fa25072f31/crates/cairo-lang-test-runner/src/test_config.rs#L42 is how the attributes are read from the cairo code. We need to read `available_gas` but there are two more: `should_panic` and `ignore`.

Open question: Should we use `should_panic`?

Also, I couldn't find any usages of `#[ignore]` unfortunately, also, should we use that one?

This PR is a draft, more like a POC as I'm not sure if this approach is any good.

### DIFF
--- a/protostar/cairo/cairo_bindings.py
+++ b/protostar/cairo/cairo_bindings.py
@@ -10,6 +10,9 @@ from protostar.cairo.cairo_function_runner_facade import RUNNER_BUILTINS_TITLE_C
 from .cairo_exceptions import CairoBindingException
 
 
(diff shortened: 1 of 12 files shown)


MaksymilianDemitraszek:
I think we merged it in another PR, plz reopen if not

PR #1669, opened 2023-03-24
Change scope of errors reraised in bindings

karol-bisztyga:
I discovered that there are errors that are thrown from rust and are not `RuntimeError`. The result was, in my case, an infinite hang which is very bad. The error was something like `pyo3_runtime.PanicException`. Are there any objections to just catching everything that comes from rust (I mean errors) and turning it into `CairoBindingException`?


Arcticae (commented):
What are the different cases of those exceptions being thrown? Have you found a pattern? If not, can we catch 2 exception types instead of broadening the exception type?

karol-bisztyga:
I cannot reproduce this now, I should've documented this better, sorry about that.

PR #1670, opened 2023-03-27
Improve error messages for declare_cairo0

MaksymilianDemitraszek:
### Checklist
- [x] Docs updated/issue for docs created
- [x] Added relevant tests

### DIFF
--- a/protostar/cheatable_starknet/callable_hint_locals/declare_cairo0_hint_local.py
+++ b/protostar/cheatable_starknet/callable_hint_locals/declare_cairo0_hint_local.py
@@ -7,9 +7,11 @@ from protostar.cheatable_starknet.controllers.contracts import (
 )
 from protostar.compiler import ProjectCompiler
 from protostar.cairo.short_string import CairoShortString, short_string_to_str
-
+from protostar.configuration_file.configuration_file import ContractNameNotFoundException
+from protostar.starknet.cheatable_starknet_exceptions import CheatcodeException
 from .callable_hint_local import CallableHintLocal
 
+from starkware.cairo.lang.compiler.parser_transformer import ParserError
 
 class DeclareCheatcodeProtocol(Protocol):
     def __call__(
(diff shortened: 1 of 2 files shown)


protostar/cheatable_starknet/callable_hint_locals/declare_cairo0_hint_local.py
```diff
@@ -38,11 +43,17 @@ def _build(self) -> Any:
...
+        try:
+            compiled_contract = (
+                self._project_compiler.compile_contract_from_contract_identifier(
+                    contract_str
+                )
             )
-        )
+        except ContractNameNotFoundException as err:
+            raise CheatcodeException(self, err.message) from err
+        except ParserError as err:
```
cptartur, line 54: Not sure why are we catching `ParserError` here. `ProjectCompiler` should throw `CompilationException` if there's a problem with compilation (though the catching logic seems incomplete there so we need some additional catching/updating it anyways).

MaksymilianDemitraszek: it seems like a bug here

MaksymilianDemitraszek: This will be tightly coupled with this change so I will fix this bug in this PR

protostar/cheatable_starknet/callable_hint_locals/declare_cairo0_hint_local.py
```diff
@@ -38,11 +43,17 @@ def _build(self) -> Any:
...
+            raise CheatcodeException(self, err.message) from err
```
karol-bisztyga, line 55:
```suggestion
        except (ContractNameNotFoundException, ParserError) as err:
            raise CheatcodeException(self, err.message) from err
```

MaksymilianDemitraszek:
Opened a issue, closing for now

PR #1678, opened 2023-03-28
Adapt CLI to Starknet 0.11.0

### LINKED ISSUE #1654: Adapt protostar cli to Starknet 0.11
We need to provide a cli equivalent with Starknet's cli

It might be splitted into separate tasks

cptartur:
This PR closes #1654

- [x] declare-cairo1 command 
- [x] integration tests
- [x] e2e tests
- [x] calculating `compiled_class_hash`


cptartur:
This is ready for initial round of reviews, but until we sync cairo compiler, e2e test swill fail because of incompatible artifacts.

tests/e2e/test_gateway_commands.py
Arcticae, line 1: We should be still able to declare old contracts. The test should be added on top of existing one

protostar/starknet_gateway/gateway_facade.py
```diff
@@ -203,6 +206,74 @@ def _load_compiled_contract(self, compiled_contract_path: Path) -> str:
...
+            )
+        except ClientError as ex:
+            account_address_found_in_message = hex(int(account_address)) in ex.message
+            message = (
+                "No account associated with provided account address found. Contact your wallet provider."
+                if "StarknetErrorCode.UNINITIALIZED_CONTRACT" in ex.message
+                and account_address_found_in_message
+                else ex.message
+            )
+            raise ProtostarException(message) from ex
```
Arcticae, line 268: This logic is repeated, right? It should be refactored IMHO if it's copied over

protostar/commands/declare/declare_messages.py
```diff
@@ -0,0 +1,30 @@
+from dataclasses import dataclass
+from typing import Optional
+
+from protostar.io import StructuredMessage, LogColorProvider
+from protostar.starknet_gateway import SuccessfulDeclareResponse
+
+
+@dataclass
+class SuccessfulDeclareMessage(StructuredMessage):
```
Arcticae, line 9: I think we should perhaps distinguish the two declare txns somehow in the console (V1 vs V2)

MaksymilianDemitraszek:
Can we link an issue 🙏

tests/e2e/test_gateway_commands.py
cptartur: There's still test for that above: `test_declaring_contract`. For whatever reason there were both `test_declaring_contract_with_signature` and `test_declaring_contract` which seem exactly the same except for one argument order.

Perhaps the leftover from times where declare without signing was possible?

cptartur:
> Can we link an issue 🙏

It's linked. The target branch is not master though, so it won't show up in the milestone.

cptartur:
> > Can we link an issue 🙏
> 
> It's linked. The target branch is not master though, so it won't show up in the milestone.

Manually added it along the comment in the description. It should show now.

protostar/commands/declare/declare_messages.py
cptartur: Not sure if this is necessary: these are different version of the same transaction.

tests/integration/commands/declare_command_test.py
```diff
@@ -26,6 +34,67 @@ def compiled_contract_path_fixture() -> Path:
...
+    class MockedProjectCompiler(ProjectCompiler):
+        def __init__(self):
+            super().__init__(MagicMock(), MagicMock())
+
+        def compile_contract_to_sierra_from_contract_name(
+            self, *args: Any, **kwargs: Any
+        ):
+            # pylint: disable=unused-argument
+            compiled = cairo_bindings.compile_starknet_contract_to_sierra_from_path(
+                input_path=datadir / "minimal.cairo",
```
cptartur, line 52: For whatever reason declaring more complicated contracts fails on devnet i.e. when I replace `minimal.cairo` with `balance.cairo` it fails.

Trying to figure out if this is devet, protostar, or starknet.py issue.

cptartur: Turned out it was local setup problem. Devnet has to have a correct path to cairo1 compiler main Cargo.toml set up for declare cairo1 contracts to work correctly.

Arcticae (approved):

tests/integration/commands/declare_command_test.py
```diff
@@ -26,6 +34,67 @@ def compiled_contract_path_fixture() -> Path:
     return Path("./build/main.json")
 
 
+async def test_declaring_cairo1_contract(
```
Arcticae, line 37: We actually should a test in which the compiled class hash doesn't match the actual one, and check what happens there. Can be done in a separate issue, but in this release.

tests/integration/commands/declare_command_test.py
```diff
@@ -26,6 +34,67 @@ def compiled_contract_path_fixture() -> Path:
...
+
+        def compile_contract_to_casm_from_contract_name(
+            self, *args: Any, **kwargs: Any
+        ):
+            # pylint: disable=unused-argument
+            compiled = cairo_bindings.compile_starknet_contract_to_casm_from_path(
+                input_path=datadir / "balance.cairo",
+            )
+            assert compiled is not None
+            return compiled
```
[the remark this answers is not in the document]
cptartur: This cannot be reasonably reused between tests. I'd leave it here, it's part of the test setup

tests/integration/commands/declare_command_test.py
Arcticae: The behavior from user's perspective is important, i think it's a good case for e2e test. UX is something we really want to have done right on declare, since it's a pain point in 0.11

cptartur: Okay, I'll introduce the test in this PR since it's related.

cptartur: Actually it doesn't make sense to test this. `compiled_class_hash` was a leftover argument in the tests. It is not being accepted by the command itself.

protostar/commands/declare/declare_messages.py
cptartur: #1724

PR #1687, opened 2023-03-29
Check cheatcodes return values

karol-bisztyga:
This is built on the top of https://github.com/software-mansion/protostar/pull/1682

Related Cairo PR https://github.com/software-mansion-labs/cairo/pull/51

### DIFF
--- a/tests/integration/cairo1_compiler_bindings/library_functions_test.py
+++ b/tests/integration/cairo1_compiler_bindings/library_functions_test.py
@@ -8,36 +8,67 @@ from protostar.cairo.cairo1_test_suite_parser import (
 from protostar.cairo.cairo_function_runner_facade import CairoRunnerFacade
 
 
+def extract_test_case_name(full_test_case_name: str):
+    return full_test_case_name.split("::")[-1]
+
+
 def get_mock_for_lib_func(
     lib_func_name: str,
     err_code: int,
     cairo_runner_facade: CairoRunnerFacade,
     test_case_name: str,
     args_validator: Optional[Callable] = None,
+    return_values_provider: Optional[Callable] = None,
 ):
     if lib_func_name in ["declare", "declare_cairo0"]:
-        ok = type("ok", (object,), {"class_hash": 0})()
+        assert return_values_provider
+        ok = type(
+            "ok",
+            (object,),
+            {"class_hash": return_values_provider(test_case_name)["class_hash"]},
+        )()
         return_value = type(
             "return_value", (object,), {"err_code": err_code, "ok": ok}
         )()
     elif lib_func_name == "deploy_tp":
-        ok = type("ok", (object,), {"deployed_contract_address": 123})()
+        assert return_values_provider
+        ok = type(
+            "ok",
+            (object,),
+            {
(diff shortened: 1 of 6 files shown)


tests/integration/cairo1_compiler_bindings/library_functions_test.py
```diff
@@ -151,19 +215,42 @@ def _args_validator(test_case_name: str, *args: Any, **kwargs: Any):
 
 def test_prepare(datadir: Path):
     expected_calldatas = {
-        "test_prepare": [101, 202, 303, 405, 508, 613, 721],
+        "test_prepare": [101, 202, 613, 721],
         "test_prepare_tp": [3, 2, 1],
         "test_prepare_no_args": [],
     }
+    return_values = {
```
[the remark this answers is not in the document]
karol-bisztyga: I know but I consider this as better looking, you have this structure now that you have all the "configs" in the beginning, then the callbacks that use these configs, and then a final invocation. I'd stay with what is but if you have a strong opinion I don't mind moving it.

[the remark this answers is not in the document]
karol-bisztyga: Ok, I will resolve this since you said it's a nitpick and "not a deal-breaker".
If you feel like this should be addressed, feel free to open an issue or something, thanks!

tests/integration/cairo1_compiler_bindings/library_functions_test/prepare_test.cairo
```diff
@@ -5,13 +5,19 @@ fn test_prepare() {
    let mut arr = ArrayTrait::new();
    arr.append(101);
    arr.append(202);
-   arr.append(303);
-   arr.append(405);
-   arr.append(508);
    arr.append(613);
    arr.append(721);
    match prepare_tp(123, arr) {
```
piotmag769, line 10: We should use prepare instead of prepare_tp here. Will address it in separate PR and create issue.

tests/integration/cairo1_compiler_bindings/library_functions_test/prepare_test.cairo
```diff
@@ -40,7 +54,12 @@ fn test_prepare_tp() {
 fn test_prepare_no_args() {
    let mut arr = ArrayTrait::new();
    match prepare_tp(123, arr) {
```
piotmag769, line 56: We should use prepare instead of prepare_tp here. Will address it in separate PR and create issue.

piotmag769, line 56: We should use prepare instead of prepare_tp here. Will address it in separate PR and create issue.

piotmag769: https://github.com/software-mansion/protostar/pull/1705

tests/integration/cairo1_compiler_bindings/library_functions_test/prepare_test.cairo
piotmag769: https://github.com/software-mansion/protostar/pull/1705

PR #1688, opened 2023-03-29
`tmp_path` fixtures wrappers

Arcticae:
This is pretty brute-force, i'll explain what it does and why:

## The cause
`tmp_path` and `tmp_path_factory` are not actually deleted after a test case/module/test session run.
This causes to clog up the disk and fail our CI

## This PR does:
- reduce scope of some protostar-fixtures to function, to keep things contain and easy to manage from fixture
- wraps pytests fixures into our own, which are deleted after a case run

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,13 +1,15 @@
 import json
+import shutil
 import subprocess
 import time
 from contextlib import contextmanager
 from pathlib import Path
 from socket import socket as Socket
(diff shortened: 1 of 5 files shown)


karol-bisztyga (commented):
looks ok but I'm not a pytest expert, would be good to get another review

MaksymilianDemitraszek (approved):
That is actually cool as hell! Amazing job ❤️

PR #1702, opened 2023-03-31
Prepare hint local

Arcticae:


piotmag769 (approved):

tests/integration/cairo1_hint_locals/prepare/data/minimal_with_args.cairo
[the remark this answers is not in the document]
Arcticae: Good catch :)

tests/integration/cairo1_hint_locals/prepare/prepare_test.cairo
```diff
@@ -0,0 +1,21 @@
...
+    let prepared_contract = prepare(class_hash, calldata).unwrap();
+    drop(prepared_contract);
+}
+
+#[test]
+fn test_prepare_with_args() {
+    let class_hash = declare('minimal_with_args').unwrap();
+
+    let mut calldata = ArrayTrait::new();
+    calldata.append('name');
```
MaksymilianDemitraszek, line 18: Can we include few more arguments here? Possibly with more complex types. I believe cairo1 supports something like `into` which lets user dump the data.

MaksymilianDemitraszek (approved):

protostar/cheatable_starknet/callable_hint_locals/prepare_hint_local.py
```diff
@@ -22,37 +21,37 @@ def __init__(self, contracts_controller: ContractsController):
...
 
     def _build(self) -> Callable[[Any], Any]:
         return self.prepare
 
     def prepare(
         self,
-        declared: DeclaredContract,
-        constructor_calldata: Optional[CairoOrPythonData] = None,
-        salt: Optional[int] = None,
+        class_hash: int,
```
karol-bisztyga, line 31: Why did you rename this? I think we should be consistent, in all other places it is called `constructor_calldata`.

karol-bisztyga (approved):

protostar/cheatable_starknet/callable_hint_locals/prepare_hint_local.py
piotmag769: Arguments must have these exact names due to the way we translate this cheatcode to casm in compiler (it is passed as **kwargs in protostar code).
https://github.com/software-mansion-labs/cairo/blob/main/crates/cairo-lang-casm/src/hints/mod.rs#L610

MaksymilianDemitraszek:
Can we also add respective tests for cairo0 contract?

tests/integration/cairo1_hint_locals/prepare/prepare_test.cairo
Arcticae: Will add more complex test in a next PR

Arcticae:
https://github.com/software-mansion/protostar/issues/1715

Issue #1684, opened 2023-03-29
Produce both sierra json and casm when building a project


cptartur:
When running `protostar build-cairo1`, both sierra json and casm compiled contracts should be produced.

Outputs should be exactly equivalent to running
`cargo run --bin starknet-compile -- /path/to/input.cairo /path/to/output.json` and `cargo run --bin starknet-sierra-compile -- /path/to/input.json /path/to/output.casm` in new cairo compiler.

MaksymilianDemitraszek:
This is not a critical feature as declare will handle compilation under the hood, I am moving it back to next milestone. Let me know it that is an issue

PR #1722, opened 2023-04-05
Produce both sierra and casm outputs when build-cairo1

abulenok:
closes https://github.com/software-mansion/protostar/issues/1684

### DIFF
--- a/protostar/commands/cairo1_commands/build_cairo1_command.py
+++ b/protostar/commands/cairo1_commands/build_cairo1_command.py
@@ -67,14 +67,14 @@ class BuildCairo1Command(ProtostarCommand):
             f"Multiple files found for contract {contract_name}, "
             f"only one file per contract is supported in cairo1!"
         )
+        sierra_file_path = output_dir / (contract_name + ".sierra.json")
         cairo1.compile_starknet_contract_to_sierra_from_path(
             input_path=contract_paths[0],
             cairo_path=cairo_path,
-            output_path=output_dir / (contract_name + ".sierra.json"),
+            output_path=sierra_file_path,
         )
-        cairo1.compile_starknet_contract_to_casm_from_path(
-            input_path=contract_paths[0],
-            cairo_path=cairo_path,
+        cairo1.compile_starknet_contract_sierra_to_casm_from_path(
+            input_path=sierra_file_path,
             output_path=output_dir / (contract_name + ".casm.json"),
         )
 
--- a/protostar/commands/cairo1_commands/build_cairo1_command.py
+++ b/protostar/commands/cairo1_commands/build_cairo1_command.py
@@ -67,10 +67,15 @@ class BuildCairo1Command(ProtostarCommand):
             f"Multiple files found for contract {contract_name}, "
             f"only one file per contract is supported in cairo1!"
         )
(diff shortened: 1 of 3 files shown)


protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -67,10 +67,15 @@ async def _build_contract(
...
+        cairo1.compile_starknet_contract_to_sierra_from_path(
+            input_path=contract_paths[0],
+            cairo_path=cairo_path,
+            output_path=output_dir / (contract_name + ".sierra.json"),
+        )
         cairo1.compile_starknet_contract_to_casm_from_path(
             input_path=contract_paths[0],
             cairo_path=cairo_path,
-            output_path=output_dir / (contract_name + ".json"),
+            output_path=output_dir / (contract_name + ".casm.json"),
```
Arcticae, line 78: Can we not use the sierra as an input here? Seems like a waste of time to compile all the way from cairo

Arcticae (dismissed):

protostar/commands/cairo1_commands/build_cairo1_command.py
abulenok: There is no `starknet_sierra_to_casm` function in `bindings`

abulenok: As the whole compilation happens quickly enough, we have decided that we can leave it as is for now (discussed in slack)

karol-bisztyga: I don't think there was an agreement that the compilation is quick enough. From what I understand what happened was: We agreed that if we don't need an additional method, we shouldn't add it. Then you said that the compilation is fast so we don't need it. Looking at this code, I'd reconsider adding this shortcut method `sierra -> casm`. What happens here now is we do the same thing (compile `cairo -> sierra`) twice for no good reason while we could do this just once. I don't see why we should do so.

tests/e2e/cairo_0_1_mixed/test_cairo_0_1_mixed_building.py
```diff
@@ -37,8 +41,8 @@ def test_cairo1_build_with_contract_names_build_together(
     assert build_path.exists()
     built_files = list(build_path.glob("*"))
     assert set(built_files) == set(
-        Path("build") / (contract + ".json") for contract in contracts
-    )
+        Path("build") / (contract + ".casm.json") for contract in contracts
+    ).union({Path("build") / (contract + ".sierra.json") for contract in contracts})
```
karol-bisztyga, line 45: I know this is a nit but please note that we go through the `contracts` collection twice which is inefficient. I think we can go ahead and use a standard loop instead of comprehensions.

karol-bisztyga (changes requested):

protostar/commands/cairo1_commands/build_cairo1_command.py
karol-bisztyga: I created a cairo PR for this https://github.com/software-mansion-labs/cairo/pull/61

karol-bisztyga (dismissed):
image

karol-bisztyga (approved):

PR #1726, opened 2023-04-06
Cairo1 call impl

karol-bisztyga:

PR #1733, opened 2023-04-06
Cairo1 invoke impl

### LINKED ISSUE #1616: Add `invoke` cheatcode implementation in cairo1
We need to make call cheatcode working in cairo1 and add integration tests for it. If tests fails, make it work.

This task being finished means
- test which calls contract method passes
- test which fails to invoke a contract provides a proper error code

Arcticae:
Built on top of #1726
Closes #1616


karol-bisztyga (approved):
It looks good but please address what I mentioned in the inline comment.

PR #1772, opened 2023-04-14
Add deploy Cairo 1 docs

karol-bisztyga:


Arcticae (approved):

website/docs/tutorials/08-cairo-1-support/03-cheatcodes/deploy.md
```diff
@@ -0,0 +1,38 @@
+# `deploy`
+
+```cairo
+fn deploy(prepared_contract: PreparedContract) -> Result::<felt252, felt252> nopanic;
+```
+
+Deploys a [preapred](./prepare.md) contract.
+
+- `prepared_contract` - an object of the struct `PreparedContract` that consists of the following fields:
```
piotmag769, line 9: Do you think it is a good idea to link a corelib file with PreparedContract definition here?

piotmag769 (approved):

website/docs/tutorials/08-cairo-1-support/03-cheatcodes/deploy.md
karol-bisztyga: It seems I put this in the Q and didn't see this comment before it was merged. If you feel like this should be done, feel free to post an issue (if you decide to do this, an instant PR would be the best probably).

piotmag769: I guess it is not that important, users can figure it out looking at the example code below.

PR #1790, opened 2023-04-18
Plugged in `available_gas`

Radinyn:


protostar/cairo_testing/cairo1_test_collector.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, cairo_path: list[str]):
     def collect_cairo1_tests_and_cache_outputs(
         self,
         file_path: Path,
-    ) -> List[str]:
+    ) -> List[Tuple[str, cairo1.AvailableGas]]:
```
karol-bisztyga, line 29: why don't we use `tuple`?

protostar/testing/test_collector.py
```diff
@@ -77,7 +77,9 @@ def fnmatch_any(name: str, pats: Iterable[str]) -> bool:
 
 
 class FunctionNameGetter(Protocol):
-    def __call__(self, file_path: Path) -> List[str]:
+    def __call__(
+        self, file_path: Path
+    ) -> Union[List[str], List[Tuple[str, AvailableGas]]]:
```
karol-bisztyga, line 82:
```suggestion
    ) -> Union[list[str], list[tuple[str, AvailableGas]]]:
```

protostar/testing/test_collector.py
```diff
@@ -275,12 +277,25 @@ def _build_test_suite_from_test_suite_info(
...
+            names: List[str] = []
+            gas: List[AvailableGas] = []
+        elif isinstance(function_names[0], str):
+            function_names = cast(List[str], function_names)
+            names: List[str] = function_names
+            gas: List[AvailableGas] = [None] * len(function_names)
+        else:
+            function_names = cast(List[Tuple[str, AvailableGas]], function_names)
+            names: List[str] = [fn[0] for fn in function_names]
+            gas: List[AvailableGas] = [fn[1] for fn in function_names]
```
karol-bisztyga, line 291:
```suggestion
        names: list[str] = []
        gas: list[AvailableGas] = []
        if len(function_names) == 0:
            pass
        elif isinstance(function_names[0], str):
            function_names = cast(list[str], function_names)
            names = function_names
            gas = [None] * len(function_names)
        else:
            function_names = cast(List[Tuple[str, AvailableGas]], function_names)
            names = [fn[0] for fn in function_names]
            gas = [fn[1] for fn in function_names]
```
This is only a quick idea, the point is that you declare types for the same variables multiple times which makes it worse for reading I think.

protostar/testing/test_collector.py
```diff
@@ -275,12 +277,25 @@ def _build_test_suite_from_test_suite_info(
...
+
+        setup_fn_name = self._collect_setup_hook_name(names)
 
         test_cases = list(
             test_suite_info.filter_test_cases(
                 self._collect_test_cases(
-                    function_names=[(fn, None) for fn in function_names],
+                    function_names=list(zip(names, gas)),
```
karol-bisztyga, line 298: This `zip` is not needed for the `else` case, right? Because you already have them zipped. If I understood correctly. If so, then, for the `else` case what you do is first unzip and zip it back, right? In such a case I think it would be better to zip only for the `elif` case and for the `else` case just assign to the `zip` the value you already have.

karol-bisztyga (changes requested):

protostar/cairo_testing/cairo1_test_collector.py
Radinyn: I just got used to pre Python 3.9 syntax, it's used everywhere in this module as well

protostar/testing/test_collector.py
Radinyn: I don't think there's that much of a difference

Arcticae: With that you can actually omit the first `if`, it's not needed. Moreover, use `function_names` directly in the condition as an empty list is falsy

protostar/testing/test_collector.py
```diff
@@ -275,12 +277,25 @@ def _build_test_suite_from_test_suite_info(
...
-        setup_fn_name = self._collect_setup_hook_name(function_names)
+
+        if len(function_names) == 0:
+            names: List[str] = []
+            gas: List[AvailableGas] = []
+        elif isinstance(function_names[0], str):
+            function_names = cast(List[str], function_names)
+            names: List[str] = function_names
+            gas: List[AvailableGas] = [None] * len(function_names)
+        else:
```
Arcticae, line 288: we should still check for instance here and throw instead of casting

Arcticae (dismissed):

protostar/cairo_testing/cairo1_test_collector.py
karol-bisztyga: From what I discussed with others quite a while ago, we should move away from using the types from `typing` and use built-ins wherever possible. 


> it's used everywhere in this module as well

Not true, it is not used in `cairo_test_execution_state.py`.

I think we should use the built-ins when changing the code / adding a new code.

protostar/testing/test_collector.py
karol-bisztyga: This function has ~20 lines, and having one reference to a variable will not ruin readability. I guess you probably will not be convinced by whatever I say, @Arcticae can you share your opinion so we can move on?

Arcticae: I think what's happenning actually is that this function does two things depending on the type and that's what causing the problems. As for the style really it's not that big of an issue, i like the code as-is but i can see your point @karol-bisztyga

protostar/testing/test_collector.py
Radinyn: type checker screams that checking for instance here is unnecessary, yet it still needs the cast to determine the type

protostar/testing/test_collector.py
Radinyn: Done

protostar/testing/test_collector.py
Radinyn: Done

protostar/cairo_testing/cairo1_test_collector.py
Radinyn: Done

protostar/testing/test_collector.py
karol-bisztyga: is it, can you recheck? I think it's still the way it was.

Radinyn: oops, changed it in a different place, now it's ready

Radinyn: we should consider moving from this typing in all modules, perhaps in a single issue

karol-bisztyga (approved):
lgtm

PR #1797, opened 2023-04-19
Output `class_hash` and `compiled_class_hash` when build cairo1

### LINKED ISSUE #1685: Output `compiled_class_hash`es when building cairo1 project
They should be printed to the output. We can also consider storing them in some kind of file format.

### LINKED ISSUE #1686: Output `class_hash`es when building cairo1 project
They should be printed to the output. We can also consider storing them in some kind of file format.

abulenok:
- [x] Add https://github.com/software-mansion/protostar/pull/1905#discussion_r1192154950


protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -12,6 +24,24 @@
 import protostar.cairo.cairo_bindings as cairo1
 
 
+def compute_class_hash_from_path(sierra_contract_file_path: Path):
+    with open(sierra_contract_file_path, mode="r", encoding="utf-8") as file:
+        sierra_compiled = json.loads(file.read())
+        sierra_compiled.pop("sierra_program_debug_info", None)
+        sierra_compiled["abi"] = json.dumps(sierra_compiled["abi"])
```
Arcticae, line 31: Is ABI necessary for this to work?

protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -67,15 +97,32 @@ async def _build_contract(
...
+
+        sierra_compiled_contract_path = output_dir / (contract_name + ".sierra.json")
         cairo1.compile_starknet_contract_to_sierra_from_path(
             input_path=contract_paths[0],
             cairo_path=cairo_path,
-            output_path=sierra_file_path,
+            output_path=sierra_compiled_contract_path,
+        )
+        class_hash = compute_class_hash_from_path(sierra_compiled_contract_path)
+        logging.info(
```
Arcticae, line 108: the command should use `messenger` pattern, to also support json output

Arcticae (changes requested):

protostar/commands/cairo1_commands/build_cairo1_command.py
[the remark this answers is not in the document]
abulenok: What is the reason to drop ABI here?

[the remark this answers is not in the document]
abulenok: An error occurs, if the ABI is dropped

```
marshmallow.exceptions.ValidationError: {'abi': ['Missing data for required field.']}
```

Arcticae (dismissed):

karol-bisztyga (commented):
I'm going to do a very high-level review for now. I think we should definitely try to reuse the logic from the `declare` hint local. 

Have you considered it/looked at it?

There's that function used there `starkware.starknet.core.os.contract_class.compiled_class_hash.compute_compiled_class_hash`, using it would be cool as we wouldn't have to write additional code for this.

I think a lot of stuff can be copied from `declare` at least (if that works, we should extract mutual logic somewhere so it can be reused in both places (here and in `declare`).

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -64,31 +67,16 @@ def declare(contract: int) -> DeclaredContract:
...
-            sierra_compiled = self._project_compiler.compile_contract_to_sierra_from_contract_identifier(
+            sierra_compiled, casm_compiled = self._project_compiler.compile_contract(
                 contract_identifier
             )
 
-            sierra_compiled = json.loads(sierra_compiled)
-            sierra_compiled.pop("sierra_program_debug_info", None)
-            sierra_compiled["abi"] = json.dumps(sierra_compiled["abi"])
+            contract_class = make_contract_class(sierra_compiled)
+            compiled_class = make_compiled_class(casm_compiled)
```
karol-bisztyga, line 78:
```suggestion
            return make_contract_class(sierra_compiled), make_compiled_class(casm_compiled)
```

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -64,31 +67,16 @@ def declare(contract: int) -> DeclaredContract:
...
 
-            return ContractClass.load(sierra_compiled)
-
-        def _make_compiled_class(contract_identifier: ContractIdentifier):
-            casm_compiled = self._project_compiler.compile_contract_to_casm_from_contract_identifier(
-                contract_identifier
-            )
-
-            return CompiledClass.loads(casm_compiled)
+            return compiled_class, contract_class
```
karol-bisztyga, line 80:
```suggestion

```

protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -1,30 +1,92 @@
...
+class SuccessfulBuildCairo1Message(StructuredMessage):
+    contract_name: str
+    class_hash: int
+    compiled_class_hash: int
+
+    def format_human(self, fmt: LogColorProvider) -> str:
+        lines: list[str] = [
+            "Building cairo1 contracts",
+            f'Class hash for contract "{self.contract_name}": {hex(int(self.class_hash))}',
+            f'Compiled class hash for contract "{self.contract_name}": {hex(int(self.compiled_class_hash))}',
```
karol-bisztyga, line 62: why do you cast to `int` variables that are already `int`s?

protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -1,30 +1,92 @@
...
+            "Contracts built successfully",
+        ]
+
+        return "\n".join(lines)
+
+    def format_dict(self) -> dict:
+        return {
+            self.contract_name: {
+                "class_hash": hex(int(self.class_hash)),
+                "compiled_class_hash": hex(int(self.compiled_class_hash)),
```
karol-bisztyga, line 72: why do you cast to int variables that are already `int`s?

protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -36,59 +98,65 @@ def name(self) -> str:
 
     @property
     def description(self) -> str:
-        return "Compile cairo1 contracts."
+        return "Compile cairo1 contracts. Writes `class_hash` and `compiled_class_hash` to the files and to stdout."
```
karol-bisztyga, line 101: I'm not sure that this is necessary tbh.

protostar/compiler/project_compiler.py
```diff
@@ -29,42 +48,55 @@ def __init__(
...
+        self._casm_file_extension = ".casm.json"
 
-    def compile_contract_to_sierra_from_contract_name(self, contract_name: str) -> str:
-        return self._compile_cairo1_contract_from_contract_name(
-            contract_name=contract_name, output_type=_OutputType.SIERRA
+    def compile_contract(
+        self,
+        contract_identifier: ContractIdentifier,
+        cairo_path: Optional[list[Path]] = None,
+        output_path: Optional[Path] = None,
```
karol-bisztyga, line 58: these repetitive args look like some boilerplate, maybe we could replace them with a data class or something?

tests/integration/_conftest/protostar_fixture.py
```diff
@@ -291,6 +291,12 @@ def _prepare_build_args(self):
         args.contract_name = ""
         return args
 
+    def _prepare_build_cairo1_args(self):
+        args = self._prepare_build_args()
+        delattr(args, "cairo_path")
+        args.linked_libraries = None
+        return args
```
karol-bisztyga, line 298: can you explain this change a bit?

karol-bisztyga:
> As of this method:
> 
> https://github.com/software-mansion/protostar/blob/1ab98364d2bfab4681541fd0ac3ba78f0df58af9/protostar/compiler/project_compiler.py#L54-L66
> 
> it is better to compile to `casm` using `sierra_compiled` string.
> 
> But I think this should be added in a separate PR, along with the tests for the to-be-created function in `cairo_python_bindings`. This PR is already big enough and the proposed refactor is not directly related with `class_hash` and `compiled_class_hash`.

This looks like a tiny change really, maybe we could add it here?

protostar/compiler/project_compiler.py
Arcticae: i disagree on this one, not that many args here, it would make things less readable

tests/e2e/cairo1/test_cairo1_building.py
```diff
@@ -9,14 +10,40 @@
...
+def test_cairo1_build_json(protostar: ProtostarFixture, copy_fixture: CopyFixture):
+    copy_fixture("cairo1_build", "./cairo_project")
+    os.chdir("./cairo_project")
+    result_json = protostar(["build-cairo1", "--json"])
+
+    output_json_parsed = json.loads(
+        result_json.split("\n")[
+            3 if "This project expects older Protostar" in result_json else 0
+        ]
+    )
```
Arcticae, line 42: Isn't this a bug? 
`This project expects older Protostar`
We shouldn't output this kind of info in plaintext if user expects ndjson on the output, right?

abulenok: But how the user will know he is using the wrong version of Protostar? Using the wrong version could result in unexpected behavior

protostar/compiler/project_compiler.py
abulenok: I also don't think there are too many args here. Grouping them would be "artificial"

tests/e2e/cairo1/test_cairo1_building.py
cptartur: I think ndjson spec allows returning multiple jsons separated by newline character. Perhaps version warning should be returned as a separate json?

tests/integration/cairo1/compiler_bindings/starknet_contracts_compiler_test.py
```diff
@@ -6,19 +6,27 @@
...
-        datadir / "test_starknet_project"
+        input_path=datadir / "test_starknet_project"
+    )
+    assert cairo1.compile_starknet_contract_to_casm_from_path(
+        input_path=datadir / "basic_starknet_project",
+        output_path=datadir / "basic_starknet_project.casm",
+    )
+    assert cairo1.compile_starknet_contract_to_casm_from_path(
+        input_path=datadir / "test_starknet_project",
+        output_path=datadir / "test_starknet_project.casm",
```
cptartur, line 20: These tests don't really verify if output_path contains anything useful. Is that on purpose?

tests/integration/commands/declare_cairo1_command_test.py
```diff
@@ -46,6 +46,16 @@ def compile_contract_to_casm_from_contract_name(
             assert compiled is not None
             return compiled
```
cptartur, line 47: Are this and method above still used by the test? If not, parts of this mock could be removed.

protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -36,59 +98,65 @@ def name(self) -> str:
...
 
     @property
     def arguments(self):
         return [
+            *MessengerFactory.OUTPUT_ARGUMENTS,
```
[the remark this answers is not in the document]
abulenok: It doesn't work like that. For instance, look at the `test-cairo1` command

protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -36,59 +98,65 @@ def name(self) -> str:
...
         )
-        assert contract_paths, f"No contract paths found for {contract_name}!"
-        assert len(contract_paths) == 1, (
-            f"Multiple files found for contract {contract_name}, "
-            f"only one file per contract is supported in cairo1!"
+
+        cairo_path = linked_libraries + maybe_fetch_linked_libraries_from_scarb(
+            package_root_path=contract_paths[0],
+            linked_libraries=linked_libraries,
         )
```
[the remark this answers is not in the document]
abulenok: The same assert is available in ProjectCompiler _compile_cairo1_contract_from_contract_name

protostar/compiler/project_compiler.py
karol-bisztyga: ok, so be it

protostar/commands/cairo1_commands/build_cairo1_command.py
```diff
@@ -1,30 +1,92 @@
...
+
+
+def compute_compiled_class_hash_from_casm_code(casm_compiled: str, output_path: Path):
+    compiled_class = make_compiled_class(casm_compiled)
+    compiled_class_hash = compute_compiled_class_hash(compiled_class)
+
+    with open(output_path, mode="w", encoding="utf-8") as output_file:
+        output_file.write(f"{hex(int(compiled_class_hash))}")
+
+    return compiled_class_hash
```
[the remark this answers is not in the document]
cptartur: I think we already have code similiar to that inside ContractsController for declaring contracts.

tests/e2e/cairo1/test_cairo1_building.py
Arcticae: It's out of scope probably even if it's true, but i thought it was worth noting. Perhaps we can create an issue for this kind of stuff? cc @abulenok

protostar/compiler/project_compiler.py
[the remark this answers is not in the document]
Arcticae: This comment seems stale, right?

tests/e2e/cairo1/test_cairo1_building.py
MaksymilianDemitraszek: warnings are redirected to stderr, ndjson should only be used in stdout. There is no need to do so.

protostar/compiler/project_compiler.py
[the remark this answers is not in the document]
cptartur: I think it's still relevant

tests/e2e/cairo1/test_cairo1_building.py
abulenok: agree with @MaksymilianDemitraszek, for test I've added `ignore_stderr=True`

tests/integration/cairo1/compiler_bindings/starknet_contracts_compiler_test.py
abulenok: Before recent changes, cairo bindings didn't return anything, when `output_path` was provided. It has changed in recent PRs.
But I have improved these tests in the newest commits

protostar/commands/cairo1_commands/build_cairo1_command.py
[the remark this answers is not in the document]
abulenok: I've moved `make_compiled_class` and `make_contract_class` to the `contracts` module. But the `compute_compiled_class_hash_from_casm_code` and `compute_class_hash_from_sierra_code` functions are specific to `build` command, I don't see a need to extract them to a separate module for now

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -46,6 +46,18 @@
 from protostar.starknet.data_transformer import CairoData
 
 
+def make_contract_class(sierra_compiled: str):
```
Arcticae, line 49: return type should be annotated

protostar/cheatable_starknet/controllers/contracts.py
```diff
@@ -46,6 +46,18 @@
...
+    sierra_compiled_dict = json.loads(sierra_compiled)
+    sierra_compiled_dict.pop("sierra_program_debug_info", None)
+    sierra_compiled_dict["abi"] = json.dumps(sierra_compiled_dict["abi"])
+
+    return ContractClass.load(sierra_compiled_dict)
+
+
+def make_compiled_class(casm_compiled: str):
```
Arcticae, line 57: return type should be annotated

protostar/commands/cairo1_commands/build_command.py
```diff
@@ -1,31 +1,91 @@
...
+
+
+def compute_compiled_class_hash_from_casm_code(casm_compiled: str, output_path: Path):
+    compiled_class = make_compiled_class(casm_compiled)
+    compiled_class_hash = compute_compiled_class_hash(compiled_class)
+
+    with open(output_path, mode="w", encoding="utf-8") as output_file:
+        output_file.write(f"{hex(compiled_class_hash)}")
+
+    return compiled_class_hash
```
Arcticae, line 50: Those along with `make_contract_class`, `make_compiled_class` could be in a separate module called `contract_class`

Arcticae (approved):
Some nitpicking, looks very good to me, feel free to merge after 2nd approve

protostar/cairo/contract_class.py
```diff
@@ -0,0 +1,43 @@
...
+    class_hash = compute_class_hash(contract_class)
+
+    with open(output_path, mode="w", encoding="utf-8") as output_file:
+        output_file.write(f"{hex(class_hash)}")
+
+    return class_hash
+
+
+def compute_compiled_class_hash_from_casm_code(casm_compiled: str, output_path: Path):
+    compiled_class = make_compiled_class(casm_compiled)
```
cptartur, line 37: Return type annotations missing in these functions

protostar/cairo/contract_class.py
```diff
@@ -0,0 +1,43 @@
...
+def make_compiled_class(casm_compiled: str) -> CompiledClass:
+    return CompiledClass.loads(casm_compiled)
+
+
+def compute_class_hash_from_sierra_code(sierra_compiled: str, output_path: Path):
+    contract_class = make_contract_class(sierra_compiled)
+    class_hash = compute_class_hash(contract_class)
+
+    with open(output_path, mode="w", encoding="utf-8") as output_file:
+        output_file.write(f"{hex(class_hash)}")
```
cptartur, line 31: I don't really like this approach of always writing to file when computing these. Write should be done separately in the methods that require it.

protostar/compiler/cairo1_contract_compiler.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @staticmethod
+    def compile_contract_to_sierra(
+        contract_name: str,
+        contract_path: Path,
+        cairo_path: Optional[list[Path]] = None,
+        output_path: Optional[Path] = None,
+    ) -> str:
+        if output_path:
+            output_path = output_path.with_suffix(".sierra.json")
```
cptartur, line 52: I'm not sure if we should be modifying the path provided.

protostar/compiler/cairo1_contract_compiler.py
```diff
@@ -0,0 +1,110 @@
...
+                    output_path=output_path,
+                )
+            )
+
+        except cairo1_bindings.CairoBindingException as ex:
+            raise SierraCompilationException(
+                contract_name=contract_name, err=ex
+            ) from ex
+
+        assert sierra_compiled is not None
```
cptartur, line 68: I think the assert here is no longer neede here. The bindings are now typed to always return a value.

tests/e2e/cairo1/test_cairo1_building.py
```diff
@@ -9,14 +10,37 @@
...
+
     compiled_sierra_path = Path("build/main.sierra.json")
     compiled_casm_path = Path("build/main.casm.json")
     assert compiled_sierra_path.exists()
     assert compiled_sierra_path.read_text()
     assert compiled_casm_path.exists()
     assert compiled_casm_path.read_text()
 
+    class_hash_path = Path("build/main.class.hash")
+    compiled_class_hash_path = Path("build/main.compiled.class.hash")
```
cptartur, line 27: I'm not sure where exactly this change was introduced, but I don't think we should be using `compiled.class.hash`.

If we want to have a custom extension for these I'd change it to `.compiled_class_hash` and `.compiled_class` respectively.

protostar/compiler/cairo1_contract_compiler.py
abulenok: I removed this and added `output_path` creation to `compile_contract`

protostar/cairo/contract_class.py
abulenok: I think we can leave writing to `output_path` here, but make it optional. This way, it will behave similarly to functions from `cairo_bindings`.

cptartur: That's better but I still would prefer if writing to file was moved entirely to the module needing it. 

I'm assuming writing output of this command to file is needed only in the single place of the library.

cptartur (approved):

PR #1804, opened 2023-04-21
Cleanup tests

### LINKED ISSUE #1696: Cairo1 tests directory structure
We need to provide a solid directory structure for cairo1 runer tests

Currently those are scatterd between
`tests/integration/cairo1`
`tests/integration/cairo1_compiler_bindings`
`tests/integration/pure_cairo_vm`

cptartur:
Closes #1696


karol-bisztyga (commented):
How about doing `tests/integration/cairo1` instead of `tests/integration/cairo1_tests`, it's obvious there are going to be tests there, right?

Also, why `commands` dir is outside of the main Cairo 1 directory (next to it to be precise)? I'd put it inside of it.

cptartur:
> How about doing `tests/integration/cairo1` instead of `tests/integration/cairo1_tests`, it's obvious there are going to be tests there, right?

Eventually I want to name it just `cairo1`, but the directory like this already exists so resolving possible conflicts between this PR and master is not going to be nice. I'll rename it before the merge.

> Also, why `commands` dir is outside of the main Cairo 1 directory (next to it to be precise)? I'd put it inside of it.

There are commands which aren't directly related to cairo1 and I don't think it makes much sense to split it between cairo0 and 1 commands directories, since some are shared between versions.

karol-bisztyga (approved):
Ok, cool. I'm going to accept, the idea is obviously right, and the implementation should be alright with what's been said here. I believe that it's about to be satisfying.

Arcticae (dismissed):

MaksymilianDemitraszek:
Can we prioritize merging this one above anything elese? It will be very prone to merge conflicts

Arcticae (dismissed):

PR #1810, opened 2023-04-21
Add cairo1 prank cheatcode

### LINKED ISSUE #1748: Add `prank` cheatcode
It also should include stop variant 

```
let x = prank(target_address, value)
stop_prank(x)
```

cptartur:
Closes #1748

Docs https://github.com/software-mansion/protostar/pull/1811


cptartur:
Docs here https://github.com/software-mansion/protostar/pull/1811

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
```diff
@@ -0,0 +1,88 @@
...
+    assert(*return_data.at(0_u32) == 2, 'check call result');
+}
+
+#[test]
+fn test_stop_prank_multiple_times() {
+    let deployed_contract_address = deploy_contract('pranked', ArrayTrait::new()).unwrap();
+    start_prank(123, deployed_contract_address).unwrap();
+
+    stop_prank(deployed_contract_address).unwrap();
+    stop_prank(deployed_contract_address).unwrap();
```
Arcticae, line 50: shouldn't that fail?

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
```diff
@@ -0,0 +1,88 @@
...
+    assert(*return_data.at(0_u32) == 2, 'check call result');
+}
+
+#[test]
+fn test_stop_prank_on_non_existent() {
+    stop_prank(1234).unwrap();
+}
+
+#[test]
+fn test_stop_prank_on_not_pranked() {
```
Arcticae, line 34: Those two should fail i believe (be broken)

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
```diff
@@ -0,0 +1,88 @@
...
+    let return_data = call(deployed_contract_address, 'get_a', ArrayTrait::new()).unwrap();
+    assert(*return_data.at(0_u32) == 3, 'check call result');
+}
+
+#[test]
+fn test_stop_prank_cancels_all_pranks() {
+    let deployed_contract_address = deploy_contract('pranked', ArrayTrait::new()).unwrap();
+
+    start_prank(123, deployed_contract_address).unwrap();
+    start_prank(123, deployed_contract_address).unwrap();
```
Arcticae, line 77: Pranking twice shouldn't really be different to pranking once, we store only a key in a map, so this test doesn't make much sense

Arcticae (changes requested):

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
cptartur: I think it useful to have test like that. There should be nothing preventing the user from pranking the same address multiple times.

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
cptartur: Previously `start_prank` returned a callable that could cancel a prank. I'm assuming it was impossible for it to fail then. 

Having the ability to call `stop_prank` multiple times/on non-existing is just the behavior I chose.

Arcticae: If others agree about this, i'm fine with it as well, don't have a strong opinion on this

protostar/cheatable_starknet/callable_hint_locals/start_prank_hint_local.py
```diff
@@ -13,13 +13,13 @@ def __init__(self, contracts_controller: ContractsController):
...
     def name(self) -> str:
-        return "prank"
+        return "start_prank"
 
     def _build(self) -> Callable:
         return self.prank
 
-    def prank(self, caller_address: int, target_address: int):
-        valid_target_address = Address.from_user_input(target_address)
+    def prank(self, caller_address: int, target_contract_address: int):
```
MaksymilianDemitraszek, line 21: so `prank` or `start_prank`, please commit and use one consistently 🥺

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
```diff
@@ -0,0 +1,88 @@
+use array::ArrayTrait;
```
MaksymilianDemitraszek, line 1: ❤️ Very solid suite

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
MaksymilianDemitraszek: In my opinion 
- pranking non existent contract makes sense, you might want to prank constructor on deploy
- stop_prank - when it cleans up everything i would also let it run on unpranked contracts without error. I interpret the semantics as - clean everything up if there is something to clean up

MaksymilianDemitraszek (dismissed):
Please do not split the docs and code in future PRs

karol-bisztyga (dismissed):
this is alright 👍

Arcticae (dismissed):

protostar/cheatable_starknet/callable_hint_locals/start_prank_hint_local.py
Arcticae: `start_prank` vs `prank` semantics are from foundry cheatcodes, if we use the names we should stick to what they do in foundry as well.

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
MaksymilianDemitraszek: Good point @Arcticae, but I don't see a reason why to not add it here

protostar/cheatable_starknet/callable_hint_locals/start_prank_hint_local.py
cptartur: It is leftover from the previous code. I will change it to `start_prank` here.

tests/integration/cairo1_hint_locals/prank/prank_test.cairo
cptartur: We agreed to allow stoping prank on non-pranked/non-existent addreses.

cptartur:
I merged the docs into this PR for easier review.

MaksymilianDemitraszek (approved):

Arcticae (approved):

Arcticae (approved):

karol-bisztyga (approved):

PR #1842, opened 2023-04-25
added compiler version info to `--version`

MaksymilianDemitraszek:


protostar/self/protostar_directory.py
```diff
@@ -109,6 +117,20 @@ def git_version(self) -> Optional[Version]:
...
+        try:
+            compiler_cargo = (
+                self._protostar_directory.protostar_cairo1_compiler_path / "Cargo.toml"
+            )
+            with open(compiler_cargo, "r") as file:
+                cargo = parse(file.read())
+                version_str: str = cargo["workspace"]["package"]["version"]  # type: ignore
+                return version.parse(version_str)
+        except BaseException:
+            return None
```
Arcticae, line 131: I feel like we should print some warning here, that no version was found

Arcticae (approved):

protostar/self/protostar_directory.py
Arcticae: Don't have a strong opinion on this, although i'd prefer to know when i have incomplete configuration of protostar instead of missing information. Resolving.

karol-bisztyga (approved):

PR #1864, opened 2023-05-05
Reduce functionality of `ProjectCompiler`

### LINKED ISSUE #1662: Reduce the functionality of cairo1 ProjectCompiler
As part of #1625, separate project compiler for cairo1 contracts will be introduced. This module should be simplified to just resolving of the contract paths from names. 

This resolving logic should be used in places where it is currently implemented manually like cairo1 `build` command.

Compilation should be handled in separate module/function.

cptartur:
Closes #1662 

### Checklist
- [ ] Added relevant tests


protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -84,11 +92,37 @@ def _make_contract_class(contract_identifier: ContractIdentifier):
...
+        try:
+            casm_compiled = compile_starknet_contract_to_casm_from_path(contract_path)
+
+            if casm_compiled is None:
+                raise TypeError
+        except (CairoBindingException, TypeError) as ex:
+            raise CheatcodeException(
+                self, f"Compilation of contract {contract_name} to casm failed"
+            ) from ex
+        return casm_compiled
```
cptartur, line 128: Kind of unsure about this: We have to both check if calling bindings doesn't raise an exception and check if value returned is not None, and this seemed like the most concise way of doing that. But the code is quite ugly.

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -29,34 +33,25 @@ class DeclareHintLocal(CallableHintLocal):
...
-                raise CheatcodeException(
-                    self, f"No contract found for the name {ex.contract_name}"
-                ) from ex
+            compiled_class, contract_class = _compile_contract(
+                contract_name=contract_name,
+            )
 
             declared_class: DeclaredSierraClass = asyncio.run(
-                self._contracts_controller.declare_sierra_contract(
+                self._contract_path_resolver.declare_sierra_contract(
```
Arcticae, line 54: this doesn't seem right (why is `declare_sierra_contract` called on this object?)

protostar/composition_root.py
```diff
@@ -107,7 +107,7 @@ def build_di_container(
         configuration_file=configuration_file,
     )
 
-    project_compiler = ProjectCompiler(
+    contract_path_resolver = ContractPathResolver(
```
Arcticae, line 110: I don't see a reason to pass this object around instead of a simple map of values, it would make migrating this to rust easier down the line.

tests/integration/commands/declare_cairo1_command_test.py
```diff
@@ -21,47 +20,31 @@ def compiled_contract_path_fixture() -> Path:
...
-    return MockedProjectCompiler()
+    return MockedContractPathResolver()
 
 
 async def test_declaring_cairo1_contract(
     devnet_gateway_url: str,
     devnet_account: DevnetAccount,
     set_private_key_env_var: SetPrivateKeyEnvVarFixture,
-    mocked_project_compiler: ProjectCompiler,
+    mocked_project_compiler: ContractPathResolver,
```
Arcticae, line 39: This says compiler but you assign path resolver

protostar/commands/cairo1_commands/declare_cairo1_command.py
```diff
@@ -96,16 +101,21 @@ async def run(self, args: Namespace) -> Any:
...
-                contract_name
-            )
+        contract_path = self._contract_path_resolver.contract_path_from_contract_name(
+            contract_name
         )
-        contract_casm = (
-            self._project_compiler.compile_contract_to_casm_from_contract_name(
-                contract_name
+
+        contract_sierra = compile_starknet_contract_to_sierra_from_path(contract_path)
```
Arcticae, line 108: prepending i.e. `cairo_1_bindings` would nicely show where this method comes from (clearer boundaries)

Arcticae: This goes for all of the invocations in this PR

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
Arcticae: I think we always want to raise an error, not return `None` (correct me if i'm wrong). Perhaps we should modify how the binding works?

Arcticae (changes requested):

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
cptartur: Related cairo compiler PR https://github.com/software-mansion-labs/cairo/pull/71

protostar/composition_root.py
cptartur: The problem is that what this module does is actually load contracts from ConfigurationFile + some extra error handling. We could replace the `ConfigurationFile` abstraction with a simple map, but I'm not sure if that would be a simple change, especially since we depend on updating this "file" in tests.

protostar/cairo/cairo_bindings.py
```diff
@@ -33,7 +33,7 @@ def compile_starknet_contract_to_casm_from_path(
     input_path: Path,
     output_path: Optional[Path] = None,
     cairo_path: Optional[list[Path]] = None,
-) -> Optional[str]:
+) -> str:
```
karol-bisztyga, line 36: These binding functions still return optionals, this looks like an error to me - what if `cairo_python_bindings.compile_starknet_contract_to_casm_from_path` returns `None` - OK I checked and these functions never do but we should merge a PR with correct types in the compiler.

I think this should be addressed separately because there will be inconsistencies in the bindings. Could you please revert this change here and we can create an issue for this wdyt?

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
```diff
@@ -29,31 +29,22 @@ class DeclareHintLocal(CallableHintLocal):
...
     def _build(self) -> Callable:
         def declare(contract: int) -> DeclaredContract:
-            contract_identifier = short_string_to_str(contract)
+            contract_name = short_string_to_str(contract)
 
-            try:
-                compiled_class, contract_class = _compile_contract(
-                    contract_identifier=contract_identifier,
-                )
-            except CompilationException as ex:
```
karol-bisztyga, line 49: Are you positive this is redundant? Why is handling for this error gone (or is it not and I don't see it)?

protostar/cairo/cairo_bindings.py
cptartur: https://github.com/software-mansion-labs/cairo/pull/71

cptartur: This PR should resolve this

protostar/cheatable_starknet/callable_hint_locals/declare_hint_local.py
cptartur: Specific handling has been moved to methods below. 

Compilation to casm and to sierra should now have distinct message and resolving contract path as well.

karol-bisztyga: Ok I looked into it and it seems that `CompilationException` is not used anymore in this context, resolving.

protostar/composition_root.py
Arcticae: If it's worth refactoring at all, we should consider changing tests as well then. Repacking it into object with another name is not really a huge improvement to the structure if i'm honest, but packing things into a map would actually simplify reasoning about the code a lot. The error handling can (and should be) part of validation/parsing of the configuration file.

karol-bisztyga (approved):

karol-bisztyga:
It looks ok to me but let's wait for @Arcticae

protostar/composition_root.py
cptartur: We've agreed to merge this as is and continue with refactoring when resolving #1882

Arcticae (approved):

PR #1875, opened 2023-05-09
Fix gas support

### LINKED ISSUE #1767: Investigate differences in gas support between `quaireaux` and `protostar`
They seem to be using [this](https://github.com/keep-starknet-strange/quaireaux/blob/main/quaireaux/utils/src/utils.cairo):
```cairo
#[inline(always)]
fn check_gas() {
    match gas::withdraw_gas_all(get_builtin_costs()) {
        Option::Some(_) => {},
        Option::None(_) => {
            let mut data = ArrayTrait::new();
            data.append('Out of gas');
            panic(data);
        }
    }
}
```

Meanwhile in our tests we use [this](https://github.com/software-mansion/protostar/blob/master/tests/integration/cairo1/gas_test/test.cairo#L5):
```cairo
    match gas::withdraw_gas() {
        Option::Some(_) => {},
        Option::None(_) => {
            let mut data = ArrayTrait::new();
            data.append('Out of gas');
            panic(data);
        }
    }
```

`withdraw_gas_all(get_builtin_costs())` causes [this]() e2e test using quaireaux as dependency fail.

cptartur:
Closes #1767 

### Checklist
- [x] Docs updated/issue for docs created
- [x] Added relevant tests


tests/integration/cairo1/misc/gas_test/test.cairo
```diff
@@ -2,7 +2,7 @@ use array::ArrayTrait;
 
 
 fn fib(a: felt252, b: felt252, n: felt252) -> felt252 {
-    match gas::withdraw_gas() {
+    match gas::withdraw_gas_all(get_builtin_costs()) {
```
karol-bisztyga, line 5: could you please explain this change a bit?

karol-bisztyga (commented):
bless

tests/integration/cairo1/misc/gas_test/test.cairo
cptartur: This shouldn't change anything in the behavior (I think/hope), but as per #1767 description, the syntax I introduced didn't work previously.

Perhaps we can have separate tests for both `withdraw_gas` and `withdraw_gas_all`?

karol-bisztyga: > Perhaps we can have separate tests for both `withdraw_gas` and `withdraw_gas_all`? 

That wouldn't be a bad idea.

karol-bisztyga (approved):
I think it is ok, please address that one open comment from me, thank you.

tests/integration/cairo1/misc/gas_test/test.cairo
MaksymilianDemitraszek: > Perhaps we can have separate tests for both withdraw_gas and withdraw_gas_all?
I would include them here

protostar/cairo/cairo_function_runner_facade.py
```diff
@@ -53,8 +53,9 @@ def run_from_offset(
             )
 
     # pylint: disable=unused-argument
-    @staticmethod
+    # @staticmethod
```
MaksymilianDemitraszek, line 56: .

MaksymilianDemitraszek (approved):

tests/integration/cairo1/misc/gas_test/test.cairo
cptartur: Updated tests

piotmag769: My description was probably a bit unclear, but I didn't mean that using `withdraw_gas` in the tests made using quaireaux fail. It was rather an observation - `withdraw_gas` worked with protostar, but quaireaux used `withdraw_gas_all(get_builtin_costs())` which made protostar fail.

MaksymilianDemitraszek:
@cptartur Can we fix and merge this one?

PR #1881, opened 2023-05-10
Rust test runner

### LINKED ISSUE #1844: Test runner in Rust MVP

karol-bisztyga:
Related cairo PR: https://github.com/software-mansion-labs/cairo/pull/74


protostar-rust/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+    RunResultValue::Panic(data) => {
+        
+      return format!("Panic {:?}", data);
+    }
+  }
+}
+
+fn internal_run_tests(input_path: &str) -> anyhow::Result<Vec<(String, String)>> {
+  let (sierra_program, test_configs) = collect_tests(&input_path.to_owned(), None, None, None)?;
```
cptartur, line 24: Not sure if `test_configs` is a best name for this. Wouldn't `test_case` or something like that be better?

protostar/cairo/protostar_rust_bindings.py
```diff
@@ -0,0 +1,33 @@
...
+from contextlib import contextmanager
+
+import rust_test_runner_bindings
+
+
+class CairoBindingException(Exception):
+    def __init__(self, message: str, details: Optional[str] = None):
+        self.message = message
+        self.details = details
+        super().__init__(message)
```
cptartur, line 11: Didn't we already have a class similar to this somewhere?

scripts/install_protostar_rust.sh
cptartur, line 1: Was this just copied over from installing compiler bindings?

protostar/cairo/protostar_rust_bindings.py
MaksymilianDemitraszek: We have the same class in compiler bindings

protostar-rust/src/lib.rs
MaksymilianDemitraszek: It is the name used by cairo test runner

protostar-rust/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+  for config in &test_configs {
+    let result = runner
+    .run_function(
+        runner.find_function(&config.name.as_str())?,
+        &[],
+        config.available_gas,
+        Default::default(),
+    )
+    .with_context(|| format!("Failed to run the function `{}`.", config.name.as_str()))?;
+    results.push((config.name.clone(), run_result_value_to_string(result.value)));
```
MaksymilianDemitraszek, line 43: Can we print results here instead of returning them?

protostar-rust/src/lib.rs
karol-bisztyga: The type is `Vec<TestConfig>` so I find this name pretty intuitive. I'm resolving this, if you still have a strong opinion here, please unresolve and state your arguments.

scripts/install_protostar_rust.sh
karol-bisztyga: yes

protostar/cairo/protostar_rust_bindings.py
karol-bisztyga: correct, should be reused

scripts/install_protostar_rust.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
karol-bisztyga: I think we should expose both, separate installation and installation of all bindings at once.

We should be able to install them separately if we decide to focus on debugging one thing, then it is good to be able to install single bindings quickly.

We should be also able to install them all together because it is simple - this is for devs who just want to install them and move forward.

protostar-rust/src/lib.rs
karol-bisztyga: Hmm, yes, right, once we start supporting parallel tests, we should print the results in Rust anyway so we can start doing it right away. Thanks for pointing this out.

scripts/install_protostar_rust.sh
karol-bisztyga: [make the structure of files more clear, reuse the exception code](https://github.com/software-mansion/protostar/pull/1881/commits/7f81331f897517135e4deefb9cbfdaa2c711f80d)

protostar/cairo/protostar_rust_bindings.py
karol-bisztyga: [make the structure of files more clear, reuse the exception code](https://github.com/software-mansion/protostar/pull/1881/commits/7f81331f897517135e4deefb9cbfdaa2c711f80d)

scripts/install_protostar_rust.sh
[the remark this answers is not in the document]
karol-bisztyga: [allow to install all bindings at once](https://github.com/software-mansion/protostar/pull/1881/commits/781d6b6b7f798353fcab2406ac3dc36cd168fe83)

protostar-rust/src/lib.rs
karol-bisztyga: [add printing](https://github.com/software-mansion/protostar/pull/1881/commits/bfad1b5d5dfa999cd6b78cfceb8d69f9bcdf2907)

MaksymilianDemitraszek: > Hmm, how about printing and returning them? The point in returning is that we can easily test the outputs in protostar.

I think we shouldn't have such tests on the protostar side.

scripts/install_protostar_rust.sh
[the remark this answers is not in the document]
MaksymilianDemitraszek: Can you update README about the bindings?

protostar-rust/src/lib.rs
karol-bisztyga: discussed offline
[do not return value from the binding](https://github.com/software-mansion/protostar/pull/1881/commits/99554b3018b98a4d792ac7c82575b14dcacf9467)

scripts/install_protostar_rust_bindings.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+set -e
+
+# clean up
+if [ "$1" == "--cleanup" ]; then
+  poetry env info -p | xargs rm -rf
+  if [[ $(uname -m) == 'arm64' ]]; then
+    CFLAGS=-I/opt/homebrew/opt/gmp/include LDFLAGS=-L/opt/homebrew/opt/gmp/lib poetry install
+  else
```
THenry14, line 10: I think this should rather check for "Darwin", as this is mac specific (and not arm specific)
```suggestion
  if [[ $(uname -s) == 'Darwin' ]]; then
    CFLAGS=-I/opt/homebrew/opt/gmp/include LDFLAGS=-L/opt/homebrew/opt/gmp/lib poetry install
  else
```

karol-bisztyga: I remember we talked about this when I originally came up with it and we decided to leave it as is for the time being. I was thinking we could address this separately but you made me think about this and this is actually a bug. The problem here is when we do `poetry run poe install_all_bindings --cleanup` it will execute both scripts with the `--cleanup` flag so it will discard the first installation.

As for your comment itself, I think you're wrong, this is actually arm-specific as it is a way to go for the m1/m2 Mac (you can see it in the protostar readme).

scripts/install_protostar_rust.sh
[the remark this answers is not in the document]
karol-bisztyga: they were updated in https://github.com/software-mansion/protostar/pull/1918/commits/c5a81ff78521f96b9d495ee15305bbad77a37de3#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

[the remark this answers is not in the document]
karol-bisztyga: or did I miss anything?

scripts/install_protostar_rust_bindings.sh
THenry14: > I remember we talked about this when I originally came up with it and we decided to leave it as is for the time being. I was thinking we could address this separately but you made me think about this and this is actually a bug. The problem here is when we do poetry run poe install_all_bindings --cleanup it will execute both scripts with the --cleanup flag so it will discard the first installation.

I suppose, but there is no such invokation on CI, right? This is only relevant to local dev env, and I suppose dev should know about it and if one wants to, be able to do the cleanup for either version; thus I don't see it as a bug, but as a feature ( ͡° ͜ʖ ͡°). 

> As for your comment itself, I think you're wrong, this is actually arm-specific as it is a way to go for the m1/m2 Mac (you can see it in the protostar readme).

- arm64 is not mac specific, and you **normally** won't have homebrew installation on linux 
- if I remember correctly I had this issue once on mac x86
- protostar readme is not the best source for this kind of information, as it is written by us
- cairo-lang docs mention macs in general, not m1/m2 (https://www.cairo-lang.org/docs/quickstart.html)

karol-bisztyga: > This is only relevant to local dev env, and I suppose dev should know about it

I think we should prevent such situations as much as we can ;)

As for the second part, I think you're partially right but not entirely ;) In fact, this is both arm and Mac specific, please note that the condition determines if we want to use the `CFLAGS`, the rest is the same (arm/m1 specific), But you're right that we use brew-related stuff (max specific).

Anyway, I removed the `--cleanup` option in [remove cleanup](https://github.com/software-mansion/protostar/pull/1881/commits/0a921e548daf44c60afbf4d381ddb14783033686). The reason is that we've been skeptical about this since the very beginning and also I don't really remember using it (probably no one does) so it may bring more confusion and problems than gain.

THenry14: > As for the second part, I think you're partially right but not entirely ;)

ok fair enough, if you're sure it is that way let's add checks for darwin && arm64 :+1:

>Anyway, I removed the --cleanup option in [remove cleanup](https://github.com/software-mansion/protostar/pull/1881/commits/0a921e548daf44c60afbf4d381ddb14783033686).

I'm ok with that :+1:

THenry14 (approved):
lgtm!

MaksymilianDemitraszek (approved):

Issue #1806, opened 2023-04-21
`start_spoof` cheatcode


Arcticae:
We should be able to provide message signature from `get_tx_info` syscall in tests, in order to enable testing accounts

cptartur:
Work on compiler side should already be mostly done (see linked PR). We've decided to continue working on python implementation later. Moving back to backlog

abulenok:
`start_spoof` allows mocking TransactionInfo returned by `get_tx_info` (https://github.com/starkware-libs/cairo/blob/main/corelib/src/starknet/info.cairo#L61)
`stop_spoof` removes this mocking

PR #1891, opened 2023-05-10
Add design document for mocking tx info

cptartur:
The purpose of this PR is to propose a solution of #1806 before I start working on the implementation. 

### Checklist
- [ ] Added relevant tests


Arcticae (commented):
That solution is fine with me 👍

design_docs/mock-tx-info.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+and corresponding `stop_mock_tx_info` cheatcode, that will allow mocking all transactions and calls to the contract
+given by `contract_address`.
+
+`NewTxInfo` accepts `Option`s instead of plain values to allow mocking only parts of `get_tx_info()` response. For
+fields with `None` provided, default values will be used instead of mocking.
+
+Consecutive calls to `start_mock_tx_info` will modify the values of fields where `Some` was provided and leave the
+previous values for fields where `None` was provided.
```
cptartur, line 62: A possible problem with this approach: There's no way of cancelling mocks for specific fields. It's only possible to cancel entire `TxInfo`. 

But resolving this would conflict with the ability of "re-mocking" only some field: We cannot support both at the same time.

Arcticae: I think lot of those would be resolved by default function argument mechanism in cairo, i think it's planned for the future (don't fact-check me though, im not sure)

design_docs/mock-tx-info.md
```diff
@@ -0,0 +1,112 @@
...
+    transaction_hash: Option<felt252>,
+    chain_id: Option<felt252>,
+    nonce: Option<felt252>,
+}
+```
+
+and corresponding `stop_mock_tx_info` cheatcode, that will allow mocking all transactions and calls to the contract
+given by `contract_address`.
+
+`NewTxInfo` accepts `Option`s instead of plain values to allow mocking only parts of `get_tx_info()` response. For
```
MaksymilianDemitraszek, line 58: I would would create one with default values and let user set only those necessary for their use-case? WDYT? I would keep optional though

design_docs/mock-tx-info.md
MaksymilianDemitraszek: I wouldn't bet on that, Rust does not support default arguments

design_docs/mock-tx-info.md
```diff
@@ -0,0 +1,112 @@
...
+## Goal
+
+Propose a solution for mocking contexts of `TxInfo`.
+
+## Proposed solution
+
+Introduce a new cheatcode `start_mock_tx_info` (working title) with a following signature
+
+```cairo
+fn start_mock_tx_info(contract_address: felt252, tx_info: NewTxInfo) -> Result::<(), felt252> nopanic
```
MaksymilianDemitraszek, line 42: In what scenario it could fail? Imo it should just return `()`

design_docs/mock-tx-info.md
cptartur: I think we can do that

design_docs/mock-tx-info.md
cptartur: I don't think there's a viable scenario for this to fail, but all our cheatcodes currently are defined to return some kind of error code.

MaksymilianDemitraszek: We could revisit that for cheatcodes that cannot fail.

design_docs/mock-tx-info.md
cptartur: I've updated the document and I think the new approach provides a good flexibility of mocking even without default arguments.

design_docs/mock-tx-info.md
```diff
@@ -0,0 +1,120 @@
...
+Propose a solution for mocking contexts of `TxInfo`.
+
+## Proposed solution
+
+Introduce a new cheatcode `start_mock_tx_info` (working title) with a following signature
+
+```cairo
+fn start_mock_tx_info(contract_address: felt252, tx_info: MockedTxInfo) -> Result::<(), felt252> nopanic
+
+struct MockedTxInfo {
```
MaksymilianDemitraszek, line 44:
```suggestion
struct TxInfoMock {
```

design_docs/mock-tx-info.md
```diff
@@ -0,0 +1,120 @@
...
+impl MockedTxInfoImpl of MockedTxInfoTrait {
+    fn default() -> MockedTxInfo {
+        MockedTxInfo {
+            version: Option::Some(1),
+            account_contract_address: Option::Some(0),
+            max_fee: Option::Some(0_u128),
+            signature: Option::Some(ArrayTrait::new()),
+            transaction_hash: Option::Some(0),
+            chain_id: Option::Some(0),
+            nonce: Option::Some(0),
```
MaksymilianDemitraszek, line 80: Why not `None`
It would be clear that field isn't mocked

design_docs/mock-tx-info.md
```diff
@@ -0,0 +1,120 @@
...
+        TxInfoMock {
+            version: Option::None,
+            account_contract_address: Option::None,
+            max_fee: Option::None,
+            signature: Option::None,
+            transaction_hash: Option::None,
+            chain_id: Option::None,
+            nonce: Option::None,
+        }
+    }
```
Arcticae, line 82: Alternatively, we could do a `current()` method and return the values that are actually there now. 
It would make for better UX overall since there's no extra logic behind providing `None`

MaksymilianDemitraszek (approved):

Arcticae (approved):

PR #1953, opened 2023-05-19
[CI] Further cost optimization

THenry14:
This one is for further cost optimizations on CI. 
It turned out mac checks ate significant chunk of our credits, so from now on we will run all the checks only on master to make sure release is stable. For PRs there will be only linux checks. + Added pre-caching for mac jobs, to further optimize credit-time for jobs that do not yet have cache created.

### Checklist
- [X] Added relevant tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -331,6 +331,12 @@ jobs:
       - notify_about_failed_test:
           message: "*End to end tests have failed* :rotating_light:"
 
+  pre_cache_mac_bindings_and_compiler:
+    executor: mac
+    steps:
+      - setup:
+          platform: mac
+
   build:
     parameters:
       platform:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -381,19 +387,16 @@ jobs:
...
     steps:
       - checkout
-      - attach_workspace:
-          at: /tmp/workspace
       - run:
           name: Verify versions
           command: |
-            tar -xzvf /tmp/workspace/protostar-Linux.tar.gz ./dist/protostar/info/constants.json
-            built_version=$(grep dist/protostar/info/constants.json -e "PROTOSTAR_VERSION\":" | awk '{print $2}' | tr -d ',')
+            build_version=$(grep pyproject.toml -e "^version =" | awk '{print $3}')
```
cptartur, line 393: This is mostly out of curiosity, but why do we have to use extended regex here?

.circleci/config.yml
```diff
@@ -530,25 +542,42 @@ workflows:
...
+            branches:
+              ignore:
+                - master
       - verify_valid_version:
           name: verify_valid_version_branch
           requires:
             - code_quality
             - unit_tests-linux
             - e2e_tests-linux
-            - build
```
cptartur, line 551: Why was build removed from here?

.circleci/config.yml
cptartur, line 1: It's just I don't understand circleci that much, but what part of this PR disables macos checks for PRs? I see that `unit_tests-mac` are set to only target master branch, but I'm not sure if that's it.

.circleci/config.yml
THenry14: extended is `-E`, `-e` is just regular pattern :)

.circleci/config.yml
THenry14: The dependencies between jobs changed - now it looks like this:
```
verify_valid_version -> approval -> build -> release
```

whereas it looked like this:
```
verify_valid_version -> build -> approval -> release
```

Building is being tested with e2e tests, so there were no real reason to test building (with packaging) explicitly for each PR, that is why it will from now on only build and package after approval

.circleci/config.yml
THenry14: Sure, this is exactly it - you can filter out test executions for various things, and I just filter all *-mac tests to run only on master (so exactly what you've pointed out).

This way we do not run macos tests per each commit in PRs (and I've seen people did it quite often, resulting in waste in credits; previous not finished checks were cancelled by each new commit push, but still checks had to start running and very often were halfway there), still ensure tests actually pass for PRs (if they work for linux, they should also work for mac), and are double sure everything is right on master (so for every main release), by running everything that is possible there.

Hope that clarifies it :+1:

cptartur: Yes, thank you!

.circleci/config.yml
cptartur: Okay, it does seem more cost efficient that way.

cptartur (approved):
Seems fine to me

PR #1974, opened 2023-05-23
MVP of CLI for rust test runner

### LINKED ISSUE #1938: Scarb integration for rust test runner
Rust test runner should support scarb in the same way that python one will support it

### LINKED ISSUE #1945: CLI for rust test runner MVP
Create a simple CLI for rust test runner.

Use scarb metadata to fetch protostar arguments from `Scarb.toml.` Initially, we'll ignore fetching dependencies from scarb.

### LINKED ISSUE #2035: Remove protostar-rust bindings from python package and ci

cptartur:
Closes #1945 

Related cairo PR https://github.com/software-mansion-labs/cairo/pull/79

### Checklist
- [ ] Added relevant tests


protostar-rust/pkg/Scarb.toml
MaksymilianDemitraszek, line 1: What this directory is for? Isn't this pushed by mistake?

protostar-rust/src/lib.rs
```diff
@@ -41,16 +69,16 @@ fn internal_run_tests(input_path: &str) -> anyhow::Result<()> {
...
+//         .map_err(|e| PyErr::new::<RuntimeError, _>(format!("{:?}", e)))?;
+//
+//     Ok(())
+// }
+//
+// #[pymodule]
+// fn rust_test_runner_bindings(_py: Python, m: &PyModule) -> PyResult<()> {
+//     m.add_wrapped(wrap_pyfunction!(run_tests))?;
+//     Ok(())
+// }
```
MaksymilianDemitraszek, line 84: I think it is used in other PR

protostar-rust/src/lib.rs
```diff
@@ -1,36 +1,64 @@
...
-
-use cairo_lang_protostar::test_collector::collect_tests;
+use cairo_lang_protostar::test_collector::{collect_tests, LinkedLibrary};
 use cairo_lang_runner::{RunResultValue, SierraCasmRunner};
+use cairo_lang_sierra_to_casm::metadata::MetadataComputationConfig;
 
 fn run_result_value_to_string(run_result: RunResultValue) -> String {
-    match run_result {
+    return match run_result {
         RunResultValue::Success(data) => {
```
MaksymilianDemitraszek, line 12: [[Nitpick]]
```suggestion
        RunResultValue::Success(data) => format!("Success {:?}", data)
```

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+use serde::Deserialize;
+use std::env::set_var;
+use std::path::PathBuf;
+
+const CONTRACT_TARGET: &str = "starknet-contract";
+
+#[derive(Parser, Debug)]
+struct Args {
+    // #[clap(required = true)]
+    // paths: Vec<PathBuf>,
```
MaksymilianDemitraszek, line 15:
```suggestion
```

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[arg(short, long)]
+    linked_libraries: Option<Vec<PathBuf>>,
```
MaksymilianDemitraszek, line 17: I assume this will be read from scarb.toml eventually? Will it be changed in this task #1938 ?

protostar-rust/src/main.rs
MaksymilianDemitraszek: Don't we need to provide it?

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    Ok((base_path, dependencies))
+}
+
+fn main() -> Result<()> {
+    let args = Args::parse();
+    // TODO resolve this path somehow
+    set_var(
+        "CARGO_MANIFEST_DIR",
+        "/Users/arturmichalek/Coding/protostar/cairo/Cargo.toml",
```
MaksymilianDemitraszek, line 88: arturmichalek

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+#[derive(Parser, Debug)]
+struct Args {
+    // #[clap(required = true)]
+    // paths: Vec<PathBuf>,
+    #[arg(short, long)]
+    linked_libraries: Option<Vec<PathBuf>>,
+}
+
+#[derive(Deserialize, Debug)]
+struct ProtostarConfig {
```
MaksymilianDemitraszek, line 21:
```suggestion
struct ProtostarTestConfig {
```

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+        .inherit_stderr()
+        .exec()?;
+
+    for package in &scarb_metadata.workspace.members {
+        let protostar_config = protostar_config_for_package(&scarb_metadata, package)?;
+        dbg!(&protostar_config);
+
+        let (base_path, dependencies) = dependencies_for_package(&scarb_metadata, package)?;
+        dbg!(&dependencies);
+        dbg!(&base_path);
```
MaksymilianDemitraszek, line 101: dbg

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+    linked_libraries: Option<Vec<PathBuf>>,
+}
+
+#[derive(Deserialize, Debug)]
+struct ProtostarConfig {
+    exit_first: Option<bool>,
+    ignore: Option<Vec<String>>,
+    json: Option<bool>,
+    last_failed: Option<bool>,
+    report_slowest_tests: Option<bool>,
```
MaksymilianDemitraszek, line 26: I don't think there is a need to support any of those right now, you can add support for `exit-first` only as a PoC of integration

protostar-rust/src/lib.rs
```diff
@@ -1,36 +1,64 @@
...
+        }
+
+        let entry = entry.unwrap();
+        let path = entry.path();
+
+        if path.is_file() && path.extension().map_or(false, |ex| ex == "cairo") {
+            run_tests_in_file(entry.path().to_path_buf(), linked_libraries.clone())?;
         }
     }
+    Ok(())
```
MaksymilianDemitraszek, line 34: Can we separate files collection from tests running?

MaksymilianDemitraszek (changes requested):
Good job but it seems unfinished yet

protostar-rust/src/lib.rs
cptartur: Python bindings will get replaced with this CLI so this will stop being relevant. Having them defined alongside the code causes problems with build: maturin must be used instead of cargo, which is very much not ideal.

protostar-rust/src/main.rs
cptartur: No. The way `protostar test` will now work is similar to `cargo test`: By default it will run all tests within the workspace. 

Scarb metadata is already aware of current working directory and detects scarb workspaces within it, so I just run tests on the paths that it provides .

protostar-rust/src/main.rs
cptartur: It does that already. Take a look at `dependencies_for_package` in main.rs - the dependencies are resolved there and passed as linked libraries. I don't want to include this kind of logic "deeper" in the runner.

protostar-rust/src/main.rs
cptartur: These are not handled anywhere atm. They will be worked on later after PoC is done - there are tasks for them already.

protostar-rust/src/lib.rs
cptartur: Sure

protostar-rust/src/main.rs
cptartur: I've updated the args to remove unnecessary code. They will be expanded in later PRs.

protostar-rust/src/lib.rs
```diff
@@ -1,36 +1,71 @@
...
+        None,
+        linked_libraries,
+        Some(builtins),
+    )?;
 
-    let runner = SierraCasmRunner::new(sierra_program, None, HashMap::new())
-        .with_context(|| "Failed setting up runner.")?;
+    let runner =
+        SierraCasmRunner::new(sierra_program, Some(Default::default()), Default::default())
+            .with_context(|| "Failed setting up runner.")?;
```
mkaput, line 61:
```suggestion
            .context("Failed setting up runner.")?;
```

protostar-rust/src/lib.rs
```diff
@@ -41,16 +76,25 @@ fn internal_run_tests(input_path: &str) -> anyhow::Result<()> {
...
-fn rust_test_runner_bindings(_py: Python, m: &PyModule) -> PyResult<()> {
-    m.add_wrapped(wrap_pyfunction!(run_tests))?;
-    Ok(())
+#[derive(Deserialize, Debug)]
+pub struct ProtostarTestConfig {
+    exit_first: Option<bool>,
+    ignore: Option<Vec<String>>,
+    json: Option<bool>,
+    last_failed: Option<bool>,
+    report_slowest_tests: Option<bool>,
```
mkaput, line 99: some of these fields could make use of default values I think?

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+fn protostar_config_for_package(
+    metadata: &Metadata,
+    package: &PackageId,
+) -> Result<ProtostarTestConfig> {
+    let raw_metadata = metadata
+        .packages
+        .iter()
+        .find(|pk| pk.id == *package)
```
mkaput, line 21: https://docs.rs/scarb-metadata/latest/scarb_metadata/struct.Metadata.html#method.get_package

protostar-rust/src/main.rs
MaksymilianDemitraszek: > They will be worked on later after PoC is done
I think it should be more distant future. Keeping them here may be misleading for other developers

protostar-rust/src/lib.rs
karol-bisztyga: I think we should remove them, no point in having them commented out like this.

protostar-rust/src/main.rs
karol-bisztyga: Cannot we use a relative path?

protostar-rust/src/lib.rs
cptartur: If not provided, these should be resolved to `None` I think. And we can then handle non values as `false` or whatever other default makes sense.

protostar-rust/src/lib.rs
```diff
@@ -1,36 +1,71 @@
...
+        run_tests_in_file(test, linked_libraries.clone())?;
+    }
+    Ok(())
+}
+
+fn run_tests_in_file(
+    input_path: PathBuf,
+    linked_libraries: Option<Vec<LinkedLibrary>>,
+) -> Result<()> {
+    let builtins = vec!["GasBuiltin", "Pedersen", "RangeCheck", "bitwise", "ec_op"];
```
karol-bisztyga, line 50: wouldn't it be better to define this as static or something? Logically, this will allocate memory multiple times, if we do this as `static` we allocate just one time, right? And it can be evaluated in the compile time.

mkaput: I think compiler APIs expect a Vec here

protostar-rust/src/main.rs
cptartur: Good idea, changing to relative

protostar-rust/src/lib.rs
cptartur: We could define a static array of these but we'd have to construct a vector from these anytime we call `collect_tests` anyway. Don't know if that would be better at all.

cc @mkaput

protostar-rust/src/main.rs
cptartur: We have to defined some kind of `ProtostarTestConfig` to test if reading protostar config from scar.toml works at all so I don't think why it shouldn't already be a mostly correct struct.

protostar-rust/src/lib.rs
mkaput: The `vec!` macro basically does what you said @cptartur

protostar-rust/src/lib.rs
mkaput: That's true, but you can make your life easier by just doing:

```rust
#[serde(default)]
json: bool
```

mkaput (approved):

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+use rust_test_runner_bindings::{run_tests, ProtostarTestConfig};
+use scarb_metadata::{Metadata, MetadataCommand, PackageId};
+use std::env::set_var;
+
+#[derive(Parser, Debug)]
+struct Args {
+    test_filter: Option<String>,
+}
+
+fn protostar_config_for_package(
```
MaksymilianDemitraszek, line 14:
```suggestion
fn get_protostar_config_for_package(
```
Maybe? [[Nitpick]]

protostar-rust/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+        })
+        .collect();
+
+    Ok((base_path, dependencies))
+}
+
+fn main() -> Result<()> {
+    let args = Args::parse();
+
+    // TODO resolve this path somehow
```
MaksymilianDemitraszek, line 66: Still TODO? Can we make it a task?

protostar-rust/src/lib.rs
```diff
@@ -1,36 +1,84 @@
...
-        RunResultValue::Success(data) => {
-            return format!("Success {:?}", data);
-        }
-        RunResultValue::Panic(data) => {
-            return format!("Panic {:?}", data);
+    return match run_result {
+        RunResultValue::Success(data) => format!("Success {:?}", data),
+        RunResultValue::Panic(data) => format!("Panic {:?}", data),
+    };
+}
```
MaksymilianDemitraszek, line 25:
```suggestion
fn run_result_value_to_string(run_result: RunResultValue) -> String {
    return match run_result {
        RunResultValue::Success(data) => format!("PASS {:?}", data),
        RunResultValue::Panic(data) => format!("FAIL {:?}", data),
    };
}
```
i know it will change but i am scared somebody will reuse those results

MaksymilianDemitraszek (approved):
Can we rename pkg to example-package?

protostar-rust/src/main.rs
MaksymilianDemitraszek: I have seen situations in my career when such interfaces are defined and many people are certain for a long period of time that those things are already implemented. Structure definition is part of dev docs, we should make it represent how things are.

MaksymilianDemitraszek: I am removing those for now, if you have strong opinion about this (I have :p) we can discuss it later

THenry14 (approved):

PR #2012, opened 2023-05-29
Disable output buffering for tests + build docs on all PRs

THenry14:
Context: https://support.circleci.com/hc/en-us/articles/360045268074?input_string=macos+medium+runner+slowness

I have observed a couple of jobs that failed for master, and they seemed to buffer the output. However, if the issue persist I will just create a ticket to circleci (because macos jobs shouldn't take much longer to run in comparison to linux jobs on similar speced machines)

+ as requested, added building docs check for all PRs

### Checklist
- [X] Added relevant tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -259,6 +259,7 @@ commands:
       - run:
           name: Run integration tests
           command: |
+            export PYTHONUNBUFFERED=1
             poetry run pytest -n4 --random-order scripts tests/integration
 
   run_e2e_tests:
@@ -268,6 +269,7 @@ commands:
       - run:
           name: Run e2e tests
           command: |
+            export PYTHONUNBUFFERED=1
             poetry run pytest -n4 tests/e2e
 
   notify_about_failed_test:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -458,13 +460,6 @@ jobs:
           key: yarn-docs-{{ checksum "website/yarn.lock" }}
           paths:
             - ~/.cache/yarn
-      - run:
-          name: Pack docs
-          command: tar -czvf protostar-docs.tar.gz website/build
-      - persist_to_workspace:
-          root: .
-          paths:
-            - protostar-docs.tar.gz
```
cptartur, line 467: Why was this step removed?

THenry14: they are not needed at the moment, as we still deploy docs from github (and since docs will be built for every PR, it is just a waste of credits/storage to keep all the built versions of docs for no real reason). There is a task to deploy docs from circle, it will be re-added then.

cptartur (approved):

PR #2033, opened 2023-05-31
Test runner CLI tests

### LINKED ISSUE #2008: Add unit and integration tests
Introduce unit and integration test of the runner code.

Especially focus on possible failures in Scarb<>Protostar interaction and incomplete error handling.

### LINKED ISSUE #2009: Add e2e tests
Introduce e2e tests that utilize the whole test runner binary, similarly to how current e2e tests of protostar work now.

cptartur:
Closes #2008
Closes #2009 

Related cairo PR https://github.com/software-mansion-labs/cairo/pull/84

### Checklist
- [x] Added relevant tests


cptartur:
CI will have to be updated to run these tests, but this will be done in separate PR.

protostar-rust/src/lib.rs
```diff
@@ -82,3 +91,174 @@ fn run_tests_in_file(
...
+pub fn protostar_config_for_package(
+    metadata: &Metadata,
+    package: &PackageId,
+) -> Result<ProtostarTestConfig> {
+    let raw_metadata = metadata
+        .get_package(package)
+        .ok_or_else(|| anyhow!("Failed to find metadata for package = {package}"))?
+        .tool_metadata("protostar")
+        .ok_or_else(|| anyhow!("Failed to find protostar config for package = {package}"))?
+        .clone();
```
piotmag769, line 104: Shouldn't we just log this part? I don't feel like it is necessary with minimal setup to have any arguments regarding protostar in Scarb.toml

piotmag769 (approved):

protostar-rust/tests/e2e/main.rs
```diff
@@ -0,0 +1,2 @@
+mod running;
+pub(crate) mod common;
```
mkaput, line 2: nit: missing endline

protostar-rust/tests/e2e/common/runner.rs
```diff
@@ -0,0 +1,17 @@
...
+    snapbox
+}
+
+pub fn corelib_path() -> String {
+    let corelib = Utf8PathBuf::from("../../cairo/corelib/src")
+        .canonicalize_utf8()
+        .expect("Failed to find corelib")
+        .to_string();
+    corelib
+}
```
mkaput, line 17: Hint: you could use https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html as a static variable (could be inside this function) to permanently store the String and thus be able to return `&'static str` from this function. What would lead to much more pleasable usage of this function in tests

protostar-rust/src/lib.rs
cptartur: Fair point, but let's change this logic in the separate PR.

cptartur: #2041

protostar-rust/src/lib.rs
```diff
@@ -82,3 +91,174 @@ fn run_tests_in_file(
...
+    }
+
+    #[test]
+    fn collecting_tests_err_on_invalid_dir() {
+        let tests_path = Utf8PathBuf::from("aaee");
+
+        let result = collect_tests_in_directory(&tests_path);
+        let err = result.unwrap_err();
+
+        assert!(format!("{}", err).contains("Failed to read directory at path"));
```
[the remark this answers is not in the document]
cptartur: I'll change it to use `to_string`. `assert_eq` cannot be used here because the actual error is like `Failed to read directory at path = path/to/file` so we cannot assert equality.

protostar-rust/tests/e2e/running.rs
```diff
@@ -0,0 +1,25 @@
...
+fn running_tests() {
+    let temp = assert_fs::TempDir::new().unwrap();
+    temp.copy_from("example_package", &["**/*"]).unwrap();
+
+    let snapbox = runner();
+    let corelib = corelib_path();
+
+    snapbox
+        .current_dir(&temp)
+        .args(&["--corelib-path", corelib])
```
MaksymilianDemitraszek, line 15: We probably should test the scenario without it being specified, is that possible?

Also I would us to have module level integration tests running some cairo code (as we did for cheatcodes), though it doesn't have to be included here necessarily

protostar-rust/tests/e2e/running.rs
```diff
@@ -0,0 +1,25 @@
+use crate::common::runner::{corelib_path, runner};
+use assert_fs::fixture::PathCopy;
+use indoc::indoc;
+
+#[test]
+fn running_tests() {
+    let temp = assert_fs::TempDir::new().unwrap();
+    temp.copy_from("example_package", &["**/*"]).unwrap();
```
MaksymilianDemitraszek, line 8: I think it should be top level, imo we should store it in `tests/test_package`

protostar-rust/tests/e2e/running.rs
cptartur: We can write a test without corelib-path specified but it will simply fail: Runner still tries to detect the corelib when constructing the db, but this detection doesn't work with e2e tests.

MaksymilianDemitraszek (approved):
I would name direcotry `data` instead of `resources`,

PR #2036, opened 2023-06-05
Printing for test runner

### LINKED ISSUE #2001: Implement test pretty printing
Tests results should be pretty printed, similarly to how it is done when using `protostar test`.

piotmag769:
An issue for progress bar https://github.com/software-mansion/protostar/issues/2050
### Checklist
- [ ] Added relevant tests

### DIFF
--- a/protostar-rust/src/lib.rs
+++ b/protostar-rust/src/lib.rs
@@ -3,6 +3,7 @@ use cairo_felt::Felt252;
 use cairo_lang_protostar::test_collector::{collect_tests, LinkedLibrary};
 use cairo_lang_runner::{RunResultValue, SierraCasmRunner};
 use camino::{Utf8Path, Utf8PathBuf};
+use log::info;
 use serde::Deserialize;
 use walkdir::WalkDir;
 
@@ -94,7 +95,7 @@ fn run_tests_in_file(
             .with_context(|| format!("Failed to run the function `{}`.", config.name.as_str()))?;
         let name = config.name.clone();
         let result_str = run_result_value_to_string(result.value);
-        println!("{}: {}", name, result_str);
+        info!("{}: {}", name, result_str);
     }
     Ok(())
 }
--- a/protostar-rust/src/lib.rs
+++ b/protostar-rust/src/lib.rs
@@ -1,21 +1,24 @@
-pub mod pretty_printing;
+use std::fmt::Debug;
 
-use crate::pretty_printing::print_test_summary;
 use anyhow::{Context, Result};
 use cairo_felt::Felt252;
+use camino::{Utf8Path, Utf8PathBuf};
+use serde::Deserialize;
+use walkdir::WalkDir;
+
 use cairo_lang_protostar::casm_generator::TestConfig;
 use cairo_lang_protostar::test_collector::{collect_tests, LinkedLibrary};
 use cairo_lang_runner::{RunResultValue, SierraCasmRunner};
 use cairo_lang_sierra::program::Program;
-use camino::{Utf8Path, Utf8PathBuf};
-use serde::Deserialize;
-use std::fmt::Debug;
-use walkdir::WalkDir;
+
+use crate::pretty_printing::print_test_summary;
+
+pub mod pretty_printing;
 
 #[derive(Default)]
 pub struct TestsStats {
-    passed: u32,
-    failed: u32,
+    passed: usize,
+    failed: usize,
 }
 
 #[derive(Deserialize, Debug)]
@@ -24,13 +27,13 @@ pub struct ProtostarTestConfig {
     exit_first: bool, // TODO Not implemented!
 }
 
-fn result_data_to_text(data: Vec<Felt252>) -> String {
+fn result_data_to_text(data: &[Felt252]) -> String {
     let mut readable_text = String::new();
 
     for felt in data {
         let felt_bytes = felt.to_bytes_be();
-        let felt_text = std::str::from_utf8(&felt_bytes).unwrap_or("");
-        readable_text.push_str(felt_text);
+        let felt_text = String::from_utf8_lossy(&felt_bytes);
+        readable_text.push_str(&*felt_text);
     }
 
     readable_text
@@ -43,11 +46,11 @@ fn get_result_str_and_update_tests_stats(
     match run_result {
         RunResultValue::Success(result_data) => {
             tests_stats.passed += 1;
-            result_data_to_text(result_data)
+            result_data_to_text(&result_data)
         }
         RunResultValue::Panic(result_data) => {
             tests_stats.failed += 1;
-            result_data_to_text(result_data)
+            result_data_to_text(&result_data)
         }
     }
 }
@@ -95,10 +98,8 @@ pub fn run_test_runner(
     }
 
     pretty_printing::print_collected_tests(
-        tests_vector
-            .iter()
-            .fold(0, |acc, (_, e, _)| acc + e.len() as u32),
-        tests_vector.len() as u32,
+        tests_vector.iter().map(|(_, e, _)| e.len()).sum(),
+        tests_vector.len(),
     );
     let mut tests_stats = TestsStats::default();
     for (sierra_program, test_configs, test_file) in tests_vector {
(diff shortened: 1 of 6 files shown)


cptartur (commented):
I think that's a good start, but we can make it better. Some things that could be useful in my opinion: 

1. Color the words `PASS` and `FAIL`
2. Perhaps group the output per file, like
```shell
Running tests in test_2.cairo

test_2::test_2::test_two: PASS 
test_2::test_2::test_two_failing: FAIL 2 == 3
test_2::test_2::test_three: PASS 

Running tests in test_my_test.cairo

test_my_test::test_my_test::test_my_test: PASS 
test_my_test::test_my_test::test_four: PASS 
```
3. Perhaps show some kind of summary at the end of file and/or run? Like
```shell
Test result: PASS: 7, FAIL: 1
```
or something along the lines? 

Also right now we don't really distinguish between failed assertions and tests that are broken or not working for whatever reason. But this might be more complicated issue than just printing these correctly, so feel free to create an issue for that.

piotmag769:
image
Pretty similar to current Protostar UI, will add progression bar and it should be fine

mkaput (dismissed):

protostar-rust/src/lib.rs
```diff
@@ -1,84 +1,147 @@
...
+        )?;
+        let a = test_file.clone();
+        let x = a.strip_prefix(input_path.clone())?.to_path_buf();
+        tests_vector.push((sierra_program, test_configs, x));
+    }
+
+    pretty_printing::print_collected_tests(
+        tests_vector
+            .iter()
+            .fold(0, |acc, (_, e, _)| acc + e.len() as u32),
```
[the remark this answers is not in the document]
piotmag769: It doesn't look as smart as the previous one :c

protostar-rust/src/pretty_printing.rs
```diff
@@ -0,0 +1,49 @@
+use super::TestsStats;
+use anyhow::Error;
+use camino::Utf8Path;
+use console::set_colors_enabled;
+use console::style;
+
+pub fn enable_colors() {
+    set_colors_enabled(true);
+}
```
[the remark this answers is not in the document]
piotmag769: I'm aware of that, but this way user has to enable it manually if they want to see colors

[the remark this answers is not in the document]
mkaput: Eee, no? They should set the env var if they want to stop seeing these…

protostar-rust/src/pretty_printing.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub fn print_error_message(error: Error) {
+    println!("[{}] {}", style("ERROR").red(), error);
+}
+
+pub fn print_collected_tests(tests_num: usize, tests_files_num: usize) {
+    let plain_text = format!(
+        "Collected {} test(s) and {} test file(s)",
+        tests_num, tests_files_num
```
cptartur, line 19:
```suggestion
        "Collected {tests_num} test(s) and {tests_files_num} test file(s)",
```

cptartur: Same in other cases

protostar-rust/src/lib.rs
```diff
@@ -1,84 +1,148 @@
...
+    for test_file in test_directories {
+        let (sierra_program, test_configs) = collect_tests(
+            test_file.as_str(),
+            None,
+            linked_libraries.clone(),
+            Some(builtins.clone()),
+        )?;
+        let a = test_file.clone();
+        let x = a.strip_prefix(input_path.clone())?.to_path_buf();
+        tests_vector.push((sierra_program, test_configs, x));
```
cptartur, line 97: This could be abstracted to a function

cptartur (changes requested):
I like the outcome of this PR, but I think we could introduce some kind of better (minimal) abstraction for pretty printing. 

Right now, there's a lot of code for printing inbetween the actual running logic, which makes the code complicated to read. This should be abstracted to some methods soo we don't have to see the actual printing logic in the lib.rs.

mkaput (commented):
I can accept this, but please elaborate on colors enforcement thread

protostar-rust/src/pretty_printing.rs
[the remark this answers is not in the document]
piotmag769: Okay, true, it just doesn't work in the pycharm console, will set the env var there, thanks!

protostar-rust/src/lib.rs
```diff
@@ -1,84 +1,148 @@
...
+    let builtins = vec!["GasBuiltin", "Pedersen", "RangeCheck", "bitwise", "ec_op"];
+    for test_file in test_directories {
+        let (sierra_program, test_configs) = collect_tests(
+            test_file.as_str(),
+            None,
+            linked_libraries.clone(),
+            Some(builtins.clone()),
+        )?;
+        let a = test_file.clone();
+        let x = a.strip_prefix(input_path.clone())?.to_path_buf();
```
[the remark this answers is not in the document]
piotmag769: Forgot to change it, will do

protostar-rust/src/lib.rs
piotmag769: I disagree, it would require us to pass 2 paths around (directory path + test file path) instead of one. Change it to be more readable now, check it out

cptartur: I don't think having to pass arguments around should prevent us from defining functions for things.

protostar-rust/src/pretty_printing.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    println!("[{result_tag}] {test_name} {result_str_bold}");
+}
+
+fn get_result_str(run_result: &RunResultValue) -> String {
+    match run_result {
+        RunResultValue::Panic(result_data) | RunResultValue::Success(result_data) => {
+            result_data_to_text(result_data)
+        }
+    }
```
cptartur, line 54: Do we need match here at all?

piotmag769: How else would you extract `result_data`?

protostar-rust/src/lib.rs
```diff
@@ -1,84 +1,136 @@
...
 }
 
-pub fn run_tests(
-    input_path: Utf8PathBuf,
-    linked_libraries: Option<Vec<LinkedLibrary>>,
+fn internal_collect_tests(
+    input_path: &Utf8PathBuf,
+    linked_libraries: &Option<Vec<LinkedLibrary>>,
+    test_directories: Vec<Utf8PathBuf>,
+) -> Result<Vec<(Program, Vec<TestConfig>, Utf8PathBuf)>> {
```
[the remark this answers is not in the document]
piotmag769: I would go with alias

protostar-rust/src/lib.rs
```diff
@@ -1,84 +1,126 @@
...
     exit_first: bool, // TODO Not implemented!
 }
 
-fn run_result_value_to_string(run_result: RunResultValue) -> String {
-    return match run_result {
-        RunResultValue::Success(data) => format!("PASS {:?}", data),
-        RunResultValue::Panic(data) => format!("FAIL {:?}", data),
-    };
-}
+type TestInstance = (Program, Vec<TestConfig>, Utf8PathBuf);
```
cptartur, line 26: A type alias like this isn't really useful for other developers, I'd argue it even makes everything less readable. 

Let's create an proper struct for this.

protostar-rust/src/lib.rs
```diff
@@ -1,84 +1,126 @@
...
 
-fn collect_tests_in_directory(input_path: &Utf8PathBuf) -> Result<Vec<Utf8PathBuf>> {
-    let mut test_directories: Vec<Utf8PathBuf> = vec![];
+fn collect_tests_from_directory(
+    input_path: &Utf8PathBuf,
+    linked_libraries: &Option<Vec<LinkedLibrary>>,
```
cptartur, line 30: Kind of nit, but `Option<&T>` seems to be more flexible than `&Option<T>`. You can convert the latter to former with `as_ref()` but not really the other way around. 

https://users.rust-lang.org/t/api-design-option-t-vs-option-t/34139

I'd change it at least in public methods.

protostar-rust/src/pretty_printing.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+fn get_if_passed(run_result: &RunResultValue) -> bool {
```
cptartur, line 57: Nit
```suggestion
fn did_pass(run_result: &RunResultValue) -> bool {
```

cptartur (approved):
👍

PR #2047, opened 2023-06-06
Hotfix short string encoding

piotmag769:
### Checklist
- [ ] Added relevant tests


piotmag769:
> It seems that the short string can only include ASCII characters
You are right, it was just the first thing that came to my mind since @Arcticae got error suggesting that there was an utf-8 character here, but it is probably caused by some other issue. Closing.

Thanks for correcting!

Issue #2041, opened 2023-06-06
Make protostar config in `Scarb.toml` optional


cptartur:
There could be usecases where user only wants to run a very simple protostar test suite that doesn't require any config in `Scarb.toml`. We should investigate that and perhaps change the code of the runner so it allows not defining a config.

Issue #1997, opened 2023-05-26
Export cheatcodes as scarb package


cptartur:
Currently corelib is configured by setting a `CARGO_MANIFEST_DIR` environment variable. We should consider a better way of detecting the corelib.

Note: as of creation of this issue, corelib provided by scarb cannot be used. Protostar uses a modified corelib.

MaksymilianDemitraszek:
Related issue #1614

cptartur:
Possible solution for this: We could define a new scarb project in protostar repo containing only cheatcodes, instead of forking the corelib. Then it could be possible to use this package as a dependency in protostar projects. This way, cheatcodes would be loaded by scarb like a normal dependency would and we wouldn't have to handle detecting corelib at all.

cptartur:
cc @mkaput

Arcticae:
#2133 needs to be included in scope of this task

PR #2051, opened 2023-06-09
Allow selecting tests by name

### LINKED ISSUE #1999: Add selecting specific tests
Test should be able to be selected/filtered by providing using a first argument, similarly to how it is implemented in `cargo test` https://doc.rust-lang.org/cargo/commands/cargo-test.html.

e.g. 

- `runner test` -> runs all test
- `runner test aabb` -> run all tests with `aabb` in their name,

cptartur:
Closes #1999 

Related cairo PR https://github.com/software-mansion-labs/cairo/pull/87

### Checklist
- [ ] Added relevant tests


protostar-rust/tests/e2e/running.rs
```diff
@@ -26,6 +28,56 @@ fn running_tests() {
...
+        .arg("qwerty")
+        .args(["--corelib-path", corelib])
+        .assert()
+        .success()
+        .stdout_matches(indoc! {r#"Collected 0 test(s) and 4 test file(s)
+            Running 0 test(s) from tests/without_prefix.cairo
+            Running 0 test(s) from tests/test_2.cairo
+            Running 0 test(s) from tests/test_my_test.cairo
+            Running 0 test(s) from src/lib.cairo
+            Tests: 0 passed, 0 failed
```
piotmag769, line 81: Okay, I did not think about that when implementing pretty printing xD We should probably "if" this print if there are 0 matching elements found + print that no tests were found if that's the case (I guess we should still return 0 exit code?)

piotmag769: https://github.com/software-mansion/protostar/issues/2069

protostar-rust/src/lib.rs
```diff
@@ -317,4 +352,156 @@ version = \"0.1.0\"";
 
         assert!(err.to_string().contains("Failed to read directory at path"));
     }
+
+    #[test]
+    fn no_filter_doesnt_filter_tests() {
```
piotmag769, line 357:
```suggestion
    fn no_filtering_tests() {
```

protostar-rust/src/lib.rs
```diff
@@ -317,4 +352,156 @@ version = \"0.1.0\"";
...
+                available_gas: None,
+            },
+        ];
+
+        let filtered = filter_tests_by_name(Some("crate"), mocked_tests).unwrap();
+        assert_eq!(filtered, vec![]);
+    }
+
+    #[test]
+    fn filtering_tests_works_without_crate_in_test_name() {
```
piotmag769, line 476: Does this test check anything? It only shows that there is no error thrown. I would check if it matches against `name: "thing".to_string()`

protostar-rust/tests/e2e/running.rs
```diff
@@ -26,6 +28,56 @@ fn running_tests() {
...
+    temp.copy_from("tests/data/example_package", &["**/*"])
+        .unwrap();
+
+    let snapbox = runner();
+    let corelib = corelib_path();
+
+    snapbox
+        .current_dir(&temp)
+        .arg("two")
+        .args(["--corelib-path", corelib])
```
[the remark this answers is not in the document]
cptartur: Yes, it's because of temporary dir. It could also be a problem when running a binary, at least until  #2066 is addressed. And also our corelib detection is very limited atm, #2066 and #1997 should address this.

protostar-rust/tests/e2e/running.rs
cptartur: Let's check how cargo test handles this. I think it's actually okay to have prints like this with non matching filter, it's consistent.

protostar-rust/src/lib.rs
```diff
@@ -137,6 +150,28 @@ fn run_tests(tests: TestsFromFile, tests_stats: &mut TestsStats) -> Result<()> {
...
+) -> Result<Vec<TestConfig>> {
+    let mut result = vec![];
+    if let Some(test_name_filter) = test_name_filter {
+        for test in test_configs {
+            let name = test
+                .name
+                .rsplit("::")
+                .next()
+                .context("Failed to get test")?;
+            if name.contains(test_name_filter) {
```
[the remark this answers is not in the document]
piotmag769: I don't think it is a good design choice especially since we match only against the name, not the full path. There might be cases where it is not possible to run test A without running test B. Keep in mind that Protostar tests might need a lot of time to execute so it's a major inconvenience imo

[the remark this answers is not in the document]
cptartur: We can extend the logic so it also matches by crate etc. I'd avoid requiring to explicitly pass the whole test name, it's even more impractical in my opinion.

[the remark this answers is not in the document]
piotmag769: Discussed offline

karol-bisztyga (changes requested):
blocked by #2059

protostar-rust/src/lib.rs
```diff
@@ -317,4 +383,224 @@ version = \"0.1.0\"";
...
+                    available_gas: None,
+                },
+            ]
+        );
+
+        let filtered =
+            filter_tests_by_name(Some("nonexistent"), false, mocked_tests.clone()).unwrap();
+        assert_eq!(filtered, vec![]);
+
+        let filtered = filter_tests_by_name(Some(""), false, mocked_tests).unwrap();
```
piotmag769, line 466: I don't think this test is useful, there will never be a case when we get an empty string from the CLI (we will just get `None`).

protostar-rust/src/lib.rs
```diff
@@ -317,4 +383,224 @@ version = \"0.1.0\"";
...
+                name: "outer::crate3::run_other_thing".to_string(),
+                available_gas: None,
+            },
+            TestConfig {
+                name: "do_thing".to_string(),
+                available_gas: None,
+            },
+        ];
+
+        let filtered = filter_tests_by_name(Some(""), true, mocked_tests.clone()).unwrap();
```
piotmag769, line 528: Same as above

protostar-rust/src/lib.rs
```diff
@@ -63,6 +88,8 @@ fn internal_collect_tests(
     linked_libraries: Option<&Vec<LinkedLibrary>>,
     test_files: Vec<Utf8PathBuf>,
     corelib_path: Option<&Utf8PathBuf>,
+    test_name_filter: Option<&str>,
+    exact_match: bool,
```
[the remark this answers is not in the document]
cptartur: I was considering this, but I though that maybe passing only actually required arguments to internal methods would be better? This could discourage doing some weird hacks, unlike when whole config is available. 

But I see your point, not really sure what would be a better approach here.

protostar-rust/src/lib.rs
cptartur: Correct, but we're also exposing this code as library and having more throughout tests usually doesn't hurt. This is a valid argument this method could take.

MaksymilianDemitraszek (changes requested):
We should follow cargo test selection closely

protostar-rust/src/lib.rs
```diff
@@ -18,6 +18,22 @@ use crate::test_stats::TestsStats;
 pub mod pretty_printing;
 mod test_stats;
 
+#[derive(Deserialize, Debug, PartialEq)]
+pub struct RunnerConfig {
+    exact_match: bool,
```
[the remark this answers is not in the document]
piotmag769: It is actually a parameter from cargo test afaik https://stackoverflow.com/questions/54585804/how-to-run-a-specific-unit-test-in-rust

protostar-rust/src/main.rs
```diff
@@ -3,16 +3,21 @@ use camino::Utf8PathBuf;
...
 use scarb_metadata::MetadataCommand;
 
-use rust_test_runner::pretty_printing;
 use rust_test_runner::run_test_runner;
+use rust_test_runner::{pretty_printing, RunnerConfig};
 
 #[derive(Parser, Debug)]
 struct Args {
+    /// Name used to filter tests
     test_filter: Option<String>,
```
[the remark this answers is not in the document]
piotmag769: It is not a test name actually, it mimics the behaviour of `cargo test` https://doc.rust-lang.org/rustc/tests/index.html#filters. The difference is that we match against test name only, not a full path, because since compiler builds phantom packages from test files and those paths can get [weird](https://github.com/software-mansion/protostar/pull/2036#issuecomment-1579291040). We planned to investigate on that later on

protostar-rust/src/main.rs
```diff
@@ -6,16 +6,20 @@ use scarb_metadata::MetadataCommand;
...
 
 use std::process::Command;
 
 static CORELIB_PATH: Dir = include_dir!("../cairo/corelib/src");
 
 #[derive(Parser, Debug)]
 struct Args {
+    /// Name used to filter tests
     test_filter: Option<String>,
+    /// Use exact matches for `test_filter`
```
karol-bisztyga, line 20: can you explain what will be the difference between exact and non-exact?

protostar-rust/src/main.rs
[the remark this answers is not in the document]
karol-bisztyga: If a further investigation's planned, there should be an issue for it.

> The difference is that we look for a substring in a test name only, not a full path

Does it refer in any way to the `exact` parameter?

Anyhow, the name "test_filter"'s not good. Filter could be anything really. If we refer to a name, we should go with "test_name", if path "test_path", etc. It feels like "utils"

protostar-rust/tests/data/declare_test/Scarb.toml
```diff
@@ -8,3 +8,4 @@ sierra = true
 [dependencies]
 starknet = ">=1.1.0"
 
+[tool.protostar]
```
karol-bisztyga, line 11: this should not be required. I assume you added this to avoid errors but we should write code that allows this section to be missing.

protostar-rust/tests/data/simple_test/tests/test_simple.cairo
```diff
@@ -8,6 +8,11 @@ fn test_simple2() {
     assert(3 == 3, 'simple check');
 }
 
+#[test]
+fn test_two() {
+    assert(2 == 2, '2 == 2');
+}
+
```
karol-bisztyga, line 15: I don't see a point in this test, what does it bring?

protostar-rust/tests/data/simple_test/tests/without_prefix.cairo
```diff
@@ -0,0 +1,4 @@
+#[test]
+fn five() {
+    assert(5 == 5, '5 == 5');
+}
```
karol-bisztyga, line 4: No new line at the end of file

karol-bisztyga (changes requested):

protostar-rust/tests/data/declare_test/Scarb.toml
cptartur: There's an issue for that #2041

protostar-rust/tests/data/simple_test/tests/test_simple.cairo
cptartur: It is used in one of the e2e tests.

protostar-rust/src/main.rs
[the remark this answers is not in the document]
cptartur: >If a further investigation's planned, there should be an issue for it.

It will either be resolved within this PR or an issue will be created.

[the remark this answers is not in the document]
cptartur: Will rename it

protostar-rust/src/lib.rs
```diff
@@ -18,6 +18,22 @@ use crate::test_stats::TestsStats;
...
+    protostar_test_config: ProtostarTestConfig,
+}
```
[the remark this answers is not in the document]
cptartur: Though I'm unsure it really makes sense to have things like `exact_match` in `Scarb.toml`. I don't really see a usecase where someone would want to configure test filtering logic for running from this file,

protostar-rust/src/lib.rs
```diff
@@ -137,6 +174,35 @@ fn run_tests(tests: TestsFromFile, tests_stats: &mut TestsStats) -> Result<()> {
...
+                    .next()
+                    .context(format!("Failed to get test name from = {}", test.name))?;
+                if name.contains(test_name_filter) {
+                    result.push(test);
+                }
+            }
+        }
+    } else {
+        result = test_configs;
+    }
```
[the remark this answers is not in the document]
cptartur: Filter would make it more complicated. We're doing error handling in the `rsplit` and afaik you cannot easily bubble errors from closures. 

It can be done but would result it more complicated code.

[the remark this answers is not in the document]
piotmag769: Bubbling errors is a good point tbh

protostar-rust/src/main.rs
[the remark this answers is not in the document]
cptartur: I've added basic stripping of paths from test names, proper handling should be done in #2093 (and only after #2092 is addressed).

protostar-rust/src/lib.rs
```diff
@@ -18,6 +20,34 @@ use crate::test_stats::TestsStats;
 pub mod pretty_printing;
 mod test_stats;
 
+#[derive(Deserialize, Debug, PartialEq)]
+pub struct RunnerConfig {
+    test_name_filter: Option<String>,
+    exact_match: bool,
+    protostar_test_config: ProtostarTestConfig,
```
MaksymilianDemitraszek, line 27: Why this is a nested structure?

cptartur: `ProtostarTestConfig` is resolved from `Scarb.toml`, other arguments are cli only, because it doesn't make sense to have them in the toml imo.

protostar-rust/src/lib.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: Discussed online, my misunderstanding of the code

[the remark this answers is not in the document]
cptartur: I've tried and this cannot be simply replaced with filter_map, there's still a problem of propagating errors from it.

protostar-rust/tests/e2e/running.rs
```diff
@@ -14,22 +14,68 @@ fn run_simple_test() {
...
-            Tests: 5 passed, 1 failed
+            Running 1 test(s) from tests/without_prefix.cairo
+            [PASS] without_prefix::without_prefix::five
+            Tests: 7 passed, 1 failed
         "#});
 }
 
 #[test]
-fn run_declare_test() {
+fn with_filter() {
```
MaksymilianDemitraszek, line 35: Do we have a test for exact match?

MaksymilianDemitraszek (approved):
It would be nice to add exact match test, rn we don't know if it works or for example argument is hardcoded to 
False somewhere by mistake

protostar-rust/tests/e2e/running.rs
cptartur: Only unit, I'll add an e2e test. 

https://github.com/software-mansion/protostar/blob/1999-selecting-tests/protostar-rust/src/lib.rs#L506

protostar-rust/src/lib.rs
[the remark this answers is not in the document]
piotmag769: Agree, maybe we can have a struct with CLI_only_args or sth like that in the future

protostar-rust/src/lib.rs
```diff
@@ -134,6 +184,66 @@ fn run_tests(tests: TestsFromFile, tests_stats: &mut TestsStats) -> Result<()> {
...
+        .get_package(package)
+        .ok_or_else(|| anyhow!("Failed to find metadata for package = {package}"))?
+        .tool_metadata("protostar")
+        .ok_or_else(|| anyhow!("Failed to find protostar config for package = {package}"))?
+        .clone();
+    let protostar_config: ProtostarConfigFromScarb = serde_json::from_value(raw_metadata)?;
+
+    Ok(protostar_config)
+}
+
```
piotmag769, line 246: I think we can make the config not required in this PR already. Since we removed it last time, it wouldn'tt increase the scope of the PR

protostar-rust/src/lib.rs
```diff
@@ -134,6 +184,66 @@ fn run_tests(tests: TestsFromFile, tests_stats: &mut TestsStats) -> Result<()> {
...
+            .rsplit('/')
+            .next()
+            .with_context(|| format!("Failed to get test name from = {}", test_config.name))?
+            .into();
+        result.push(TestConfig {
+            name,
+            available_gas: test_config.available_gas,
+        });
+    }
+    Ok(result)
```
piotmag769, line 202: I feel like map would be better here, but I won't insist

protostar-rust/src/lib.rs
```diff
@@ -134,6 +184,66 @@ fn run_tests(tests: TestsFromFile, tests_stats: &mut TestsStats) -> Result<()> {
...
+) -> Result<Vec<TestConfig>> {
+    let mut result = vec![];
+    for test in test_configs {
+        if exact_match {
+            if test.name == test_name_filter {
+                result.push(test);
+            }
+        } else if test_name_contains(test_name_filter, &test)? {
+            result.push(test);
+        }
```
piotmag769, line 218: Feel like a few filters (as proposed offline) would be better here, but won't insist. Nested ifs are a bit messy

piotmag769 (commented):
Make sure to rebase and apply new panic printing format to e2e tests

piotmag769 (approved):

karol-bisztyga (approved):

PR #2059, opened 2023-06-12
Add declare rust

### LINKED ISSUE #1888: Implement `declare` cheatcode for rust test runner
We should implement `declare` cheatcode as a proving ground for other cheatcodes. It should declare cairo 1 contracts and use starknet in rust. It can be split into smaller tasks.

We probably should use https://github.com/lambdaclass/starknet_in_rust/blob/f4b80ed222381aea6e3f5a81047fa1c829d2fe97/crates/starknet-rs-py/src/starknet_state.rs#L35 to declare the contract.

We should build on 0.11 branch of starknet in rust.

We should use custon HintProcessor to handle cheatcode hints https://github.com/software-mansion-labs/cairo/blob/4094d447c77990142361f14b057f2e477a4d3130/crates/cairo-lang-runner/src/casm_run/mod.rs#L1315

### LINKED ISSUE #2048: Fetch the map contract name => path in rust CLI
This blocks #1888 and other cheat codes - we need to have this map from the config file in the rust code in order to be able to properly implement cheat codes.

karol-bisztyga:
### Checklist
- [ ] Added relevant tests

Cairo PR: https://github.com/software-mansion-labs/cairo/pull/88


protostar-rust/src/main.rs
```diff
@@ -66,16 +77,38 @@ fn dependencies_for_package(
 fn main_execution() -> Result<()> {
     let _args = Args::parse();
 
+    let project_root_path = project_root::get_project_root()?
+        .to_str()
+        .expect("error reading project root path")
+        .to_owned();
```
cptartur, line 83: We shoudn't rely on functions like this. The new runner will run all tests in scarb workspace, we have to resolve paths like this using scarb metadata.

protostar-rust/src/main.rs
```diff
@@ -66,16 +77,38 @@ fn dependencies_for_package(
...
+    };
+    for package in &scarb_metadata.packages {
+        for target in &package.targets {
+            if target.kind == "starknet-contract" {
+                // TODO consider multiple targets of this kind
+                protostar_test_config
+                    .contracts_paths
+                    .insert(target.name.clone(), target.source_path.to_string());
+            }
+        }
```
cptartur, line 103: Why limit the runner to starknet-contract target? `dependencies_for_package` is defined in such way that it will fetch `starknet-contract` target dependencies by default, but if it's not present, it will use `lib` target as fallback, because it's always present in scarb.

I don't see any reason to only run on starknet target.

pyrightconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ignore": ["./starknet_in_rust/", "./blockifier/"]
+}
```
cptartur, line 3: Newline missing

cptartur (changes requested):

protostar-rust/src/main.rs
piotmag769: Actually I think this approach is correct if we want to commit building contracts with scarb (I think that is the plan in the starknet CLI for now)
cc @MaksymilianDemitraszek

pyrightconfig.json
piotmag769: Why do we have starknet_in_rust here?

protostar-rust/src/lib.rs
```diff
@@ -82,7 +76,7 @@ fn internal_collect_tests(
 pub fn run_test_runner(
     input_path: &Utf8PathBuf,
     linked_libraries: Option<&Vec<LinkedLibrary>>,
-    config: &ProtostarTestConfig,
+    protostar_test_config: &ProtostarTestConfig,
```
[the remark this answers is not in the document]
karol-bisztyga: I think it is more descriptive = better

protostar-rust/src/main.rs
```diff
@@ -1,13 +1,23 @@
...
+use rust_test_runner::run_test_runner;
+
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug)]
+pub struct ProtostarTestConfigForDeserialization {
+    #[serde(default)]
+    #[warn(dead_code)]
+    exit_first: bool,
+}
```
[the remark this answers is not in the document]
karol-bisztyga: Well, this structure was added for deserialization that was not used in the end. I think we can remove this if we don't use it yet. I left it as is so it will be easier in the future to implement using the config.
This PR is not about deserialization/config and that should be done in a separate PR.

Options I see:
- we remove this
- we leave it as is

protostar-rust/cairo-tests/declare_test/protostar.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+protostar-version = "0.0.0"
+lib-path = "lib"
+
+[contracts]
+simple_contract = ["src"]
```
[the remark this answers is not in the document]
karol-bisztyga: [remove protostar toml](https://github.com/software-mansion/protostar/pull/2059/commits/864a79aa8a9f21168238bd298c9a7fc1c294b431)

.circleci/config.yml
```diff
@@ -142,15 +151,16 @@ commands:
               poetry run poe install_cairo_bindings prod
               poetry run poe install_cairo_bindings
               pushd cairo
-              rustup override set nightly-2023-06-01
+              rustup override set nightly-2022-11-03
```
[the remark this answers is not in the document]
karol-bisztyga: [fix version in config yml](https://github.com/software-mansion/protostar/pull/2059/commits/315ccc1fa7073b6dfb5af6d2751e199ed094580b)

pyrightconfig.json
karol-bisztyga: [fix pyright config](https://github.com/software-mansion/protostar/pull/2059/commits/d0b32557a8b4f9e580f0849bc99799badc445afb)

protostar-rust/src/lib.rs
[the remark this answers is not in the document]
piotmag769: Agree with Karol since we have `test_configs` variable that is of type `TestConfig` in this file

protostar-rust/src/main.rs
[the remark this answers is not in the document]
piotmag769: I would remove it if we don't use it yet, it will result in unnecessary clutter

protostar-rust/src/lib.rs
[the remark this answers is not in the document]
cptartur: Okay, let's keep it like this.

protostar-rust/src/main.rs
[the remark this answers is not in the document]
cptartur: Okay, we can remove it. Just let's remember that `ProtostarTestConfig` should be used directly for deserialization once it is needed, without any extra structs.

protostar-rust/src/main.rs
cptartur: We have to make this code working with what we have now. Once building is moved to rust, changes like this can be introduced.

cptartur: And if we already want to force the runner to only work with scarb packages with `starknet-contract` target, `dependencies_for_package` should be updated to only work with `starknet-contract` target.

I.e. this code should be updated
```rust
.min_by_key(|unit| match unit.target.name.as_str() {
            name @ "starknet-contract" => (0, name),
            name @ "lib" => (1, name),
            name => (2, name),
        })
        .ok_or_else(|| anyhow!("Failed to find metadata for package = {package}"))?;
```

piotmag769: We can stick with MVP I guess so I would lean towards your first suggestion

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "cairo"]
 	path = cairo
 	url = https://github.com/software-mansion-labs/cairo
+[submodule "blockifier"]
+	path = blockifier
+	url = https://github.com/karol-bisztyga/blockifier.git
```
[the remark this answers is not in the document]
MaksymilianDemitraszek: I agree we should move with fork for now

protostar-rust/src/lib.rs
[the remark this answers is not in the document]
MaksymilianDemitraszek: > I think it is more descriptive = better

It is nitpick but the information is already contatined in type singature, so it is kinda redundant this way

pyproject.toml
```diff
@@ -103,6 +103,9 @@ shell = "pytest -n auto protostar/*"
 target-version = [
     "py39",
 ]
+exclude = '''
+starknet_in_rust
+'''
```
MaksymilianDemitraszek, line 108: Why do we need to exclude starknet in rust?

protostar-rust/src/main.rs
```diff
@@ -66,16 +77,38 @@ fn dependencies_for_package(
...
     // TODO #1997
-    set_var("CARGO_MANIFEST_DIR", "../../cairo/Cargo.toml");
+    set_var(
+        "CARGO_MANIFEST_DIR",
+        format!("{}/../cairo/Cargo.toml", project_root_path),
+    );
 
     let scarb_metadata = MetadataCommand::new().inherit_stderr().exec()?;
 
+    let mut protostar_test_config = ProtostarTestConfig {
```
MaksymilianDemitraszek, line 92: As i mentioned in another PR, this is not a config, it is more of a TestData. Config suggests user configuration

protostar-rust/src/main.rs
MaksymilianDemitraszek, line 103: This info shouldn't be loaded from metadata, we should run scarb build to build contracts and use produced json artifact instead. It will be simpler to reason about and we will have less code on our side.

protostar-rust/src/main.rs
```diff
@@ -66,16 +77,38 @@ fn dependencies_for_package(
...
 
+    let project_root_path = project_root::get_project_root()?
+        .to_str()
+        .expect("error reading project root path")
+        .to_owned();
     // TODO #1997
-    set_var("CARGO_MANIFEST_DIR", "../../cairo/Cargo.toml");
+    set_var(
+        "CARGO_MANIFEST_DIR",
+        format!("{}/../cairo/Cargo.toml", project_root_path),
```
MaksymilianDemitraszek, line 87: Why we need to add project root path here 🤔 
cc @cptartur

protostar-rust/cairo-tests/declare_test/tests/declare_test.cairo
```diff
@@ -0,0 +1,8 @@
+use result::ResultTrait;
+use protostar_print::PrintTrait;
+
+#[test]
+fn test_declare_simple() {
+    let class_hash = declare('simple_contract').unwrap();
+    assert(class_hash == 273, 'proper class hash');
```
MaksymilianDemitraszek, line 7: Please inlcude tests:
- multiple contracts in a single package
- declare of not-exisiting contract

MaksymilianDemitraszek (changes requested):

protostar-rust/src/main.rs
cptartur: We don't need to. This code was removed entirely on the master branch. This PR should be rebased.

.gitmodules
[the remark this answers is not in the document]
karol-bisztyga: [update blockifier fork](https://github.com/software-mansion/protostar/pull/2059/commits/28c9067b0fdd3ee3a6fb585791c32d99d0a2c9fa)

protostar-rust/src/main.rs
karol-bisztyga: with new changes we also have `corelib_path` there which makes it a config I think.

protostar-rust/cairo-tests/declare_test/tests/declare_test.cairo
karol-bisztyga: I do not think this is a top priority. I think a top priority is: to declare a simple contract, deploy it, and invoke/call. Only then I would add what you suggest.

protostar-rust/src/main.rs
MaksymilianDemitraszek: > we also have corelib_path there which makes it a config I think.
We shouldn't have it there, Cairo hint processor shouldn't do anything related to compilation.

protostar-rust/cairo-tests/declare_test/tests/declare_test.cairo
MaksymilianDemitraszek: @karol-bisztyga I don't agree here, what we do is much more coplicated that the solution offloading compilation to Scarb. Tests can be done in a separate PR, but the compilation should be offloaded to scarb here. It will simplify things a lot in terms of complexity offloading a lot of stuff to scarb, which is a good thing

MaksymilianDemitraszek: If you think conflicts are problem, I am ok with introducing those changes in a separate PR but it should be done before anything else.

protostar-rust/src/main.rs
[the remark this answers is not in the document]
karol-bisztyga: [remove unnecessary config](https://github.com/software-mansion/protostar/pull/2059/commits/3ff2bdfb5fde9ceec5982f58545177381d65a0cb)

protostar-rust/cairo-tests/declare_test/tests/declare_test.cairo
karol-bisztyga: ok let's defer to separate PRs

karol-bisztyga: I think there's a misunderstanding here: I only said that I think we should do the tests you mentioned later, not now. I agree we should move building to scarb. I only referred to the tests.

protostar-rust/src/main.rs
karol-bisztyga: it is removed with building in scarb
[adapt for building in scarb](https://github.com/software-mansion/protostar/pull/2059/commits/b104f483c4ffb234aba0256cd0a8d988e727484b)

protostar-rust/src/main.rs
karol-bisztyga: this is removed with building in scarb
[adapt for building in scarb](https://github.com/software-mansion/protostar/pull/2059/commits/b104f483c4ffb234aba0256cd0a8d988e727484b)

protostar-rust/src/main.rs
```diff
@@ -29,17 +29,10 @@ fn main_execution() -> Result<()> {
     let scarb_metadata = MetadataCommand::new().inherit_stderr().exec()?;
 
     for package in &scarb_metadata.workspace.members {
-        let protostar_config =
-            rust_test_runner::protostar_config_for_package(&scarb_metadata, package)?;
         let (base_path, dependencies) =
             rust_test_runner::dependencies_for_package(&scarb_metadata, package)?;
```
cptartur, line 33: Where are these dependencies resolved now if they got removed from here? Does outputs of `scarb build` handled in the cairo runner account for these?

protostar-rust/src/lib.rs
```diff
@@ -60,7 +54,7 @@ fn find_cairo_files_in_directory(input_path: &Utf8PathBuf) -> Result<Vec<Utf8Pat
 
 fn internal_collect_tests(
     input_path: &Utf8PathBuf,
-    linked_libraries: Option<&Vec<LinkedLibrary>>,
+    linked_libraries: Option<Vec<LinkedLibrary>>,
```
cptartur, line 57: Why was this changed? We aren't consuming this value anyways, it's causing the lints to fail.

protostar-rust/cairo-tests/declare_test/tests/declare_test.cairo
karol-bisztyga: Discussed offline, the last 2 comments of Maks are irrelevant, resolving.

protostar-rust/src/lib.rs
karol-bisztyga: Yes, I changed it to a ref back again, this time not nested.

protostar-rust/src/lib.rs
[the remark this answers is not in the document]
karol-bisztyga: Sorry, but no, the type is shown in just one place and further in the code if you have a descriptive name it's way easier to read.

protostar-rust/src/main.rs
karol-bisztyga: Can you specify what you mean exactly? 
This is not removed:
```
let (base_path, dependencies) =
            rust_test_runner::dependencies_for_package(&scarb_metadata, package)?;
```
This is removed:
```
let protostar_config =
            rust_test_runner::protostar_config_for_package(&scarb_metadata, package)?;
```
as we agreed and it's not done anywhere instead, we agreed to resign from this in https://github.com/software-mansion/protostar/pull/2059#discussion_r1226632012

protostar-rust/src/lib.rs
```diff
@@ -32,7 +26,7 @@ struct TestsFromFile {
 
 fn collect_tests_from_directory(
     input_path: &Utf8PathBuf,
-    linked_libraries: Option<&Vec<LinkedLibrary>>,
+    linked_libraries: &Option<Vec<LinkedLibrary>>,
```
piotmag769, line 29: I think it is more idiomatic to do it the way it was before (here and in other places)
https://github.com/software-mansion/protostar/pull/2036#discussion_r1224144914

protostar-rust/tests/data/declare_test/Scarb.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "declare_test"
+version = "0.1.0"
+
+[[target.starknet-contract]]
+sierra = true
+
+[dependencies]
+starknet = ">=1.0.0"
```
piotmag769, line 9:
```suggestion
starknet = ">=1.1.0"
```
It should match our current compiler version

protostar-rust/src/main.rs
piotmag769: I think we meant to remove the config for serialization, not the config in general. ProtostarTestConfig is used by the test runner, we cannot remove it

karol-bisztyga: @cptartur can u confirm/clarify?

protostar-rust/src/lib.rs
karol-bisztyga: ok fair

protostar-rust/src/main.rs
karol-bisztyga: Discussed offline, @piotmag769 misunderstood, @cptartur pls either clarify or resolve, thx.

protostar-rust/tests/data/declare_test/Scarb.toml
karol-bisztyga: [fdbck](https://github.com/software-mansion/protostar/pull/2059/commits/f83f77d95fc28d6557dbd04d98a83d3bce5938b9)

protostar-rust/src/lib.rs
karol-bisztyga: [fdbck](https://github.com/software-mansion/protostar/pull/2059/commits/f83f77d95fc28d6557dbd04d98a83d3bce5938b9)

protostar-rust/tests/data/declare_test/tests/test_declare.cairo
```diff
@@ -0,0 +1,8 @@
+use result::ResultTrait;
+use protostar_print::PrintTrait;
+
+#[test]
```
MaksymilianDemitraszek, line 4: Do we have an issue to extend those tests as requested?

protostar-rust/src/lib.rs
```diff
@@ -60,7 +56,7 @@ fn find_cairo_files_in_directory(input_path: &Utf8PathBuf) -> Result<Vec<Utf8Pat
 
 fn internal_collect_tests(
     input_path: &Utf8PathBuf,
-    linked_libraries: Option<&Vec<LinkedLibrary>>,
+    linked_libraries: Option<Vec<LinkedLibrary>>,
```
karol-bisztyga, line 59: Ok, I changed it because otherwise we'd end up with errors from the compiler, I'm not going to spend more time on this, we can clone some arrays of length < 10 2 or 3 times and no one will get hurt and nothing's gonna break. If that's so much of a big deal, you can always submit an issue and we can address this later, it's a micro-optimization that doesn't matter much.

protostar-rust/src/lib.rs
```diff
@@ -99,6 +94,12 @@ pub fn run_test_runner(
         tests.len(),
     );
 
+    SnapboxCommand::new("scarb")
+        .arg("build")
+        .current_dir(std::env::current_dir().expect("failed to obtain current dir"))
+        .assert()
+        .success();
```
MaksymilianDemitraszek, line 101: I believe we could call it once at the beginning of the execution, right now we call it in the loop, once for each target.

MaksymilianDemitraszek: I also feel like we shouldn't use Snapbox here I created a task to change it #2082

protostar-rust/src/main.rs
MaksymilianDemitraszek: Those are resolved by scarb under the hood when we call build, does it clarifies it @cptartur ?

MaksymilianDemitraszek (approved):

protostar-rust/tests/data/declare_test/tests/test_declare.cairo
karol-bisztyga: https://github.com/software-mansion/protostar/issues/2083

protostar-rust/src/main.rs
cptartur: Yes

protostar-rust/src/lib.rs
karol-bisztyga: [updt](https://github.com/software-mansion/protostar/pull/2059/commits/2e3d9eb44a8158df46fa667e9b5f470643759b29)

cptartur (approved):

PR #2061, opened 2023-06-12
Add print rust

### LINKED ISSUE #1993: Implement `Print` cheat code in rust
We should add the code for the `Print` cheat code.
The implementation should be put [here](https://github.com/software-mansion-labs/cairo/blob/a8ca26f2c50cab560c2e2ac98c6a8deabb332996/crates/cairo-lang-runner/src/casm_run/mod.rs#L774).

We were considering using [starknet_in_rust](https://github.com/lambdaclass/starknet_in_rust) and [blockifier](https://github.com/starkware-libs/blockifier), and we decided to go with the latter.

karol-bisztyga:
### Checklist
- [x] Added relevant tests

https://github.com/software-mansion-labs/cairo/pull/89


protostar-rust/run_tests.sh
```diff
@@ -0,0 +1,20 @@
...
+  exit 0;
+fi
+
+for DIR in */ ; do
+  pushd "$DIR"
+  cargo run
+  popd # $DIR
+done
+
+popd # cairo-tests
```
cptartur, line 20: We shouldn't have to use scripts like this. See e2e tests in #2033. This should be rebased.

cptartur (changes requested):

protostar-rust/cairo-tests/declare_test/protostar.toml
piotmag769, line 1: We got rid of protostar.toml in the new runner

protostar-rust/cairo-tests/print_test/tests/simple_test.cairo
piotmag769, line 1: Can we add tests for printing felts that:
a) do not fit into 31 bytes
b) fit into 31 bytes but are not decodable to ascii

We didn't handle this cases in python runner, currently being fixed in https://github.com/software-mansion/protostar/pull/2070, I think we can do it in a similar way here

pyrightconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ignore": ["./starknet_in_rust/", "./blockifier/"]
```
piotmag769, line 2: Why do we ignore `starknet_in_rust` here?

piotmag769 (changes requested):
I assume a lot will change in this PR when suggestion from https://github.com/software-mansion/protostar/pull/2059 are applied, requesting changes just to make sure to review it later on

piotmag769 (approved):

cptartur (approved):

PR #2070, opened 2023-06-12
Check if int is decodable to ascii before doing `short_string_to_str`

### LINKED ISSUE #2067: Error when cairo panics with felt252 that is not a short string
When cairo panics with `felt252` that is not a short string, we will have an error decoding it to ascii in 
https://github.com/software-mansion/protostar/blob/9dba4f3475e1d887b0e44d3edacdd347107663fa/protostar/cairo_testing/cairo_injectable_function_runner.py#L52-L54

https://github.com/software-mansion/protostar/blob/9dba4f3475e1d887b0e44d3edacdd347107663fa/protostar/cairo/short_string.py#L8

Example of error:
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0x88 in position 30: ordinal not in range(128)
```

Example of the test that produces the above error
```
#[test]
fn test_panic() {
    let temp = 1234;
    assert(1==temp, temp);
}
```

abulenok:
### Checklist
- [x] Added relevant tests


tests/integration/cairo1/misc/panic_test/panic_test.cairo
```diff
@@ -0,0 +1,5 @@
+#[test]
+fn test_panic() {
+    let temp = 1234;
+    assert(1==temp, temp);
```
piotmag769, line 4: [nit]
```suggestion
    assert(1 == temp, temp);
```

protostar/cairo_testing/cairo_injectable_function_runner.py
```diff
@@ -50,7 +50,8 @@ def run_cairo_function_by_offset(
             if self._cairo_runner_facade.did_panic():
                 panic_data = self._cairo_runner_facade.get_panic_data()
                 panic_data_short_strs = [
-                    short_string_to_str(datum) for datum in panic_data
+                    short_string_to_str(datum) if is_decodable_to_ascii(datum) else None
+                    for datum in panic_data
```
piotmag769, line 54: I think we can hide this logic in `short_string_to_str` and just return `None` from it if datum is not decodable to ascii

protostar/cairo/short_string.py
```diff
@@ -1,8 +1,21 @@
...
+
+def is_decodable_to_ascii(value: int) -> bool:
+    for byte in _to_bytes(value):
+        if byte > LAST_ASCII_CHAR_DEC_VAL:
+            return False
+    return True
+
+
+def _to_bytes(value: int) -> bytes:
+    return value.to_bytes(length=31, byteorder="big")
```
piotmag769, line 21: Isn't there technically a possibility that we will encounter a value that occupies more than 31 bytes? Currently  P = 2^251 + 17 * 2^192 + 1 so in the worst case the felt passed there could be 252 bits = 31.5 bytes long. Of course it wouldn't happen if someone passes there a short string literal (since every ascii character is encoded using one whole byte), but if the program panics with some large felt252 like P - 1 we can get an error here. I think we should just set length to 32

tests/integration/cairo1/misc/panic_test/data/simple/lib.cairo
piotmag769, line 1: Do we need this file?

tests/integration/cairo1/misc/panic_test/test_panic.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pytest.fixture(name="protostar_project", scope="function")
+def protostar_fixture(create_protostar_project: CreateProtostarProjectFixture):
+    with create_protostar_project() as protostar_project:
+        yield protostar_project
+
+
+async def test_panic(protostar_project: ProtostarProjectFixture, shared_datadir: Path):
+    protostar_project.create_contracts_cairo1({"simple": shared_datadir / "simple"})
```
piotmag769, line 19: Do we need it here?

piotmag769 (changes requested):

protostar/cairo_testing/cairo_injectable_function_runner.py
abulenok: I think this function should fail / throw error whenever we try to convert something that is not a short string to a string. This function is also used in other places, and I believe that we want to know when the argument we are passing to the function is invalid.

abulenok: But the error reporting could be improved, I'll make a commit in a sec

protostar/cairo/short_string.py
abulenok: I guess there is no reason to convert large felt252 which is not a short string to string. Furthermore, it could contain bytes that are not decodable to ascii. If the felt is a short string, then it contains 31 bytes of meaningful information. If it's not then there is no reason to convert it to ascii string (and this conversion can fail)

protostar/cairo/short_string.py
```diff
@@ -1,8 +1,24 @@
...
     """
     Reverse of starkware.cairo.lang.compiler.test_utils.short_string_to_felt
     """
-    return value.to_bytes(length=32, byteorder="big").decode("ascii").lstrip("\x00")
+    return _to_bytes(value).decode("ascii").lstrip("\x00")
+
+
+def is_short_string(value: int) -> bool:
+    if value > int.from_bytes(b"\xff" * 31, "big"):
+        return False
```
cptartur, line 15: Isn't this check redundant if we are validating every character to be in ASCII range below?

protostar/cairo/short_string.py
```diff
@@ -1,8 +1,24 @@
...
+
+    for byte in _to_bytes(value):
+        if byte > LAST_ASCII_CHAR_DEC_VAL:
+            return False
```
cptartur, line 19: Should we be also checking that it's in printable character range? That is >= 32 I think?

tests/e2e/cairo1/test_cairo1_testing.py
```diff
@@ -51,8 +51,12 @@ def test_failing_tests(protostar: ProtostarFixture, copy_fixture: CopyFixture):
         "Test failed with data: ",
         "[1870930782904301745253, 482670963043, 31066316372818838395891839589] (integer representation)",
         "['elaborate', 'panic', 'data is here'] (short-string representation)",
+        "[FAIL] tests/failing_test.cairo test_assert_err_code_not_ascii",
+        "Test failed with data:",
+        "[1234] (integer representation)",
+        "[None] (short-string representation)",
```
cptartur, line 57: Shouldn't non decodable values be presented as `""` or something like that instead of `None`?

tests/integration/cairo1/misc/test_panic/panic_test.cairo
```diff
@@ -0,0 +1,5 @@
+#[test]
+fn test_panic() {
+    let temp = 1234;
+    assert(1 == temp, temp);
+}
```
cptartur, line 5: Newline missing

protostar/cairo/short_string.py
piotmag769: No, it is not. There is a possibility that passed fetl252 will be too large to fit in 31 bytes (248 bits). In this case we can be sure it was not a short string literal and just refrain from decoding it, as it doesn't make any sense

piotmag769 (commented):
Can we also add a tests for:
1) printing large felts that doesn't fit in 31 bytes (just add to `tests/integration/cairo1/compiler_bindings/test_print/print_test.cairo`)
2) panicing with large felts that doesn't fit in 31 bytes

protostar/cairo_testing/cairo_injectable_function_runner.py
piotmag769: Could we reuse this logic in `protostar/cheatable_starknet/controllers/io.py` though? It would be more readable imo

protostar/cairo_testing/cairo_injectable_function_runner.py
```diff
@@ -50,7 +50,8 @@ def run_cairo_function_by_offset(
             if self._cairo_runner_facade.did_panic():
                 panic_data = self._cairo_runner_facade.get_panic_data()
                 panic_data_short_strs = [
-                    short_string_to_str(datum) for datum in panic_data
+                    short_string_to_str(datum) if is_short_string(datum) else None
+                    for datum in panic_data
```
piotmag769, line 54: Could we reuse this logic in `protostar/cheatable_starknet/controllers/io.py`? It would be more readable imo

protostar/cairo/short_string.py
abulenok: I think we should be consistent with cairo, which states that the short string consists of ascii characters. I don't think they limit ascii characters somehow

cptartur: Right, but how would decoding work if someone passed `0` for example? This would get decoded to `null`. Or maybe it's impossible for values like this to show up?

tests/e2e/cairo1/test_cairo1_testing.py
cptartur: I guess we can keep it as `None` then.

tests/integration/cairo1/misc/test_panic/panic_test.cairo
abulenok: That's funny but it seems that GitHub is not displaying empty lines at the end of the file

protostar/cairo_testing/cairo_injectable_function_runner.py
abulenok: added

protostar/cairo/short_string.py
abulenok: It's possible to panic with such values
In case of something like `assert(1 == 2, 20)` protostar will print
```
Test failed with data: 
[20] (integer representation)
['\x14'] (short-string representation)
```

abulenok: so we could limit it to a printable character range, but at the same time some numbers can represent printable ascii chars, not carrying any meaningful information at the same time
like `16705` will be decoded to `'AA'`

abulenok: I guess we can leave it as is, but it's not a strong opinion

tests/integration/cairo1/misc/test_panic.py
```diff
@@ -0,0 +1,29 @@
...
+        datadir / "panic_test.cairo",
+    )
+
+    assert_cairo_test_cases(
+        testing_summary,
+        expected_failed_test_cases_names=[
+            "test_panic_when_felt252_not_short_str",
+            "test_panic_with_felt252_max_val",
+        ],
+    )
```
piotmag769, line 29: I think we should check the output just like in `tests/integration/cairo1/compiler_bindings/test_print.py`

abulenok: Added

piotmag769 (approved):

PR #2077, opened 2023-06-14
Include corelib in rust runner binary

### LINKED ISSUE #2066: Include corelib in rust test runner binary
Currently, rust test runner corelib detection depends on either passing `--corelib-path` or hardcoded paths relative to protostar directory root. First solution is problematic and the second one will not work in case we try to run test runner in any other context than its current directory in protostar repository.

Until #1997 is resolved, we have to figure out a solution for bundling `corelib` with the runner binary so that it works without passing any arguments.

New CI tasks should be introduced as well for building the binary.

cptartur:
Closes #2066 

### Checklist
- [ ] Added relevant tests

### DIFF
--- a/protostar-rust/src/main.rs
+++ b/protostar-rust/src/main.rs
@@ -3,7 +3,8 @@ use camino::Utf8PathBuf;
 use clap::Parser;
 use include_dir::{include_dir, Dir};
 use scarb_metadata::MetadataCommand;
-use tempfile::tempdir;
+use std::path::PathBuf;
+use tempfile::{tempdir, TempDir};
 
 use rust_test_runner::pretty_printing;
 use rust_test_runner::run_test_runner;
@@ -17,20 +18,22 @@ struct Args {
     test_filter: Option<String>,
 }
 
-fn load_corelib() -> Result<Utf8PathBuf> {
+fn load_corelib() -> Result<TempDir> {
     let tmp_dir = tempdir()?;
     CORELIB
         .extract(&tmp_dir)
         .expect("Failed to copy corelib to temporary directory");
-    Utf8PathBuf::try_from(tmp_dir.into_path())
-        .context("Failed to convert corelib path to Utf8PathBuf")
+    Ok(tmp_dir)
 }
 
 fn main_execution() -> Result<()> {
     let _args = Args::parse();
 
     // TODO #1997
-    let corelib = load_corelib()?;
+    let corelib_dir = load_corelib()?;
+    let corelib_path: PathBuf = corelib_dir.path().into();
+    let corelib = Utf8PathBuf::try_from(corelib_path)
+        .context("Failed to convert corelib path to Utf8PathBuf")?;
 
     let scarb_metadata = MetadataCommand::new().inherit_stderr().exec()?;
 
@@ -45,6 +48,9 @@ fn main_execution() -> Result<()> {
 
         run_test_runner(&base_path, Some(dependencies.clone()), Some(&corelib))?;
     }
+
+    // Explicitly close the temporary directory so we can handle the error
+    corelib_dir.close().expect("Failed to close temporary directory with corelib. Corelib files might have not been released from filesystem");
     Ok(())
 }
 
--- a/protostar-rust/src/main.rs
+++ b/protostar-rust/src/main.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Context, Result};
 use camino::Utf8PathBuf;
 use clap::Parser;
 use include_dir::{include_dir, Dir};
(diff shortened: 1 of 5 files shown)


piotmag769 (approved):
LGTM

karol-bisztyga (changes requested):
blocked by https://github.com/software-mansion/protostar/pull/2059

piotmag769 (approved):

protostar-rust/src/main.rs
```diff
@@ -1,30 +1,37 @@
...
+    let tmp_dir = tempdir()?;
+    CORELIB
+        .extract(&tmp_dir)
+        .expect("Failed to copy corelib to temporary directory");
+    Ok(tmp_dir)
 }
 
 fn main_execution() -> Result<()> {
-    let args = Args::parse();
+    let _args = Args::parse();
```
karol-bisztyga, line 28: why is this not removed?

protostar-rust/src/main.rs
```diff
@@ -1,30 +1,37 @@
...
 use clap::Parser;
+use include_dir::{include_dir, Dir};
 use scarb_metadata::MetadataCommand;
+use std::path::PathBuf;
+use tempfile::{tempdir, TempDir};
 
 use rust_test_runner::pretty_printing;
 use rust_test_runner::run_test_runner;
 
+static CORELIB: Dir = include_dir!("../cairo/corelib/src");
```
karol-bisztyga, line 12:
```suggestion
static CORELIB_PATH: Dir = include_dir!("../cairo/corelib/src");
```
unclear naming

protostar-rust/src/main.rs
```diff
@@ -1,30 +1,37 @@
...
+    let _args = Args::parse();
 
     // TODO #1997
-    let corelib = match args.corelib_path {
-        Some(corelib) => Utf8PathBuf::from(corelib),
-        None => Utf8PathBuf::from("../../cairo/corelib/src")
-            .canonicalize_utf8()
-            .context("Failed to resolve corelib path")?,
-    };
+    let corelib_dir = load_corelib()?;
```
karol-bisztyga, line 31: Why is `load_corelib` a separate function? This code isn't reused and is not even big. I think it should be inlined here.

protostar-rust/src/main.rs
karol-bisztyga, line 12: why is this static? It's used only in one place.

karol-bisztyga (changes requested):

protostar-rust/src/main.rs
MaksymilianDemitraszek: I am not sure tbh, it abstracts the operation nicely

karol-bisztyga: It can be brought down to literally one instruction:
```
CORELIB
        .extract(&tempdir()?)
        .expect("Failed to copy corelib to temporary directory");
```

MaksymilianDemitraszek: Yea but looking at it i am not sure what it does at first glance and function names it

karol-bisztyga: LoL, the error message tells you exactly what it does:

> Failed to copy corelib to temporary directory

which means that what it does is: "copy corelib to temporary directory". It's even more descriptive than the function name "load_corelib" (btw isn't this name misleading? "load corelib" does not indicate that we copy it to a tmp dir, it might also mean that we load this into the memory or something which brings confusion).

> I would stick with a function, this way we reduce cognitive complexity imo

Extracting single instructions into separate once-used functions looks like an anti-pattern to me. This leads to a code I'd not wish to maintain. Also, as I mentioned above, the name of this function is not quite compatible with the error message and with what's actually happening and it is misleading.

Long story short, I know it's 2v1 guys and usually in such a case I back off right away but here I find your arguments not super legit, sorry 🤷 At the end of the day I'm only an advisor here, the author may do as he wishes.

cptartur: I'd rather have multiple small functions in the codebase that make it readable. From the point of person doing something in context of `fn main`, how exactly is corelib loaded is not important at all.

protostar-rust/src/main.rs
cptartur: The directory is being included in the binary at the compile time so there's no reason to do the same with the path to it.

protostar-rust/src/main.rs
cptartur: Because it is being used in other PR (#1999). The CLI main interface prototype was designed for a specific purpose of handling arguments of the CLI.

karol-bisztyga (approved):

protostar-rust/src/main.rs
karol-bisztyga: Discussed offline

> ... there's no reason to do the same with the path to it.

It should be ("not" missing, therefore confusion)

```
... there's no reason not to do the same with the path to it.
```

PR #2088, opened 2023-06-17
Panic printing ascii

### LINKED ISSUE #2078: Ascii instead of utf8 in panic printing
We should just use `as_cairo_short_string` from cairo.

TODO:
- [x] Change decoding from utf8 to ascii
- [x] Print original values, decoded values only if possible

Add tests for panicking with/encoding values that:
- [x] do not fit into 31 bytes
- [x] are not ascii encoded
- [x] are shortstring values

piotmag769:
### Checklist
- [x] Added relevant tests

### DIFF
--- a/protostar-rust/src/pretty_printing.rs
+++ b/protostar-rust/src/pretty_printing.rs
@@ -3,6 +3,7 @@ use cairo_felt::Felt252;
 use camino::Utf8PathBuf;
 use console::style;
 
+use cairo_lang_runner::short_string::as_cairo_short_string;
 use cairo_lang_runner::RunResultValue;
 
 use crate::test_stats::TestsStats;
@@ -45,7 +46,7 @@ pub fn print_test_result(test_name: &str, result_value: &RunResultValue) {
         println!("[{result_tag}] {test_name}");
     } else {
         let result_str_bold = style(result_str).bold();
-        println!("[{result_tag}] {test_name} {result_str_bold}");
+        println!("[{result_tag}] {test_name}{result_str_bold}");
     }
 }
 
@@ -68,10 +69,11 @@ fn result_data_to_text(data: &[Felt252]) -> String {
     let mut readable_text = String::new();
 
     for felt in data {
-        let felt_bytes = felt.to_bytes_be();
-        let felt_text = String::from_utf8_lossy(&felt_bytes);
-        readable_text.push_str(&felt_text);
+        readable_text.push_str(&format!("\noriginal value: [{felt}]"));
+        if let Some(short_string) = as_cairo_short_string(felt) {
+            readable_text.push_str(&format!(", converted to a string: [{short_string}]"));
+        }
     }
 
-    readable_text.trim().to_string()
+    readable_text
 }
--- a/protostar-rust/src/pretty_printing.rs
+++ b/protostar-rust/src/pretty_printing.rs
@@ -71,7 +71,7 @@ fn result_data_to_text(data: &[Felt252]) -> String {
     let mut readable_text = String::new();
 
     for felt in data {
-        readable_text.push_str(&format!("\noriginal value: [{felt}]"));
+        readable_text.push_str(&format!("\n    original value: [{felt}]"));
         if let Some(short_string) = as_cairo_short_string(felt) {
             readable_text.push_str(&format!(", converted to a string: [{short_string}]"));
         }
--- a/protostar-rust/src/pretty_printing.rs
+++ b/protostar-rust/src/pretty_printing.rs
@@ -34,6 +34,7 @@ pub fn print_test_summary(tests_stats: TestsStats) {
 
 pub fn print_test_result(test_name: &str, result_value: &RunResultValue) {
     let result_str = get_result_str(result_value);
+    let result_str_bold = style(result_str.clone()).bold();
 
     let passed = did_pass(result_value);
     let result_tag = if passed {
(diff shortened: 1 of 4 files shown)


karol-bisztyga (approved):

protostar-rust/tests/e2e/running.rs
```diff
@@ -73,3 +74,29 @@ fn run_print_test() {
...
+        .stdout_matches(indoc! {r#"Collected 1 test(s) and 2 test file(s)
+            Running 0 test(s) from src/lib.cairo
+            Running 1 test(s) from tests/test_panic_decoding.cairo
+            [FAIL] test_panic_decoding::test_panic_decoding::test_panic_decoding
+            original value: [123], converted to a string: [{]
+            original value: [6381921], converted to a string: [aaa]
+            original value: [3618502788666131213697322783095070105623107215331596699973092056135872020480]
+            original value: [152]
+            original value: [124], converted to a string: [|]
+            original value: [149]
```
cptartur, line 99: Can we change it so we have the "trace" so the error message is visually separated? 

So it's like this:

```
Running 0 test(s) from src/lib.cairo
Running 1 test(s) from tests/test_panic_decoding.cairo

[FAIL] test_panic_decoding::test_panic_decoding::test_panic_decoding

Failure data:
    original value: [123], converted to a string: [{]
    original value: [6381921], converted to a string: [aaa]
    original value: 
    [3618502788666131213697322783095070105623107215331596699973092056135872020480]
    original value: [152]
    original value: [124], converted to a string: [|]
    original value: [149]

Tests: 0 passed, 1 failed
```

Also I'd add extend this test case with couple passing tests after and before this so we can see how this behaves in context of running multiple tests.

cptartur (changes requested):

protostar-rust/tests/e2e/running.rs
piotmag769: Good idea

protostar-rust/tests/e2e/running.rs
```diff
@@ -73,3 +77,39 @@ fn run_print_test() {
...
+            [PASS] test_panic_decoding::test_panic_decoding::test_simple
+            [FAIL] test_panic_decoding::test_panic_decoding::test_panic_decoding
+            
+            Failure data:
+            original value: [123], converted to a string: [{]
+            original value: [6381921], converted to a string: [aaa]
+            original value: [3618502788666131213697322783095070105623107215331596699973092056135872020480]
+            original value: [152]
+            original value: [124], converted to a string: [|]
+            original value: [149]
```
cptartur, line 105: I think this could be indented so if you have a lot of tests it's more clearly visible. What do you think?

protostar-rust/src/pretty_printing.rs
```diff
@@ -43,9 +45,10 @@ pub fn print_test_result(test_name: &str, result_value: &RunResultValue) {
 
     if result_str.is_empty() {
         println!("[{result_tag}] {test_name}");
+    } else if passed {
+        println!("[{result_tag}] {test_name}\n\nSuccess data:{result_str_bold}\n");
```
cptartur, line 49: Is it possible to have success data at all?

protostar-rust/tests/e2e/running.rs
piotmag769: Tried it and I think in console it looks a bit worse, especially a smaller one
image
image
But I don't have a strong opinion on it

protostar-rust/src/pretty_printing.rs
```diff
@@ -68,10 +71,11 @@ fn result_data_to_text(data: &[Felt252]) -> String {
     let mut readable_text = String::new();
 
     for felt in data {
-        let felt_bytes = felt.to_bytes_be();
-        let felt_text = String::from_utf8_lossy(&felt_bytes);
-        readable_text.push_str(&felt_text);
+        readable_text.push_str(&format!("\n    original value: [{felt}]"));
```
piotmag769, line 74: I cannot `\t` due to `stdout_matches` not recognising 4 spaces as a tab (and I guess we want to keep raw strings in tests)

protostar-rust/src/pretty_printing.rs
piotmag769: Not sure, but the compiler seems to allow it
```rust
/// The ran function return value.
#[derive(Debug, Eq, PartialEq)]
pub enum RunResultValue {
    /// Run ended successfully, returning the memory of the non-implicit returns.
    Success(Vec<Felt252>),
    /// Run panicked, returning the carried error data.
    Panic(Vec<Felt252>),
}
```

protostar-rust/tests/e2e/running.rs
cptartur: I think I like the indented one a bit more.

protostar-rust/src/pretty_printing.rs
cptartur: Okay, better to explicitly handle it then. We can consider creating an issue in triage for investigating this further.

cptartur: cc @MaksymilianDemitraszek

cptartur (approved):

protostar-rust/src/pretty_printing.rs
MaksymilianDemitraszek: It is determined based on first felt if 0 then it is success otherwise panic,

Issue #2107, opened 2023-06-22
Normalize outputs in CLI


THenry14:
We should decide on one approach to outputting information to users, and apply it (Felts, Ints, formatting etc).

Suggestions: 
https://docs.rs/console/latest/console/
https://docs.rs/color-eyre/latest/color_eyre/

We could take a look at what scarb is using, and try to conform to that.

piotmag769:
I would also emphasise on the need of returning non-zero exit code on failure - it will be useful in scripts and that's what we do in test runner as well.
cc @THenry14 @war-in

PR #2106, opened 2023-06-21
Add invoke to CLI

cptartur:
Closes #2053 

### Checklist
- [ ] Relevant issue is linked
- [ ] Docs updated/issue for docs created
- [ ] Added relevant tests

PR #2110, opened 2023-06-22
[cli] add call command

### LINKED ISSUE #2054: Add call command
Add call command to protostar-cli.

The command should be documented and have tests.

war-in:
### Checklist
- [ ] Added relevant tests

Closes #2054

### DIFF
--- a/protostar-cli/src/lib.rs
+++ b/protostar-cli/src/lib.rs
@@ -9,6 +9,8 @@ use starknet::{
     signers::{LocalWallet, SigningKey},
 };
 use std::fs;
+use starknet::core::types::BlockId;
+use starknet::core::types::BlockTag::{Latest, Pending};
 use url::Url;
 
 #[derive(Deserialize, Serialize)]
@@ -77,4 +79,15 @@ pub fn get_account<'a>(
     let mut account = SingleOwnerAccount::new(provider, signer, address, network.get_chain_id());
 
     Ok(account)
+}
+
+pub fn get_block_id(value: &str) -> Result<BlockId> {
+    match value {
+        "pending" => Ok(BlockId::Tag(Pending)),
+        "latest" => Ok(BlockId::Tag(Latest)),
+        _ => Err(anyhow::anyhow!(
+            "No such block id {}! Possible values are pending and latest for now.",
+            value
+        ))
+    }
 }
--- a/protostar-cli/src/lib.rs
+++ b/protostar-cli/src/lib.rs
@@ -2,6 +2,8 @@ use anyhow::{anyhow, Result};
 use camino::Utf8PathBuf;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use starknet::core::types::BlockId;
+use starknet::core::types::BlockTag::{Latest, Pending};
 use starknet::{
     accounts::SingleOwnerAccount,
     core::{chain_id, types::FieldElement},
@@ -9,8 +11,6 @@ use starknet::{
     signers::{LocalWallet, SigningKey},
 };
 use std::fs;
-use starknet::core::types::BlockId;
-use starknet::core::types::BlockTag::{Latest, Pending};
 use url::Url;
 
 #[derive(Deserialize, Serialize)]
@@ -88,6 +88,6 @@ pub fn get_block_id(value: &str) -> Result<BlockId> {
         _ => Err(anyhow::anyhow!(
             "No such block id {}! Possible values are pending and latest for now.",
             value
-        ))
+        )),
     }
 }
--- a/protostar-cli/src/lib.rs
+++ b/protostar-cli/src/lib.rs
@@ -82,6 +82,7 @@ pub fn get_account<'a>(
 }
 
 pub fn get_block_id(value: &str) -> Result<BlockId> {
+    // todo: add more block ids (hash, number)
     match value {
         "pending" => Ok(BlockId::Tag(Pending)),
         "latest" => Ok(BlockId::Tag(Latest)),
--- a/protostar-cli/src/main.rs
+++ b/protostar-cli/src/main.rs
@@ -87,11 +87,11 @@ async fn main() -> Result<()> {
             let block_id = get_block_id(&args.block_id).unwrap();
 
             let result = starknet_commands::call::call(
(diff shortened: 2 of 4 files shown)


protostar-cli/src/starknet_commands/call.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Block identifier on which call should be performed
+    #[clap(short = 'b', long = "block-id", default_value = "pending")]
+    pub(crate) block_id: String,
+}
+
+pub async fn call(
+    contract_address: &str,
+    func_name: &str,
+    calldata: &Option<Vec<String>>,
```
cptartur, line 31: https://users.rust-lang.org/t/api-design-option-t-vs-option-t/34139

cptartur: I think we should prefer `Option<&Vec<String>>`.

protostar-cli/src/starknet_commands/call.rs
```diff
@@ -0,0 +1,48 @@
...
+    provider: &JsonRpcClient<HttpTransport>,
+    block_id: &BlockId,
+) -> Result<Vec<FieldElement>, Error> {
```
cptartur, line 34: With `anyhow::Result` you can just write `Result<Vec<FieldElement>>` without specifying the Error.

cptartur, line 34: With `anyhow::Result` you can just write `Result<Vec<FieldElement>>` without specifying the Error.

protostar-cli/src/starknet_commands/call.rs
```diff
@@ -0,0 +1,48 @@
...
+    let function_call = FunctionCall {
+        contract_address: FieldElement::from_hex_be(contract_address)?,
+        entry_point_selector: get_selector_from_name(func_name)?,
```
cptartur, line 37: You can consider adding additional context to the errors so the resulting error is clearer when displayed to the user.

`.context("some message")?`

protostar-cli/src/starknet_commands/call.rs
```diff
@@ -0,0 +1,48 @@
...
+        calldata: calldata
+            .clone()
+            .unwrap()
```
cptartur, line 40: I'd avoid using unwrap in production code

cptartur: You can take a look at my calldata converting in #2106, or perhaps write some better code to do that.

protostar-cli/src/main.rs
```diff
@@ -94,6 +97,23 @@ async fn main() -> Result<()> {
...
+                &args.contract_address,
+                &args.func_name,
+                &args.calldata,
+                &provider,
+                &block_id,
+            )
+            .await?;
+
+            // todo (#2107): Normalize outputs in CLI
+            eprintln!("Call response: {:?}", result);
```
cptartur, line 113: Why are we printing to stderr? If it's for debugging you can use `dbg!` macro though it prints to stdout I believe.

protostar-cli/src/lib.rs
```diff
@@ -80,3 +82,15 @@ pub fn get_account<'a>(
 
     Ok(account)
 }
+
+pub fn get_block_id(value: &str) -> Result<BlockId> {
+    // todo: add more block ids (hash, number)
```
cptartur, line 87: Should we create an issue for this?

protostar-cli/src/lib.rs
```diff
@@ -80,3 +82,15 @@ pub fn get_account<'a>(
...
+    match value {
+        "pending" => Ok(BlockId::Tag(Pending)),
+        "latest" => Ok(BlockId::Tag(Latest)),
+        _ => Err(anyhow::anyhow!(
+            "No such block id {}! Possible values are pending and latest for now.",
+            value
```
cptartur, line 93: I think you can use `anyhow!` macro here.

protostar-cli/src/main.rs
war-in: I've consulted it with @THenry14 and we can leave it as is for now. There will be bigger issue to address all commands and their prints

protostar-cli/src/lib.rs
war-in: 50/50

protostar-cli/src/main.rs
THenry14: https://github.com/software-mansion/protostar/issues/2107

protostar-cli/src/starknet_commands/call.rs
```diff
@@ -0,0 +1,48 @@
...
+    #[clap(short = 'a', long = "contract-address")]
+    pub(crate) contract_address: String,
+
+    /// Name of the contract function to be called
+    #[clap(short = 'e', long = "func-name")]
+    pub(crate) func_name: String,
+
+    /// Arguments of the called function (list of hex)
+    #[clap(short = 'c', long = "calldata", num_args = 1.., value_delimiter = ' ')]
+    pub(crate) calldata: Option<Vec<String>>,
```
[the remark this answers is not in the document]
war-in: But without an `Option` `--calldata` flag will be required

protostar-cli/src/starknet_commands/call.rs
war-in: But your calldata is not Optional :/

protostar-cli/src/lib.rs
war-in: Am I not using it? :hmmm:

protostar-cli/src/starknet_commands/call.rs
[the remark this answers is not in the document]
cptartur: I thin it works without `Option`  and if you don't provide any it'll just use empty `Vec`. But you can verify because I'm not 100% sure.

cptartur (approved):
Looks okay, can be merged after open discussions are resolved.

protostar-cli/src/lib.rs
war-in: https://github.com/software-mansion/protostar/issues/2119

THenry14 (approved):
🔥

PR #2058, opened 2023-06-11
Add MVP for cli

THenry14:
### Checklist
- [x] Relevant issue is linked
- [ ] Docs updated/issue for docs created
- [ ] Added relevant tests

Closes https://github.com/software-mansion/protostar/issues/1948
Closes https://github.com/software-mansion/protostar/issues/2104
Closes https://github.com/software-mansion/protostar/issues/2118

PR #2122, opened 2023-06-23
[CLI] Add deploy command

### LINKED ISSUE #2055: Add deploy command
Add deploy command to protostar-cli.

The command should be documented and have tests.

war-in:
Closes #2055 

### Checklist
- [ ] Added relevant tests

### DIFF
--- a/cast/src/lib.rs
+++ b/cast/src/lib.rs
@@ -130,7 +130,7 @@ pub async fn wait_for_tx(provider: &JsonRpcClient<HttpTransport>, tx_hash: Field
         let receipt = provider
             .get_transaction_receipt(tx_hash)
             .await
-            .expect("Couldnt get transaction with hash: {tx_hash}");
+            .expect("Could not get transaction with hash: {tx_hash}");
 
         let status = match receipt {
             Receipt(receipt) => match receipt {
--- a/cast/src/lib.rs
+++ b/cast/src/lib.rs
@@ -1,6 +1,6 @@
 use anyhow::{anyhow, Context, Result};
 use camino::Utf8PathBuf;
-use console::{style};
+use console::style;
 use serde::{Deserialize, Serialize};
 use starknet::core::types::BlockTag::{Latest, Pending};
 use starknet::core::types::MaybePendingTransactionReceipt::{PendingReceipt, Receipt};
@@ -132,22 +132,23 @@ pub async fn wait_for_tx(provider: &JsonRpcClient<HttpTransport>, tx_hash: Field
             .await
             .expect("Could not get transaction with hash: {tx_hash}");
 
-        let status = match receipt {
-            Receipt(receipt) => match receipt {
+        let status = if let Receipt(receipt) = receipt {
+            match receipt {
                 Invoke(receipt) => receipt.status,
                 Declare(receipt) => receipt.status,
                 Deploy(receipt) => receipt.status,
                 DeployAccount(receipt) => receipt.status,
                 L1Handler(receipt) => receipt.status,
-            },
-            PendingReceipt(_) => continue 'a,
+            }
+        } else {
+            continue 'a;
         };
 
         match status {
             TransactionStatus::Pending => {
                 sleep(Duration::from_secs(5));
                 true
-            },
+            }
             TransactionStatus::AcceptedOnL2 | TransactionStatus::AcceptedOnL1 => false,
             TransactionStatus::Rejected => {
                 println!("{}", style("Transaction has been rejected").red());
--- a/cast/src/lib.rs
+++ b/cast/src/lib.rs
@@ -17,6 +17,8 @@ use starknet::{
 };
 use std::collections::HashMap;
 use std::fs;
+use std::thread::sleep;
+use std::time::Duration;
 use url::Url;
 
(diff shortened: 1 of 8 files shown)


cast/src/lib.rs
```diff
@@ -10,9 +9,18 @@ use starknet::{
     providers::jsonrpc::{HttpTransport, JsonRpcClient},
     signers::{LocalWallet, SigningKey},
 };
+use std::collections::HashMap;
 use std::fs;
 use url::Url;
 
+// Taken from starknet-rs
+pub const UDC_ADDRESS: FieldElement = FieldElement::from_mont([
```
THenry14, line 17: we could also take it from starkscan/starkware docs in hex representation and use it here, right? I guess it could be easier to track

cast/src/main.rs
```diff
@@ -76,13 +79,36 @@ async fn main() -> Result<()> {
             eprintln!("Transaction hash: {}", declared_contract.transaction_hash);
             Ok(())
         }
-        Commands::Call(args) => {
-            let block_id = get_block_id(&args.block_id).unwrap();
+        Commands::Deploy(deploy) => {
+            let account = get_account(&cli.account, &cli.accounts_file_path, &provider, &network)?;
```
THenry14, line 83: get_account returns `&mut account`, are your sure there are no issues with this?
we can change what it returns, as right now nothing actually uses mut (we can change it back in the future if needed)

cast/src/starknet_commands/deploy.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[clap(short, long)]
+    pub constructor_calldata: Vec<String>,
+
+    /// Salt for the address
+    #[clap(short, long)]
+    pub salt: Option<String>,
+
+    /// If false, salt will not be modified with account address
+    #[clap(short, long, default_value = "true")]
+    pub unique: String,
```
THenry14, line 32: why not `bool` ?

cast/src/starknet_commands/deploy.rs
```diff
@@ -0,0 +1,77 @@
...
+        None => FieldElement::from(OsRng.next_u32()),
+    };
+    let class_hash = FieldElement::from_hex_be(class_hash)?;
+    let unique = matches!(unique, "true");
+    let raw_constructor_calldata = constructor_calldata
+        .iter()
+        .map(|cd| {
+            FieldElement::from_hex_be(cd).context("Failed to convert calldata to FieldElement")
+        })
+        .collect::<Result<Vec<_>>>()?;
```
THenry14, line 58: I believe the constructor should be able to infer type if you do simething like:
```
let raw_constructor_calldata: Result<Vec<FieldElement>> = constructor_calldata ....
```

+ perhaps `.with_context` could be more desirable? https://github.com/software-mansion/protostar/pull/2058#discussion_r1238534875

cast/src/starknet_commands/deploy.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[clap(short, long)]
+    pub salt: Option<String>,
+
+    /// If false, salt will not be modified with account address
+    #[clap(short, long, default_value = "true")]
+    pub unique: String,
+
+    /// Max fee for the transaction. If not provided, max fee will be automatically estimated
+    #[clap(short, long)]
+    pub max_fee: Option<String>,
```
THenry14, line 36: maybe this could be u32/u64?

cast/src/starknet_commands/deploy.rs
war-in: > perhaps .with_context could be more desirable? https://github.com/software-mansion/protostar/pull/2058#discussion_r1238534875

Agree with that, but not in this particular case. We are not performing any costly string operations here.

cast/src/main.rs
war-in: I think there are no issues, but also I don't know if there can be some. Could you elaborate on such cases?

THenry14: nvm, account will no longer be mut https://github.com/software-mansion/protostar/pull/2126

cast/src/starknet_commands/declare.rs
```diff
@@ -47,5 +50,7 @@ pub async fn declare(
     let declaration = account.declare(Arc::new(contract_definition.flatten()?), casm_class_hash);
     let declared = declaration.send().await?;
 
+    wait_for_tx(account.provider(), declared.transaction_hash).await;
```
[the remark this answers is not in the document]
war-in: For now, it returns 
`Transaction has been rejected
Class hash: 233725228316200119834039379114988415372286690470331219113395524995289739146
Transaction hash: 2145090637473955348127825649177401711201765879895238539301684137989309824520`

I will add more descriptive error messages in the `normalize cli output PR`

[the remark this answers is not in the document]
war-in: I will add more descriptive error messages in the `normalize cli output` PR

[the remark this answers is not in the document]
THenry14: Sure, my issue is that I'm not sure we should return Ok, but it can be handled in time for normalize cli output

THenry14 (approved):

PR #2126, opened 2023-06-26
Improve cli args

### LINKED ISSUE #2117: Normalize inputs for commands
We should follow inputs formats as closely as possible with foundry.
Let's start with call, and then go command by command (this task can be split, or all commands can be done under it)

https://book.getfoundry.sh/reference/cast/cast-call#examples

karol-bisztyga:
### Checklist
- [ ] Added relevant tests


cast/src/lib.rs
```diff
@@ -78,7 +78,7 @@ pub fn get_account<'a>(
     ));
     let address = FieldElement::from_hex_be(&account_info.address)
         .with_context(|| format!("Failed to convert account address {} to FieldElement", &account_info.private_key))?;
-    let mut account = SingleOwnerAccount::new(provider, signer, address, network.get_chain_id());
+    let account = SingleOwnerAccount::new(provider, signer, address, network.get_chain_id());
```
THenry14, line 81: Please also change commands (they expect &mut account)

karol-bisztyga: From what I understand it doesn't matter for the outer code, the `mut` was redundant here because the function just returns an object, and whether it is mut or not, it depends on the outer declaration.

cast/src/starknet_commands/helpers.rs
```diff
@@ -0,0 +1,10 @@
+use starknet::core::types::FieldElement;
+use anyhow::Result;
+
+pub fn parse_number(contract_address_arg: &str) -> Result<FieldElement> {
```
war-in, line 4: Since this function is universal let's change the `contract_address_` prefix to something more suitable

war-in (approved):

cast/src/starknet_commands/helpers.rs
karol-bisztyga: [address fdbck](https://github.com/software-mansion/protostar/pull/2126/commits/292f09298a7bc2981fd4bf4a0a4c980ec2ab6bed)

cast/src/lib.rs
karol-bisztyga: can we resolve this?

THenry14: let's. but please mark this PR as draft for now, so it won't get merged by accident (we should still discuss changes to interface with Maks when he's back)

karol-bisztyga:
Converting to draft, we should discuss whether we should have this.

PR #2135, opened 2023-06-27
Normalize outputs in CLI

### LINKED ISSUE #2107: Normalize outputs in CLI

war-in:
Closes #2107 

### Checklist
- [ ] Added relevant tests


cast/src/lib.rs
```diff
@@ -149,11 +152,38 @@ pub async fn wait_for_tx(provider: &JsonRpcClient<HttpTransport>, tx_hash: Field
...
             }
-            TransactionStatus::AcceptedOnL2 | TransactionStatus::AcceptedOnL1 => false,
+            TransactionStatus::AcceptedOnL2 | TransactionStatus::AcceptedOnL1 => {
+                return Ok("Transaction accepted")
+            }
             TransactionStatus::Rejected => {
-                println!("{}", style("Transaction has been rejected").red());
-                false
+                return Err(anyhow!(
+                    "Transaction has been rejected. Check if max-fee argument is high enough"
```
THenry14, line 160: Is this most common reason? Maybe we can get a real reason somehow? If not, maybe we should remove this, so it is not misleading?

cast/src/lib.rs
```diff
@@ -149,11 +152,38 @@ pub async fn wait_for_tx(provider: &JsonRpcClient<HttpTransport>, tx_hash: Field
...
+                ));
             }
         }
     } {}
+
+    Err(anyhow!("Should not reach this line"))
```
THenry14, line 166: `Unexpected error happened` + error body ?

cast/src/lib.rs
```diff
@@ -149,11 +152,38 @@ pub async fn wait_for_tx(provider: &JsonRpcClient<HttpTransport>, tx_hash: Field
...
+        StarknetError::InvalidTransactionIndex => "There is no transaction with such an index",
+        StarknetError::ClassHashNotFound => "Provided class hash does not exist",
+        StarknetError::ContractError => "An error occurred in the called contract",
+        StarknetError::InvalidContractClass => {
+            "Contract class is invalid. Make sure to pass correct one"
+        }
+        StarknetError::ClassAlreadyDeclared => {
+            "Contract with the same class hash is already declared"
+        }
+        _ => "Unknown RPC error",
```
THenry14, line 187: I don't think we should be verbose in some cases (eg "Make sure to pass correct one", we can get rid of such things in my opinion). Error messages should be informative, but concise.

although, this is probably an open question, so we can discuss.

cast/src/main.rs
```diff
@@ -74,9 +73,12 @@ async fn main() -> Result<()> {
                 &mut account,
             )
             .await?;
-            // todo: #2107
-            eprintln!("Class hash: {}", declared_contract.class_hash);
-            eprintln!("Transaction hash: {}", declared_contract.transaction_hash);
+
+            println!("Class hash: {:#x}", declared_contract.class_hash);
```
THenry14, line 77: Have you thought about making it possible to allow for other formats to be returned? it could be a flag.

cast/src/main.rs
```diff
@@ -56,7 +56,6 @@ async fn main() -> Result<()> {
 
     // todo: #2052 take network from scarb config if flag not provided
     let network_name = cli.network.unwrap_or_else(|| {
-        // todo: #2107
```
THenry14, line 59: How about we added `--machine-readable` flag as well?

cast/src/starknet_commands/call.rs
```diff
@@ -44,7 +48,16 @@ pub async fn call(
...
 
-    Ok(res)
+    match res {
+        Ok(res) => Ok(res),
+        Err(error) => match error {
+            Other(RpcError(Code(error))) | StarknetError(error) => {
+                Err(anyhow!(get_rpc_error_message(error)))
+            }
+            Other(RpcError(Unknown(error))) => Err(anyhow!(error.message)),
+            _ => Err(anyhow!("Other RPC error")),
```
THenry14, line 60: Are there any codes not declared in our codebase, that were added to starknet-spec? 
Maybe we could change message to something like `"Unknown RPC error {}", code` ?

cast/src/starknet_commands/declare.rs
```diff
@@ -48,9 +52,18 @@ pub async fn declare(
...
+            Ok(_) => Ok(declared),
+            Err(message) => Err(anyhow!(message)),
+        },
+        Err(error) => match error {
+            Provider(Other(RpcError(Code(error))) | StarknetError(error)) => {
+                Err(anyhow!(get_rpc_error_message(error)))
+            }
+            Provider(Other(RpcError(Unknown(error)))) => Err(anyhow!(error.message)),
+            _ => Err(anyhow!("Other RPC error")),
+        },
```
THenry14, line 67: let's add this to lib (or new file errors.rs) and reuse in other commands

cast/src/starknet_commands/deploy.rs
```diff
@@ -65,16 +70,34 @@ pub async fn deploy(
...
+                            udc_contract_address: FieldElement::from_hex_be(UDC_ADDRESS)?,
+                        })
+                    } else {
+                        NotUnique
+                    },
+                    &raw_constructor_calldata,
+                ),
+            )),
+            Err(message) => Err(anyhow!(message)),
+        },
```
THenry14, line 94: I wonder if we could also abstract this snippet, to be a reusable function for our commands. what do you think?

cast/src/lib.rs
war-in: Unfortunately, we can't :/ I would really like to get it from the net but it is not possible in the RPC. I can remove it

cast/src/main.rs
THenry14: yes, this is one possibility and perhaps mvp scenario. we could take a look at foundry/scarb/come other tools to see how it is handled there.

can be done under separate issue in different PR if you wish

war-in: https://github.com/software-mansion/protostar/issues/2146 I would take it just after this PR

cast/src/main.rs
war-in: **https://github.com/software-mansion/protostar/pull/2135/commits/869286f056d40d2daf71aa17008694609e55503b**

cast/src/main.rs
```diff
@@ -31,6 +31,10 @@ struct Cli {
     )]
     accounts_file_path: Utf8PathBuf,
 
+    /// If passed, values will be displayed as integers, otherwise as hexes
+    #[clap(short, long)]
+    int_format: bool,
```
THenry14, line 36: it's fine I guess, but maybe `--format int | hex` would be a bit better? the default would be hex

cast/src/main.rs
```diff
@@ -97,9 +104,8 @@ async fn main() -> Result<()> {
             )
             .await?;
 
-            // todo: #2107
-            eprintln!("Contract address: {contract_address}");
-            eprintln!("Transaction hash: {transaction_hash}");
+            print_formatted("Contract address: ", contract_address, cli.int_format);
```
THenry14, line 107: eprintln! redirects to stderr, where println! to stdout. I think we want to preserve outputting to stderr here.

THenry14 (approved):

PR #2120, opened 2023-06-22
Multiple contracts in package

cptartur:
Closes #1847 
Closes #2081 

### Checklist
- [x] Relevant issue is linked
- [ ] Docs updated/issue for docs created
- [x] Added relevant tests

Issue #2113, opened 2023-06-22
Default arguments from scarb.toml


THenry14:
As discussed with @war-in , let's simplify ui/ux further, and try to fetch default values for things like rpc_url from scarb.toml.
That would imply, that we no longer need to verify network with provider (as per todo 2113 in code), as the proposed template would be:

```
[tool.protostar.cli.mainnet]
...

[tool.protostar.cli.devnet]
...
```

Possibly to do after mvp for scarb integration

PR #2138, opened 2023-06-27
Build with scarb for declare

### LINKED ISSUE #2052: Scarb integration with protostar-cli
protostar-cli should use scarb to build contracts and then get file maps to sierra and casm compiled files.
It should also parse Scarb.toml for all the necessary things like getting info about contracts, file paths etc.

karol-bisztyga:


cast/src/main.rs
```diff
@@ -53,6 +54,10 @@ enum Commands {
 #[tokio::main]
 async fn main() -> Result<()> {
     let cli = Cli::parse();
+    let accounts_file_path = match cli.accounts_file_path {
+        Some(path) => path,
+        None => Utf8PathBuf::from(format!("{}/.starknet_accounts/starknet_open_zeppelin_accounts.json", dirs::home_dir().expect("failed to read the home directory").to_str().expect("failed to convert home directory to string"))),
+    };
 
     // todo: #2052 take network from scarb config if flag not provided
```
THenry14, line 62: 👀

cast/src/main.rs
```diff
@@ -65,12 +70,15 @@ async fn main() -> Result<()> {
 
     match cli.command {
         Commands::Declare(declare) => {
+            let parent = accounts_file_path.parent().unwrap();
+            if !accounts_file_path.exists() {
+                anyhow::bail!("accounts file ({}) does not exist", parent);
+            }
```
THenry14, line 76: Maybe this check should be applicable to other commands as well, and moved out of `match`?

cast/src/starknet_commands/declare.rs
```diff
@@ -14,36 +13,64 @@ use starknet::{
...
-    casm_contract_path: &Utf8PathBuf,
+    contract: &str,
     max_fee: Option<u128>,
     account: &mut SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet>,
 ) -> Result<DeclareTransactionResult> {
+    let _ = Command::new("scarb")
+        .current_dir(std::env::current_dir().expect("failed to obtain current dir"))
+        .arg("build")
+        .output()
+        .expect("Failed to build contracts with Scarb");
```
THenry14, line 38: how does this work, it builds all the contracts in current directory?

THenry14, line 38: Let's be consistent with error messages - eg. they all should start with capital letter (applicable to other lines/files as well)

cast/src/starknet_commands/declare.rs
```diff
@@ -14,36 +13,64 @@ use starknet::{
...
+
+    let current_dir = std::env::current_dir().expect("failed to obtain current dir");
+    let paths = std::fs::read_dir(format!("{}/target/dev", current_dir.to_str().unwrap()))
+        .expect("failed to read the file maybe build failed");
```
THenry14, line 42: what are possible issues here? I can think of a scenario that files does not exist, because a build failed (and probably we should not reach this point in that case). Is there any other case? if not, let's maybe adjust the message accordingly.

cast/src/starknet_commands/declare.rs
```diff
@@ -14,36 +13,64 @@ use starknet::{
...
+            if path_str.contains(".sierra.json") {
+                maybe_sierra_contract_path = Some(path_str);
+            } else if path_str.contains(".casm.json") {
+                maybe_casm_contract_path = Some(path_str);
+            }
+        }
+    }
+
+    let sierra_contract_path = maybe_sierra_contract_path.expect(&format!("no sierra found for contract: {}", contract)[..]);
+    let casm_contract_path = maybe_casm_contract_path.expect(&format!("no casm found for contract: {}", contract)[..]);
```
THenry14, line 63: Snce we will declare one contract at a time, I wonder if it wouldn't be easier to just try to check if specific files exist, eg in pseudocode:

```
    let file_path = Path::new("{}.sierra.json", &contract);

    if file_path.exists() {
        read
        ...
    } else {
        println!("File does not exist!");
        ...
    }
```

What do you think?

cast/tests/declare_test/Scarb.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
THenry14, line 1: Let's move this to conform to: https://github.com/software-mansion/protostar/pull/2140

cast/src/main.rs
THenry14: + any other ideas on what we can extract from scarb config?

cast/tests/declare_test/Scarb.toml
THenry14: btw, this is planned structure for declare tests: https://github.com/software-mansion/protostar/commit/166c3191d242a2e48d72bb370078fd668180f90e

cast/src/main.rs
karol-bisztyga: > 👀

Sorry, could you clarify what's wrong? This emoji doesn't tell me what you mean :(

THenry14: sure, sorry :D there is a todo for the issue this PR addresses, so I guess it should be resolved in this PR

cast/src/main.rs
karol-bisztyga: yes, good point, thx!

cast/src/starknet_commands/declare.rs
karol-bisztyga: yes

karol-bisztyga: sure

cast/src/starknet_commands/declare.rs
karol-bisztyga: I encountered weird behaviors with the scarb build command (it should break the execution but sometimes it didn't)., so I thought a hint would be ok here. If you have any specific suggestions, I'd appreciate it if you provided them. 

> what are possible issues here? I can think of a scenario that files does not exist, because a build failed (..). Is there any other case?

I don't see any other case.

cast/src/starknet_commands/declare.rs
karol-bisztyga: the problem is, scarb outputs files like this:
```
declare_test_HelloStarknet.sierra.json
```
The point is, I wanted to avoid assuming the exact name of the output files.

cast/tests/declare_test/Scarb.toml
karol-bisztyga: Ok I will remove the tests.

cast/src/main.rs
karol-bisztyga: Right, sorry, I did not notice, thanks!

karol-bisztyga: Yes, I see now, didn't notice, thx

cast/src/starknet_commands/declare.rs
THenry14: > it should break the execution but sometimes it didn't
this does not sound good. do you have some idea on what could be the cause? Ideally this is something that should be deterministic, and we should always end execution when such scenario happens.

cast/src/starknet_commands/declare.rs
THenry14: btw, is it possible to compile only one contract (the one we want to declare) ?

cast/src/starknet_commands/declare.rs
THenry14: hmm, I see. 
If I remember correctly, scarb should be able to return a map of compiled artifacts, so maybe we could use this here instead?

karol-bisztyga: I'll take a look 👀

cast/src/main.rs
karol-bisztyga: [remove resolved todo](https://github.com/software-mansion/protostar/pull/2138/commits/05adf8aabd338e7d877f767f22285de4df2771cc)

cast/src/main.rs
karol-bisztyga: [move account file check outside of match case](https://github.com/software-mansion/protostar/pull/2138/commits/f2c1dc2f40a37b6a7b6bc75942ee051d84aac23e)

cast/src/starknet_commands/declare.rs
karol-bisztyga: [start err msgs from capital letter](https://github.com/software-mansion/protostar/pull/2138/commits/c9efb8e2fbf9fd293ef7c15c76c24fba16c81c38)

piotmag769: > btw, is it possible to compile only one contract (the one we want to declare) ?
I don't think so, but let's ask @maciektr

maciektr: @THenry14 @piotmag769 
It builds all contracts in Scarb project indicated by current directory. 
As of now, it is not possible to indicate which contracts to build.

THenry14: cool, thanks for clarification.
in that case, does it make sense to also expect an error, where we can't find Scarb.toml in current or parent directories?

cast/src/starknet_commands/declare.rs
piotmag769: cc @maciektr @mkaput

cast/src/starknet_commands/declare.rs
karol-bisztyga: Scarb will raise an error if there's no `Scarb.toml` I think

cast/src/starknet_commands/declare.rs
piotmag769: You could reuse the code from https://github.com/software-mansion/protostar/pull/2120
I guess it is related to https://github.com/software-mansion/protostar/issues/2141 - it would have to be done after merging these two PRs

cast/src/starknet_commands/declare.rs
karol-bisztyga: Why are you cc'ing them for? It is for me to confirm/check this again.

cast/src/starknet_commands/declare.rs
```diff
@@ -14,36 +13,64 @@ use starknet::{
...
+            if path_str.contains(".sierra.json") {
+                maybe_sierra_contract_path = Some(path_str);
+            } else if path_str.contains(".casm.json") {
+                maybe_casm_contract_path = Some(path_str);
+            }
+        }
+    }
+
+    let sierra_contract_path = maybe_sierra_contract_path.expect(&format!("No sierra found for contract: {}", contract)[..]);
+    let casm_contract_path = maybe_casm_contract_path.expect(&format!("No casm found for contract: {}", contract)[..]);
```
piotmag769, line 63: Can we please make it consistent with https://github.com/software-mansion/protostar/pull/2120?

cast/src/main.rs
```diff
@@ -53,8 +54,14 @@ enum Commands {
 #[tokio::main]
 async fn main() -> Result<()> {
     let cli = Cli::parse();
+    let accounts_file_path = match cli.accounts_file_path {
+        Some(path) => path,
+        None => Utf8PathBuf::from(format!("{}/.starknet_accounts/starknet_open_zeppelin_accounts.json", dirs::home_dir().expect("failed to read the home directory").to_str().expect("failed to convert home directory to string"))),
```
piotmag769, line 59: Why do we use this file if no account file is provided? I don't think we should expect this file in users home dir at all. We should match it later on if and only if the command being executes needs this argument to succeed -in this case the command should fail if the argument is not found.

piotmag769 (changes requested):

cast/src/starknet_commands/declare.rs
maciektr: If you have a case where build fails with exit code 0, please let us know :D

cast/src/main.rs
THenry14: this is a sane default. if there's no flag and this file does not exist, we should fail. but I'd leave this file as default, that's just good ux and nothing out of the blue in my opinion

cast/src/starknet_commands/declare.rs
piotmag769: > What are you cc'ing them for? It is for me to confirm/check this again.
To make sure we report the bug

karol-bisztyga: It's not the scarb team's responsibility that we report bugs to them, still mine, no need to bother them.

cast/src/main.rs
piotmag769: I don't really think it is a good UX. 

First of all, the file we are expecting here is created when using `cairo-lang` afaik (but I may be wrong), which will be deprecated. 

Secondly, we are expecting a file in a home directory which is a little bit weird imo. 

Moreover, if this file does not exist a user will get an error message informing about the missing file. It does not inform about the actual problem - not providing the required argument. 

Last but not least, a user could just forget to provide 

If we want to have a default argument for the account file, I think we should:
a) include it in in the struct we use for args parsing - this way it is readable and won't be lost/forgotten inside the codebase + it will be displayed in the help message for the cli
b) document it in docs - I guess that will be done automatically using `clap`, so point a should satisfy it

cast/src/starknet_commands/declare.rs
karol-bisztyga: [fix executing scarb command](https://github.com/software-mansion/protostar/pull/2138/commits/1c5b95973e1dec69b7a55b03f3a8f03a940c18d1)

karol-bisztyga: turned out I used Rust's `Command` incorrectly. The `expect` there is for starting a command but we should also check the result's status code.

piotmag769 (commented):
Is it possible to replace the `expect`s with `context("abc")?` ?

cast/src/starknet_commands/declare.rs
karol-bisztyga: Could you specify in what way, please?

cast/src/main.rs
THenry14: > First of all, the file we are expecting here is created when using cairo-lang afaik (but I may be wrong), which will be deprecated.

the file is for openzeppelin accounts, we will also provide a command to create such account in the future

> Secondly, we are expecting a file in a home directory which is a little bit weird imo.

this is the default path where it's being created, it's not up to us

> Moreover, if this file does not exist a user will get an error message informing about the missing file. It does not inform about the actual problem - not providing the required argument.

Here I agree. 

> Last but not least, a user could just forget to provide this arg and create an account they did not intend to create by accident.

I don't think it's possible.

> If we want to have a default argument for the account file, I think we should:
a) include it in in the struct we use for args parsing - this way it is readable and won't be lost/forgotten inside the codebase + it will be displayed in the help message for the cli,
b) document it in docs - I guess that will be done automatically using clap, so point a should satisfy it.

this PR changes exactly that behavior. I don't have strong feelings about this, but if you believe it should not be moved from args parsing, let's revert to that.

> Imo we should remove it entirely though.

strongly disagree

piotmag769: > this PR changes exactly that behavior. I don't have strong feelings about this, but if you believe it should not be moved from args parsing, let's revert to that.

I insist that we do that - otherwise we will just forget about it + the automatic cli help generated by clap will not include that info.

karol-bisztyga: Yes, so the problem was `~` which was not read correctly, we need something like `/Users/root/home`, where `root` is your user name and it varies so this cannot be hard coded. Therefore, we need to obtain the home directory in the runtime and this could not be done in `#[clap]`'s `default_value`, which probably needs to be specified in the compile time (but I may be wrong).

THenry14: That's weird, it was working for me and Marcin. let's try to debug offline

cast/src/starknet_commands/declare.rs
piotmag769: Basically I would use code from https://github.com/software-mansion/protostar/pull/2120/files#diff-e94fd807a4275d652e6b7372b7c207fd996d38ae3250db9eb18b1b63cc511472 - scarb artifact map is already used there and I think it is more readable. If you wish so you can create another crate with this code as described in https://github.com/software-mansion/protostar/issues/2141

karol-bisztyga: If there's a task for it, I think we can address this separately. I'm going to add a note with the issue number.

cast/src/main.rs
karol-bisztyga: discussed offline, we're about to address this separately. [revert change with default path](https://github.com/software-mansion/protostar/pull/2138/commits/1239607d6eeedf59787c186728beec19d5d1a262)

cast/src/starknet_commands/declare.rs
karol-bisztyga: [add todo issue for starknet artifacts](https://github.com/software-mansion/protostar/pull/2138/commits/d1af2faad1eaaf2fa5cc584dc9e42e95790a08df)

cast/src/starknet_commands/declare.rs
karol-bisztyga: the same as in https://github.com/software-mansion/protostar/pull/2138#discussion_r1246244782: let's address this separately (issue already linked in the code).

cast/tests/declare_test/Scarb.toml
karol-bisztyga: [remove tests](https://github.com/software-mansion/protostar/pull/2138/commits/38e8c6d78bf31316cb103b03da1d19389cf70db3)

cast/src/starknet_commands/declare.rs
THenry14: sorry, but unless you plan to take on it straight away (or already doing this) then I'd rather not. 

I see this part as a crucial element of scarb integration to be honest, and without using those maps it seems like we have half-baked solution, that has to be re-done from scratch basically.

If you wish you can do it in another PR and set this branch as target, but I don't think this would make things easier.

🙏

cast/src/main.rs
```diff
@@ -53,8 +53,11 @@ enum Commands {
 #[tokio::main]
 async fn main() -> Result<()> {
     let cli = Cli::parse();
+    // TODO #2147
+    if !cli.accounts_file_path.exists() {
+        anyhow::bail!("accounts file ({}) does not exist", cli.accounts_file_path);
+    }
```
THenry14, line 59: let's comment it out or remove for now, we'll re-add it in #2147

cast/src/main.rs
```diff
@@ -53,8 +53,11 @@ enum Commands {
...
 
-    // todo: #2052 take network from scarb config if flag not provided
```
THenry14, line 57: can you point me to where it's done? sorry, I can't find it

THenry14:
> let's target this branch to master. Changes not related to interface change from `inprove-cli-args` can be moved in here.

Please also rebase to/merge master

cast/src/main.rs
karol-bisztyga: [fdbck](https://github.com/software-mansion/protostar/pull/2138/commits/c18edf2e55891fc1040b8fa3fc3614efa658116b)

cast/src/main.rs
karol-bisztyga: [fdbck](https://github.com/software-mansion/protostar/pull/2138/commits/c18edf2e55891fc1040b8fa3fc3614efa658116b)

cast/src/main.rs
THenry14: cool, but we should rather add this functionality in this PR, so let's lleave unresolved until it's done :+1:

cast/src/starknet_commands/declare.rs
karol-bisztyga: [take sierra and casm from starknet artifacts](https://github.com/software-mansion/protostar/pull/2138/commits/038d08f2a5fb77cd0fa3baf100a4472e4bbeb231)

cast/src/starknet_commands/declare.rs
```diff
@@ -14,41 +13,110 @@ use starknet::{
...
-    pub casm_contract_path: Utf8PathBuf,
+    /// contract name
+    pub contract: String,
 
     /// Max fee for the transaction. If not provided, max fee will be automatically estimated
     #[clap(short, long)]
     pub max_fee: Option<u128>,
 }
 
 pub async fn declare(
```
piotmag769, line 54: Is it possible to get rid of panicking in this function and bubble errors instead?

cast/src/starknet_commands/declare.rs
```diff
@@ -14,41 +13,110 @@ use starknet::{
...
 ) -> Result<DeclareTransactionResult> {
+    let contract_name: String = contract_name.to_string();
+    let command_result = Command::new("scarb")
+        .current_dir(std::env::current_dir().context("Failed to obtain current dir")?)
+        .arg("build")
+        .output()
+        .context("Failed to start building contracts with Scarb")?;
+    let result_code = command_result.status.code().context("failed to obtain status code from scarb build")?;
+    if result_code != 0 {
+        panic!("scarb build returned non-zero exit code: {}", result_code);
```
maciektr, line 67: Consider returning an error (`bail`) instead of `panic` for non-invariants.

cast/src/starknet_commands/declare.rs
```diff
@@ -14,41 +13,110 @@ use starknet::{
...
+    }
+
+    // TODO #2141 improve handling starknet artifacts
+    let current_dir = std::env::current_dir()
+                .expect("Failed to get current directory")
+                .join("target/dev");
```
maciektr, line 73: Consider using `scarb manifest-path` instead of current dir as base path. 

For instance, in project structured as following:
```
-> project
    -> src
        -> lib.cairo
    -> Scarb.toml
```
`scarb build` run should exit successfully, producing `project/target` subdirectory.

cast/src/starknet_commands/declare.rs
maciektr: Is Scarb output passed to the UI? Especially in case of failure?
Consider 
```
.stdout(Stdio::piped()).stderr(Stdio::piped())
```

cast/src/main.rs
```diff
@@ -53,7 +54,6 @@ enum Commands {
 #[tokio::main]
 async fn main() -> Result<()> {
     let cli = Cli::parse();
-
     // todo: #2052 take network from scarb config if flag not provided
     let network_name = cli.network.unwrap_or_else(|| {
         // todo: #2107
```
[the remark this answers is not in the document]
karol-bisztyga: [pull network from scarb config](https://github.com/software-mansion/protostar/pull/2138/commits/a0f6a1e9f013d57e1a486ff8321ff5d23562d585)

cast/src/main.rs
karol-bisztyga: [pull network from scarb config](https://github.com/software-mansion/protostar/pull/2138/commits/a0f6a1e9f013d57e1a486ff8321ff5d23562d585)

cast/src/starknet_commands/declare.rs
karol-bisztyga: ok im gonna replace some of them but they do not work right away in with_context, i dont think this is a top prio, we can focus on that in the task for reusing the artifacts code.

cast/src/starknet_commands/declare.rs
karol-bisztyga: Yes, this can be done but in such a case, we'd have to either pass this from the main or obtain it once again using scarb metadata (that uses `current_dir` anyway, so there's no difference after all). If you know how to solve this without using `current_dir` at some point, please let me know.

cast/src/main.rs
```diff
@@ -53,13 +56,36 @@ enum Commands {
...
+                        maybe_network_name = Some(network_str.to_string());
+                    }
+                }
+            }
+        }
+    }
+    if maybe_network_name.is_none() {
         // todo: #2107
         eprintln!("{}", style("No --network flag passed!").red());
         std::process::exit(1);
```
maciektr, line 86: Consider moving to type system
https://github.com/software-mansion/scarb/blob/6484cf7b60e166475becc175e6a66e6d95dff6e8/scarb/src/bin/scarb/errors.rs#L7

cast/src/starknet_commands/declare.rs
maciektr: There is a command `scarb manifest-path`, which offloads Scarb manifest search to Scarb.

cast/src/starknet_commands/declare.rs
karol-bisztyga: [address feedback](https://github.com/software-mansion/protostar/pull/2138/commits/1fbe96f5d95497671ce5af742d0fc2b906d5f63d)

cast/src/main.rs
karol-bisztyga: 1. this is not a change introduced in this PR, but ok we can change this.
2. what do u mean? you want to raise a scarb error instead of returning process code 1 or something else? i didnt understand.

cast/src/starknet_commands/helpers.rs
```diff
@@ -0,0 +1,10 @@
+use starknet::core::types::FieldElement;
```
THenry14, line 1: let's move this to lib.rs, as in this directory we only store starknet commands, as a convention

cast/src/main.rs
```diff
@@ -57,12 +60,36 @@ enum Commands {
...
+        let package = &metadata.packages[0];
+        if let Some(tool) = &package.manifest_metadata.tool {
+            if let Some(protostar_tool) = tool.get("protostar") {
+                if let Some(network) = protostar_tool.get("network") {
+                    if let Some(network_str) = network.as_str() {
+                        maybe_network_name = Some(network_str.to_string());
+                    }
+                }
+            }
+        }
```
THenry14, line 85: Maybe something like:

```
if let Some(network_str) = package.manifest_metadata.tool
    .and_then(|tool| tool.get("protostar"))
    .and_then(|protostar_tool| protostar_tool.get("network"))
    .and_then(|network| network.as_str())
{
    maybe_network_name = Some(network_str.to_string());
}
```

?

maciektr: With this you can remove mut 
```
    let maybe_network_name = if let Some(network_name) = cli.network {
        Some(network_name)
    } else {
        let metadata = scarb_metadata::MetadataCommand::new()
            .inherit_stderr()
            .current_dir(current_dir().expect("failed to read current directory"))
            .no_deps()
            .exec()
            .unwrap();
        if metadata.packages.len() != 1 {
            anyhow::bail!("invalid number of packages obtained from scarb metadata");
        }
        let package = &metadata.packages[0];
        package
            .manifest_metadata
            .tool
            .as_ref()
            .and_then(|t| t.get("protostar"))
            .and_then(|t| t.get("network"))
            .and_then(|t| t.as_str())
            .map(|t| t.to_string())
    };
```

cast/src/starknet_commands/declare.rs
```diff
@@ -15,41 +15,112 @@ use starknet::{
...
+    package_name: String,
+    contract_name: String,
+    artifacts: ScarbStarknetContractArtifact,
+}
+
+#[allow(dead_code)]
+#[derive(Deserialize)]
+struct ScarbStarknetContractArtifact {
+    sierra: PathBuf,
+    casm: Option<PathBuf>,
```
maciektr, line 42: Generally speaking, `sierra` is also optional.

cast/src/main.rs
THenry14: Sorry for a sudden change, but we can remove it completely. We have discussed with @war-in how we can improve ux for cli, and decided that we'll remove `--network` flag altogether + use different schema in scarb.toml.
Because of this, the #2113 has been redefined, and fetching things will be done under this task, to speed up this PR merge.

cast/src/main.rs
THenry14: Irrelevant as per #2113, I will remove this altogether under this task, but thanks for the tip Maciek, I will remember that! :)

THenry14 (commented):
looking good, last few things to resolve and we're cooking 🔥

cast/src/starknet_commands/call.rs
```diff
@@ -6,15 +6,15 @@ use starknet::core::utils::get_selector_from_name;
...
+
 #[derive(Args)]
 #[command(about = "Call a contract instance on Starknet", long_about = None)]
 pub struct Call {
     /// Address of the called contract (hex)
-    #[clap(short = 'a', long = "contract-address")]
     pub contract_address: String,
 
     /// Name of the contract function to be called
-    #[clap(short = 'f', long = "function-name")]
```
war-in, line 17: Hmm, haven't we decided not to remove those clap configs? 🤔

cast/src/starknet_commands/declare.rs
```diff
@@ -15,41 +15,114 @@ use starknet::{
...
+#[derive(Deserialize)]
+struct ScarbStarknetContract {
+    id: String,
+    package_name: String,
+    contract_name: String,
+    artifacts: ScarbStarknetContractArtifact,
+}
+
+#[allow(dead_code)]
+#[derive(Deserialize)]
```
war-in, line 39: I think those structs should be moved to some new file. wdyt @THenry14 ?

THenry14: we can leave them here for now, as they are directly connected to declare. there is a possibility we'll move them in #2113

cast/src/main.rs
piotmag769: I guess it would be good to wrap the errors in a custom error that makes the program exit with non-zero exit code in the future

piotmag769 (approved):

cast/src/starknet_commands/declare.rs
piotmag769: I think we will never have a case where we want `sierra=false` in starknet target. I added a description for better error handling in case a user makes changes to Scarb.toml in https://github.com/software-mansion/protostar/issues/2141

cast/src/starknet_commands/declare.rs
karol-bisztyga: Discussed offline, deferring to https://github.com/software-mansion/protostar/issues/2154

cast/src/main.rs
karol-bisztyga: I'm going to leave it as is and you can remove it in #2113. I don't see a point in removing it here - removing it is part of task #2113.

cast/src/starknet_commands/helpers.rs
karol-bisztyga: [move parse number to lib](https://github.com/software-mansion/protostar/pull/2138/commits/cc68f5b024e3e6ae3311446387db32ef6431f36b)

cast/src/main.rs
```diff
@@ -24,6 +26,7 @@ struct Cli {
     account: String,
 
     /// Path to the file holding accounts info, defaults to ~/.starknet_accounts/starknet_open_zeppelin_accounts.json
+    // TODO #2147
```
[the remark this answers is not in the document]
karol-bisztyga: [remove todo 2147](https://github.com/software-mansion/protostar/pull/2138/commits/76be159f129cf80ea80fda7c66e77fa4170a20a0)

cast/src/starknet_commands/declare.rs
karol-bisztyga: This change is taken from https://github.com/software-mansion/protostar/pull/2120/files#diff-e94fd807a4275d652e6b7372b7c207fd996d38ae3250db9eb18b1b63cc511472R104-R107

1. You should've noticed that in https://github.com/software-mansion/protostar/pull/2120 (I see you didn't do a review but that's not my fault after all).
2. We should have a consistent code base. Therefore if I do this change here, it will not be consistent. If I decide to go through the code and do this change also in `protostar-rust/src/cheatcodes_hint_processor.rs`, I will have to consider if that is a good change, debug stuff, etc. So I do not think it should be done here.

I think we should leave it as is. If you feel this is very bad, feel free to create an issue for this, thanks.

cast/src/starknet_commands/declare.rs
THenry14: Can you give us tldr here or under issue? we (Marcin and I) did not participate in this discussion, and have no idea what was the result of it, where one of us will probably have to handle this

cast/src/main.rs
THenry14: please remove it. I'd rather re-add it from scratch than figure out what should and what should not be done

cast/src/starknet_commands/call.rs
THenry14: I don't see how is it related? this change is not on master, so this PR is the one that makes the change right? if so, please revert this change as it should be discussed before it is merged

karol-bisztyga: as https://github.com/software-mansion/protostar/issues/2117 is closed we can revert [revert claps](https://github.com/software-mansion/protostar/pull/2138/commits/25015334ead2b43450aaa8b86c10061795a508a6)

war-in (approved):

THenry14 (approved):
great job mate 🔥

cast/src/main.rs
karol-bisztyga: [revert network changes](https://github.com/software-mansion/protostar/pull/2138/commits/d1be50459470f663e30651fe860ea18034360df8)

PR #2143, opened 2023-06-28
Added exit first to the rust runner

### LINKED ISSUE #2003: Support `--exit-first`
Support exiting on first failed tests.

Radinyn:
### Checklist
- [x] Added relevant tests


protostar-rust/src/lib.rs
```diff
@@ -163,12 +181,20 @@ fn run_tests(tests: TestsFromFile, tests_stats: &mut TestsStats) -> Result<()> {
...
-    for config in tests.tests_configs {
-        let result = run_from_test_config(&mut runner, &config)?;
+    for (i, config) in tests.tests_configs.iter().enumerate() {
+        let result = run_from_test_config(&mut runner, config)?;
+
         tests_stats.update(&result.value);
         pretty_printing::print_test_result(&config.name.clone(), &result.value);
+
+        if let RunResultValue::Panic(_) = result.value {
+            if runner_config.exit_first {
```
piotmag769, line 191: [nit]
```suggestion
        if runner_config.exit_first {
            if  let RunResultValue::Panic(_) = result.value {
```

piotmag769 (approved):

protostar-rust/src/pretty_printing.rs
```diff
@@ -24,12 +24,21 @@ pub fn print_running_tests(test_file: &Utf8PathBuf, tests_num: usize) {
 }
 
 pub fn print_test_summary(tests_stats: TestsStats) {
-    println!(
-        "{}: {} passed, {} failed",
+    let skipped_str = if tests_stats.skipped > 0 {
+        format!(", {} skipped", tests_stats.skipped)
+    } else {
+        String::new()
+    };
```
piotmag769, line 31: Can we please print it every time? We are trying to be consistent with Cargo behaviour. Please remember to change tests accordingly.

cptartur (changes requested):

protostar-rust/src/lib.rs
```diff
@@ -146,15 +151,28 @@ pub fn run(
...
-        run_tests(tests_from_file, &mut tests_stats)?;
+        match runner_status {
+            RunnerStatus::Run => {
+                runner_status = run_tests(tests_from_file, &mut tests_stats, runner_config)?;
+            }
+            RunnerStatus::Exit => {
+                tests_stats.skipped += tests_from_file.tests_configs.len();
+            }
+        }
     }
```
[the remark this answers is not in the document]
Radinyn: I've proposed another solution, not sure if it's better though.

protostar-rust/src/lib.rs
```diff
@@ -146,15 +153,29 @@ pub fn run(
...
+    for tests_from_file in tests_iterator.by_ref() {
+        if run_tests(tests_from_file, &mut tests_stats, runner_config)? == RunnerStatus::TestFailed
+        {
+            break;
+        }
     }
+
+    for tests_from_file in tests_iterator {
+        tests_stats.skipped += tests_from_file.tests_configs.len();
+    }
```
cptartur, line 167: This could probably be replaced with single `reduce`, but I'm fine with either.

cptartur, line 167: I think it's safer that the previous approach with a flag.

@mkaput can you take a look at this and compare it with the version from this commit? https://github.com/software-mansion/protostar/pull/2143/commits/71764779af88b914247968fd8b0bcd6e4106c1a9

Or perhaps you have an idea for alternative solution?

mkaput: it's fine

cptartur (approved):

Issue #2146, opened 2023-06-29
`--machine-readable` flag for CLI


war-in:
Currently, CLI shows the output as a string. It would be nice to have it displayed in a JSON format

Arcticae:
We should consider doing it as `NDJSON`

PR #2150, opened 2023-06-29
[CLI] Verify network with provider

### LINKED ISSUE #2113: Default arguments from scarb.toml

THenry14:
### Checklist
- [X] Added relevant tests

Closes https://github.com/software-mansion/protostar/issues/2113


cast/src/lib.rs
```diff
@@ -66,9 +67,17 @@ impl Network {
...
-pub fn get_provider(url: &str) -> Result<JsonRpcClient<HttpTransport>> {
+pub fn get_provider(url: &str, network: &Network) -> Result<JsonRpcClient<HttpTransport>> {
     let parsed_url = Url::parse(url)?;
     let provider = JsonRpcClient::new(HttpTransport::new(parsed_url));
+
+    let rt = Runtime::new()?;
+    let provider_chain_id = rt.block_on(provider.chain_id())?;
+    let cli_chain_id = network.get_chain_id();
+    if provider_chain_id != cli_chain_id {
+        bail!("Networks mismatch: requested network is different than provider network!")
```
war-in, line 78: Should we migrate all of `Err(anyhow!(...))` to `bail!`?

cast/src/lib.rs
```diff
@@ -66,9 +67,17 @@ impl Network {
...
 }
 
-pub fn get_provider(url: &str) -> Result<JsonRpcClient<HttpTransport>> {
+pub fn get_provider(url: &str, network: &Network) -> Result<JsonRpcClient<HttpTransport>> {
     let parsed_url = Url::parse(url)?;
     let provider = JsonRpcClient::new(HttpTransport::new(parsed_url));
+
+    let rt = Runtime::new()?;
+    let provider_chain_id = rt.block_on(provider.chain_id())?;
+    let cli_chain_id = network.get_chain_id();
```
war-in, line 76: I've started wondering if we could get rid of the `network` parameter entirely.
Since we are calling network for its chain_id here we do not need `network` parameter anymore.
It's useful only when getting an account but we could change the logic of accessing the network's value.
What do you think?

cast/tests/e2e/main.rs
```diff
@@ -1 +1,2 @@
+mod helper_functions;
 pub(crate) mod helpers;
```
war-in, line 2: I think we should create a command to invoke all tests at once

cast/src/lib.rs
war-in: I think UX would be improved because it's one parameter less :)

cast/src/lib.rs
THenry14: let's. do you want me to handle this, or do you want to do this?

cast/src/lib.rs
THenry14: hmm, makes sense. I guess we should still get network from scarb.toml though, right? in order to compare the networks and throw an error when there is mismatch, to avoid accidental declare/deploy to wrong network?

cast/tests/e2e/main.rs
THenry14: can you clarify this, please? I'm not sure I follow

war-in: My bad. I forgot there is `cargo test` 🤦

cast/src/lib.rs
war-in: I will in a separate PR after #2146 is merged

cast/tests/e2e/helper_functions.rs
```diff
@@ -0,0 +1,128 @@
+use crate::helpers::devnet::URL;
```
karol-bisztyga, line 1: Sorry but I do not understand this file, it is very misleading. You called it `helper_functions.rs` but all I see here are tests. It doesn't make sense
Could you explain why did you move all these tests to a separate file in the first place? It is basically against the rust convention that says we should have such tests in the same file as functions. Can we move them back there?

cast/tests/data/accounts/user1_representation
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+SingleOwnerAccount { provider: JsonRpcClient { transport: HttpTransport { client: Client { accepts: Accepts, proxies: [Proxy(System({}), None)], referer: true, default_headers: {"accept": "*/*"} }, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(5055), path: "/rpc", query: None, fragment: None } } }, signer: LocalWallet { private_key: SigningKey { secret_scalar: FieldElement { inner: 0x0000000000000000000000000000000055ae34c86281fbd19292c7e3bfdfceb4 } } }, address: FieldElement { inner: 0x0112b61669a84940af30544786a6b615755b192173c4665a7d38fb3aa0541cac }, chain_id: FieldElement { inner: 0x0000000000000000000000000000000000000000000000534e5f474f45524c49 }, block_id: Tag(Latest) }
```
karol-bisztyga, line 1: can you explain why this changed?

cast/tests/e2e/helper_functions.rs
karol-bisztyga: Also, tests here don't look like e2e tests, are you sure they are e2e?

karol-bisztyga (changes requested):

cast/tests/data/accounts/user1_representation
THenry14: I have moved those tests to e2e as they are end to end. becasue of that devnet is no longer mocked, and thus we are using the port from config (which is 5055)

cast/tests/e2e/helper_functions.rs
THenry14: Sure, happy to help.

I have named this file helper_functions, as it is supposed to test all the helper functions we have (and I was not sure I can name it lib to idjicate those helper functions are from lib exclusively). This resides in test directory, where all the test files are present. If you have a better idea how to name it, let me know :) 

>Could you explain why did you move all these tests to a separate file in the first place? It is basically against the rust convention that says we should have such tests in the same file as functions. Can we move them back there?

the convention is that all the unit tests for specific file are to be in this file, but e2e tests should not be there + I nned them to be in e2e directory since 1. they are e2e 2. They require devnet, which is assured only for tests in e2e directory. We cannot, and we should not move them back, becasue they will break.

> 
Also, tests here don't look like e2e tests, are you sure they are e2e?

they are, they all setup up a provider, which makes them e2e.

THenry14:
Convertged to draft, because this PR will change a lot https://github.com/software-mansion/protostar/issues/2113

cast/tests/e2e/helper_functions.rs
karol-bisztyga: Discussed offline, and agreed we should move this to the `integration` folder as they are indeed integration tests, not e2e.

> If you have a better idea how to name it, let me know :)

Sure, I think this should be named `cast/tests/integration/lib_tests.rs` because it tests functions from `lib.rs`. `helper_functions.rs` doesn't tell much tbh.

karol-bisztyga: > the convention is that all the unit tests for specific file are to be in this file, but e2e tests should not be there

Yes, this is correct, thanks for clarifying, that also applies to integration tests.

PR #2151, opened 2023-06-29
[CLI] Add check for accounts file

### LINKED ISSUE #2147: Fix clap arg default value in rust CLI
```
#[derive(Parser)]
#[command(version)]
#[command(about = "cast - a starknet-foundry CLI", long_about = None)]
struct Cli {
    // ...
    #[clap(
        short = 'f',
        long = "accounts-file",
        default_value = "~/.starknet_accounts/starknet_open_zeppelin_accounts.json"
    )]
    accounts_file_path: Utf8PathBuf,

    // ...
}


#[tokio::main]
async fn main() -> Result<()> {
    let cli = Cli::parse();
    if !cli.accounts_file_path.exists() {
        anyhow::bail!("accounts file ({}) does not exist", cli.accounts_file_path);
    }
}
```

The `cli.accounts_file_path.exists()` will return `false` even though the file exists. This is because `Utf8PathBuf` does not seem to work properly with `~` (if you pass a full home path there like `/Users/root/home/...`, it will work).

THenry14:
### Checklist
- [X] Added relevant tests

Closes https://github.com/software-mansion/protostar/issues/2147

### DIFF
--- a/cast/src/main.rs
+++ b/cast/src/main.rs
@@ -1,5 +1,5 @@
 use crate::starknet_commands::{call::Call, declare::Declare, deploy::Deploy, invoke::Invoke};
-use anyhow::Result;
+use anyhow::{bail, Result};
 use camino::Utf8PathBuf;
 use cast::{get_account, get_block_id, get_network, get_provider, print_formatted};
 use clap::{Parser, Subcommand};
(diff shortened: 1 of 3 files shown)


cast/src/main.rs
war-in, line 1: Are we able to add a test for this case?

THenry14: I was thinking about how to approach it, and since we are doing e2e tests for commands I thought that maybe we should include main cli testing there as well? but am still unsure if that makes sense.

On one hand, unit tests for this would be probably more sensible, but on the other hand those tests could be a part of tests for some particular command, since we already test binary (with the cli arguments) there.

what do you think? any preference?

war-in: You're right. Those tests should appear in the test file for some particular command

war-in (approved):

Issue #2114, opened 2023-06-22
Add CI


THenry14:
Add code quality checks, building and tests to CI

PR #2160, opened 2023-07-03
[CLI] Invoke tests

### LINKED ISSUE #2111: Add tests for invoke
Add e2e tests for invoke command

war-in:
### Checklist
- [x] Added relevant tests

Closes #2111


karol-bisztyga:
This should target the master branch

karol-bisztyga (changes requested):
This should target the master branch

cast/tests/e2e/helpers/fixtures.rs
```diff
@@ -0,0 +1,75 @@
+use crate::helpers::constants::{
```
karol-bisztyga, line 1: the same as in https://github.com/software-mansion/protostar/pull/2150#discussion_r1250601906, I don't think these are e2e tests, or are they?

cast/tests/e2e/helpers/fixtures.rs
```diff
@@ -0,0 +1,75 @@
...
+        ACCOUNT,
+        &Utf8PathBuf::from(ACCOUNT_FILE_PATH),
+        provider,
+        &get_network(NETWORK).unwrap(),
+    )
+    .unwrap()
+}
+
+pub async fn declare_simple_balance_contract() {
+    let provider = get_provider(URL).unwrap();
```
karol-bisztyga, line 28: I do not think we should be using `unwrap` as much as you are, even in tests. Can you replace them at least with `expect`? Even this tiny error message can sometimes be helpful, you don't need to dig into the code, you sometimes just see the message and know what is wrong right away.

karol-bisztyga: What happens when a test panics? All tests break or just one fails?

karol-bisztyga (changes requested):

cast/tests/e2e/helpers/fixtures.rs
THenry14: these are not tests

cast/tests/e2e/helpers/fixtures.rs
```diff
@@ -0,0 +1,75 @@
+use crate::helpers::constants::{
+    ACCOUNT, ACCOUNT_FILE_PATH, CASM_CONTRACT_PATH, NETWORK, SIERRA_CONTRACT_PATH, URL,
+};
+use camino::Utf8PathBuf;
+use cast::{get_account, get_network, get_provider};
+use starknet::accounts::{Account, SingleOwnerAccount};
```
THenry14, line 6: did you try to construct raw requests? were there any problems?
I'm just afraid that we may be overly relying on starknet-rs, and wonder if we could somehow mitigate that

cast/tests/e2e/helpers/fixtures.rs
karol-bisztyga: right, sorry wrong file

cast/tests/e2e/helpers/fixtures.rs
```diff
@@ -0,0 +1,75 @@
...
+            "0x8448a68b5ea1affc45e3fd4b8b480ea36a51dc34e337a16d2567d32d0c6f8a",
+        )
+        .unwrap(),
+        account,
+    );
+    let deployment = factory.deploy(Vec::new(), FieldElement::from_hex_be("0x1").unwrap(), false);
+    deployment.send().await.ok();
+}
+
+pub fn common_cli_args() -> Vec<&'static str> {
```
THenry14, line 64: maybe `default_cli_args`? or something like this?

cast/tests/e2e/helpers/fixtures.rs
war-in: I've tried but there is still an issue with signing. We have to have an account to sign those transactions :/

war-in:
Please take a look at the new contract data setup @karol-bisztyga @THenry14 🙏

cast/tests/e2e/helpers/fixtures.rs
war-in: afaik just one

war-in: https://github.com/software-mansion/protostar/pull/2160/commits/d7c616450e7a4dc87ba63e50d23bc1f1ae4516de

cast/tests/data/contracts/balance/src/lib.cairo
```diff
@@ -0,0 +1,18 @@
+#[contract]
```
THenry14, line 1: It's a nitpick, but is the contract's name somehow imposed by scarb or something? if not, let's maybe rename it to something more meaningful?

cast/tests/utils/prepare_for_tests.sh
```diff
@@ -5,20 +5,43 @@ if ! which starknet-devnet > /dev/null 2>&1; then
...
-    pushd "$COMPILER_DIRECTORY"
-    tar -xzvf "$COMPILER_DIRECTORY/release-x86_64-unknown-linux-musl.tar.gz" cairo/bin/starknet-sierra-compile || exit 1
+    wget "$CAIRO_REPO/$COMPILER_VERSION/scarb-$SCARB_VERSION-x86_64-unknown-linux-musl.tar.gz" -P "$DIRECTORY" || exit 1
+    pushd "$DIRECTORY"
+    tar -xzvf "$DIRECTORY/scarb-$SCARB_VERSION-x86_64-unknown-linux-musl.tar.gz" scarb-$SCARB_VERSION-aarch64-apple-darwin/bin/scarb || exit 1
     popd
   fi
 fi
+
+pushd "$DIRECTORY/../data/contracts/balance"
```
THenry14, line 45: let's move contracts compilation to another script. 
We should probably call the compile_contract scrtips before each test run (firt check if scripts were compiled, if not do it, otherwise do nothing).

cast/tests/utils/prepare_for_tests.sh
```diff
@@ -5,20 +5,43 @@ if ! which starknet-devnet > /dev/null 2>&1; then
   exit 1
 fi
 
-COMPILER_DIRECTORY="$(git rev-parse --show-toplevel)/cast/tests/utils/"
+DIRECTORY="$(git rev-parse --show-toplevel)/cast/tests/utils/"
```
THenry14, line 8:
```suggestion
UTILS_DIRECTORY="$(git rev-parse --show-toplevel)/cast/tests/utils/"
```

cast/tests/utils/prepare_for_tests.sh
```diff
@@ -5,20 +5,43 @@ if ! which starknet-devnet > /dev/null 2>&1; then
...
+
+  elif [[ $(uname -s) == 'Linux' ]]; then
+    wget "$CAIRO_REPO/$COMPILER_VERSION/release-x86_64-unknown-linux-musl.tar.gz" -P "$DIRECTORY" || exit 1
+    pushd "$DIRECTORY"
+    tar -xzvf "$DIRECTORY/release-x86_64-unknown-linux-musl.tar.gz" cairo/bin/starknet-sierra-compile || exit 1
+    popd
+  fi
+fi
+
+if [ ! -x "$COMPILER_DIRECTORY/scarb*/bin/scarb" ]; then
```
THenry14, line 30: 1. can scarb do multiple versions? if so, we could just check if there's required scarb version in system
2. if not, let's get move `if` statement to separate function and reuse it

karol-bisztyga (changes requested):

war-in:
`[Disclaimer]` Many changes are here because the `master` branch was not correctly formatted

karol-bisztyga:
> `[Disclaimer]` Many changes are here because the `master` branch was not correctly formatted

I do not think this is good. I think there should be another PR for formatting and then this PR (invoke tests) should be rebased to the fmt PR.

war-in:
> I do not think this is good. I think there should be another PR for formatting and then this PR (invoke tests) should be rebased to the fmt PR.
Agree, but we should create some checks for formatting to avoid such a situation in the future

THenry14:
> Agree, but we should create some checks for formatting to avoid such a situation in the future

Next in pipeline #2114

cast/tests/utils/build_contracts.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+CONTRACTS_DIRECTORY="$(git rev-parse --show-toplevel)/cast/tests/data/contracts"
+
+if command -v scarb &> /dev/null; then
+  for contract_dir in "$CONTRACTS_DIRECTORY"/*; do
+    if ! test -d "$contract_dir"/target; then
+      pushd $contract_dir
```
karol-bisztyga, line 8:
```suggestion
      pushd "$contract_dir"
```

karol-bisztyga (changes requested):

cast/tests/data/contracts/balance/src/lib.cairo
THenry14: by contract name I meant both the name ( ( ͡° ͜ʖ ͡°) ) and the file name

cast/tests/e2e/helpers/devnet.rs
```diff
@@ -46,6 +45,10 @@ fn start_devnet() {
             std::process::exit(1);
         }
     }
+
+    let rt = Runtime::new().expect("Could not instantiate Runtime");
+    rt.block_on(declare_simple_balance_contract());
+    rt.block_on(deploy_simple_balance_contract());
```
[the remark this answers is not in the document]
THenry14: I suppose declare/deploy will work on different/seeded contracts, right? in that case, this is fine. otherwise, let's adjust to Karol's comment

cast/tests/e2e/helpers/fixtures.rs
```diff
@@ -0,0 +1,93 @@
...
+    provider: &JsonRpcClient<HttpTransport>,
+) -> SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet> {
+    get_account(
+        ACCOUNT,
+        &Utf8PathBuf::from(ACCOUNT_FILE_PATH),
+        provider,
+        &get_network(NETWORK).expect("Could not get the network"),
+    )
+    .expect("Could not get the account")
+}
```
THenry14, line 31: what's the purpose of this?

cast/tests/e2e/helpers/fixtures.rs
```diff
@@ -0,0 +1,93 @@
...
+        .expect("Could not create FieldElement from hex string"),
+        account,
+    );
+    let deployment = factory.deploy(
+        Vec::new(),
+        FieldElement::from_hex_be("0x1").expect("Could not create FieldElement from hex string"),
+        false,
+    );
+    deployment.send().await.unwrap();
+}
```
THenry14, line 80: I suppose that those functions could be merged into one? I believe you use them only in one place, one after another right? that way you don't need to hardcode hashes (I suppose), and get cleaner code

cast/tests/utils/build_contracts.sh
THenry14: actually, let's `""` all the variables

cast/tests/utils/prepare_for_tests.sh
```diff
@@ -5,20 +5,32 @@ if ! which starknet-devnet > /dev/null 2>&1; then
   exit 1
 fi
```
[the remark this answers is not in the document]
THenry14: > I know this was here already but I guess we can add it here it's just one line

Context matters. You can do `set -e`, but then `|| exit 1` are to be removed, because they are no longer needed. up to you if you prefer one way or another.

cast/tests/utils/prepare_for_tests.sh
```diff
@@ -5,20 +5,32 @@ if ! which starknet-devnet > /dev/null 2>&1; then
...
 
-COMPILER_DIRECTORY="$(git rev-parse --show-toplevel)/cast/tests/utils/"
+UTILS_DIRECTORY="$(git rev-parse --show-toplevel)/cast/tests/utils/"
```
THenry14, line 9: it seems thid directory is no longer needed for other things than compiler? so maybe we can revert this change altogether?

cast/tests/e2e/helpers/devnet.rs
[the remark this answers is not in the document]
war-in: Deploy tests are not compromised because we can use different salt for deployment

cast/tests/utils/prepare_for_tests.sh
```diff
@@ -5,20 +5,32 @@ if ! which starknet-devnet > /dev/null 2>&1; then
...
   fi
 fi
+
+if command -v asdf &> /dev/null; then
+  asdf plugin add scarb https://github.com/software-mansion/asdf-scarb.git
+  asdf install scarb $SCARB_VERSION
+  asdf global scarb $SCARB_VERSION
+  scarb --version
+else
+  printf "Please install asdf\n https://asdf-vm.com/guide/getting-started.html#_2-download-asdf\n"
```
THenry14, line 36: let's require it at the top of file, similarly to how we require devnet. Moreover, we should also check if scarb in need version is installed, and do it only if it is not.

cast/tests/e2e/helpers/devnet.rs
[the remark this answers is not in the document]
war-in: In the case of Declare, we will have to create some script which temporarily modifies the balance contract (eg. adds a character to one of the methods). It will change the class_hash

cast/tests/utils/prepare_for_tests.sh
war-in: If the required scarb version is here it will not do anything. It will just print scarb version

cast/tests/e2e/helpers/fixtures.rs
```diff
@@ -0,0 +1,72 @@
...
+use starknet::signers::LocalWallet;
+use std::sync::Arc;
+
+pub fn sierra_balance_path() -> String {
+    CONTRACTS_DIR.to_string() + "/balance/target/dev/balance_SimpleBalance.sierra.json"
+}
+
+pub fn casm_balance_path() -> String {
+    CONTRACTS_DIR.to_string() + "/balance/target/dev/balance_SimpleBalance.casm.json"
+}
```
THenry14, line 19: this is unnecessary probably, right?

cast/tests/utils/prepare_for_tests.sh
```diff
@@ -1,24 +1,38 @@
 #!/bin/bash
+set -e
 
 if ! which starknet-devnet > /dev/null 2>&1; then
   echo "starknet-devnet not found, exiting."
   exit 1
 fi
 
+SCARB_VERSION="0.4.1"
```
THenry14, line 9: All variables should be set in one place, preferably at the top

THenry14 (approved):
🔥

cast/tests/utils/prepare_for_tests.sh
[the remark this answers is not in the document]
karol-bisztyga: You are right, I think `set -e` is better (less boilerplate).

karol-bisztyga (approved):

